PR #10, opened 2022-01-07
Dbudzinski/fix configs

budzworthy:
## Description

This is a first step before fixing all of the failing integration tests.  Configs used during testing were coming from remote hosted locations which made it difficult to debug and impossible to protect against them changing unpredictably. I downloaded the files, put them into the repo, and mounted the directory for non-prod builds.

I updated the default (non-test) config to use the ott-test-account cleeng sandbox environment.

I also resolved the TODO regarding moving the code that allows the configs to be loaded. I moved it from an inline script in the html into a conditional in index.tsx. By using a compile time constant and the webpack DefinePlugin, this code is now statically disabled and subsequently optimized away by production builds.

Lastly, I changed the script that loads the configs to use session storage instead of local storage. This makes it harder to load a config that you aren't expecting, because the override used for testing only lasts within the tab/session instead of persisting indefinitely.

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


rcawkwell (approved):
Looks good. I appreciate all the in-line comments.

PR #16, opened 2022-02-15
Should it be _shelf_ or _shelve_?

rmnl:
## Description

IMHO, we should probably use the noun `shelf` vs the verb `shelve`.

This is a suggestion. If people do not agree, please say so and I will close the PR.


marcovandeveen:
Agreed: Shelf is the noun. Shelve is the verb.  https://writingexplained.org/shelf-or-shelve-difference 

The config.json change will break existing configs. 

So we should inform our users and change existing demo configs.

rmnl:
@marcovandeveen The nice thing about self-hosted solutions is that things will only break if they update to the latest version.

We should still notify them. A good way to start is probably to add [a `CHANGELOG.md` file](https://keepachangelog.com/en/1.0.0/) to the source code to indicate the changes (after the initial real release).

budzworthy (approved):

PR #17, opened 2022-03-04
fix: remove unfinished mutations for consents and customer data and move logic to account store

budzworthy:
## Description

The user / account components had a half-working query / mutation implementation for retrieving and updating customer and consent data.  The mutations and queries worked to initially load data but did not properly update nor handle errors. Additionally, there was a confusing duplication of customer data between the queries and the Account store and there were spaghetti dependencies on account store and account service at different levels.  To clean up and solve all of these issues, I moved the consent logic into the Account Store and removed all the mutation and query stuff.

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [x] Unit tests passing
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


src/components/Account/Account.tsx
```diff
@@ -69,32 +49,68 @@ const Account = ({
...
+    errors?.map((error) => {
+      switch (error) {
+        case 'Invalid param email':
+          formErrors.email = 'Invalid email address!';
+          break;
+        case 'Customer email already exists':
+          formErrors.email = 'Email already exists!';
+          break;
+        case 'Please enter a valid e-mail address.':
+          formErrors.email = 'Please enter a valid e-mail address.';
```
CarinaDraganJW: Maybe we should have `email` instead of `e-mail` to be consistent, just a small detail 🙂
Or have the other informative errors written as `e-mail `

src/components/Account/__snapshots__/Account.test.tsx.snap
```diff
@@ -115,12 +127,55 @@ exports[`<Account> renders and matches snapshot 1`] = `
...
+        class="flexBox"
+      >
+        <div
+          class="checkbox"
+        >
+          <div
+            class="row"
+          >
+            <input
+              id="check-box_1235_marketing"
```
CarinaDraganJW, line 150: Do we have numbered ids for all inputs? Like in this example? 🙂

CarinaDraganJW (approved):
Looks good to me 🙂

src/components/Account/Account.tsx
budzworthy: Good point.  And these should probably all be added to translations instead of how they are hard-coded.  Will update.

src/components/Account/__snapshots__/Account.test.tsx.snap
budzworthy: Yes, these ID's are automatically created by the "useOpaqueId" function

PR #19, opened 2022-03-08
Dbudzinski/fix form component

budzworthy:
## Description

This PR touches a lot of files, but the bulk of the changes are in the Form component.

The changes in Checkbox, DateField, DropDown, Radio, and TextField all are to abstract the "HelperText" functionality into a common reusable component.

The changes in the random .scss files are just to clean up errors and warnings.

Lastly, the changes in Form, FormSection, and Account all relate to redesigning the logic of the form component.  I moved the section logic that was in Account to be built into Form and FormSection.  I also moved a lot of the boilerplate (buttons, events, layout, etc.) out of Account and into the Form so that the reusable Form provides more complete, encapsulated functionality.  The Form serves as a common context provider to pass state into all of the child sections.  Then each section is a fairly independent entity, each with their own layout, buttons, submit action, etc.  As a result, Account only contains the logic and layout specific to the User values being edited on this page.

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [x] Unit tests passing
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


src/components/HelperText/HelperText.test.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import HelperText from './HelperText';
+
+describe('<HelperText>', () => {
+  test('renders and matches snapshot', () => {
+    const { container } = render(<HelperText error={true}>Hello world!</HelperText>);
+
+    expect(container).toMatchSnapshot();
```
AntonLantukh, line 10: Do we need simple snapshot tests? Maybe it is better to check presence / absence of the elements simulating necessary branch of the if condition?

src/components/HelperText/HelperText.module.scss
```diff
@@ -0,0 +1,12 @@
+@use '../../styles/variables';
+@use '../../styles/theme';
+
+.helperText {
+  margin-top: 4px;
+  font-size: 12px;
```
AntonLantukh, line 6: Maybe we will need a common component to deal with font (like we have in the dashboard). That is just a suggestion for future development :)

src/components/RegistrationForm/RegistrationForm.tsx
```diff
@@ -120,6 +117,7 @@ const RegistrationForm: React.FC<Props> = ({
           name={consent.name}
           value={consent.value || ''}
           error={consentErrors?.includes(consent.name)}
+          helperText={consentErrors?.includes(consent.name) ? 'This field is required' : undefined}
```
AntonLantukh: I think we also need to include 'This field is required' in our i18n translations pack.

src/components/Checkbox/Checkbox.module.scss
```diff
@@ -108,8 +101,9 @@
...
-.helperText {
-  margin-top: 4px;
-  font-size: 12px;
-  text-align: left;
-}
+.helperTextError {
+  margin-top: -5px;
+  margin-left: 35px;
+  color: theme.$text-field-error-color;
+  font-size: 10px;
```
AntonLantukh, line 108: Do we really need custom font-size just for the Checkbox case? In addition, color seems to be already present in the common component.

src/components/Dialog/Dialog.module.scss
```diff
@@ -11,6 +11,7 @@
   background-color: theme.$modal-bg;
   border-radius: 6px;
   -ms-overflow-style: none;
+  //noinspection CssUnknownProperty
```
AntonLantukh, line 14: That is strange that the property in unknown. Does PyCharm highlights this line?

src/components/Form/Form.tsx
```diff
@@ -1,70 +1,77 @@
...
-    event && event.preventDefault();
-    onSubmit && onSubmit(values);
-    setHasChanged(false);
-  };
-
-  const handleReset = () => {
-    setValues(initialValues);
-    setHasChanged(false);
-  };
+  function cancel() {
```
AntonLantukh: I think it is better to wrap it using useCallback to recreate the function only if dependencies change.

src/components/Account/Account.tsx
```diff
@@ -78,187 +79,154 @@ const Account = ({ panelClassName, panelHeaderClassName }: Props): JSX.Element =
             formErrors.lastName = t('account.errors.last_name_too_long');
             break;
           }
-          default:
-            console.info('Unknown error', error);
-            return;
+          default: {
+            formErrors.form = 'An unknown error occurred';
+            // @ts-ignore
+            if (typeof NODE_ENV_COMPILE_CONST === 'undefined' || NODE_ENV_COMPILE_CONST !== 'production') {
```
AntonLantukh: If ts can't find NODE_ENV_COMPILE_CONST we can add typing for it somewhere in the root place:

```
declare namespace globalThis {
  interface Window {
    NODE_ENV_COMPILE_CONST: 'production' | 'undefined';
  }
}

```

src/components/Account/Account.tsx
```diff
@@ -78,187 +79,154 @@ const Account = ({ panelClassName, panelHeaderClassName }: Props): JSX.Element =
             formErrors.lastName = t('account.errors.last_name_too_long');
             break;
           }
-          default:
-            console.info('Unknown error', error);
-            return;
+          default: {
+            formErrors.form = 'An unknown error occurred';
```
AntonLantukh: I think it is also better to include it in i18n translations pack

src/components/Account/Account.tsx
```diff
@@ -78,187 +79,154 @@ const Account = ({ panelClassName, panelHeaderClassName }: Props): JSX.Element =
...
-          <div className={panelClassName}>
-            <div className={panelHeaderClassName}>
-              <h3>{t('account.email')}</h3>
-            </div>
-            <div className={styles.flexBox}>
+    <Form initialValues={initialValues}>
+      {[
+        formSection({
+          label: t('account.email'),
+          submit: (values) => updateUser({ email: values.email || '', confirmationPassword: values.confirmationPassword }),
```
AntonLantukh: Maybe it is better to call it onSubmit like all event handlers?

AntonLantukh (changes requested):
My first experience of working with ott code base :) Separation of logic looks nice, even though I would prefer to have just one submit button / handler being on the Form level. So that we could work with the submit logic in one place.

src/components/Account/Account.tsx
budzworthy: Ah good catch, thank you

src/components/Account/Account.tsx
budzworthy: Sure, if that makes it clearer

src/components/Form/Form.tsx
budzworthy: Good suggestion, thanks!

src/components/HelperText/HelperText.test.tsx
budzworthy: Snapshot tests are quick and easy to add. The e2e tests exercise different branches and conditions.  But then there's a whole backlog of UT's that we want/need to add.

src/components/Account/Account.tsx
budzworthy: I added the typing and I also moved the construct to a common util.

src/components/Checkbox/Checkbox.module.scss
budzworthy: In this case I just moved the styles, I didn't change the design at all.  We can review the design when we get to a good point.

src/components/Dialog/Dialog.module.scss
budzworthy: It does. Not sure why, but a google search didn't help, so I just added the disable comment.

src/components/HelperText/HelperText.module.scss
budzworthy: I will have to check that implementation out.  It sounds like it might help reduce complexity, which would be nice.

src/components/RegistrationForm/RegistrationForm.tsx
budzworthy: Very thorough review. Thank you for catching these.

src/components/Form/FormSection.tsx
```diff
@@ -24,140 +24,154 @@ export interface FormSectionProps<TData extends GenericFormValues, TErrors> {
...
-    const value = event.currentTarget.type === 'checkbox' ? (event.currentTarget as HTMLInputElement).checked : event.currentTarget.value;
+  const onChange = useCallback(
+    function onChange(event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>) {
+      if (!event.currentTarget) return;
 
-    if (activeSectionId !== sectionId) {
-      cancel();
-    }
+      const name = event.currentTarget.name;
+      const value = event.currentTarget.type === 'checkbox' ? (event.currentTarget as HTMLInputElement).checked : event.currentTarget.value;
```
AntonLantukh:
```
      const {name, type, checked, value} = event.currentTarget as HTMLInputElement;
      const value = type === 'checkbox' ? checked : value;
```

src/utils/common.ts
```diff
@@ -49,3 +49,7 @@ export function calculateContrastColor(color: string) {
   // http://stackoverflow.com/a/3943023/112731
   return rgb.r * 0.299 + rgb.g * 0.587 + rgb.b * 0.114 > 186 ? '#000000' : '#FFFFFF';
 }
+
+// NODE_ENV_COMPILE_CONST is defined at compile time using the webpack define plugin.
+// NODE_ENV_COMPILE_CONST is undefined when running the snowpack dev server, so allow test code in that case
+export const isDevCompileTimeConstant = typeof NODE_ENV_COMPILE_CONST === 'undefined' || NODE_ENV_COMPILE_CONST !== 'production';
```
AntonLantukh: Maybe it is also better to capitalize it, like `IS_DEV_COMPILE_TIME`

src/components/Form/FormSection.tsx
```diff
@@ -0,0 +1,181 @@
...
+                <Button label={saveButton} type="submit" onClick={handleSubmit} disabled={!isDirty || isLoading || (canSave && !canSave(values))} />
+              )}
+              {cancelButton && <Button label={cancelButton} type="reset" variant="text" onClick={cancel} />}
+            </>
+          ) : (
+            editButton &&
+            (typeof editButton === 'object' ? (
+              (editButton as ReactElement)
+            ) : (
+              <Button label={editButton as string} type="button" onClick={edit} disabled={isLoading} />
```
AntonLantukh: `onEdit`

src/components/Form/FormSection.tsx
```diff
@@ -0,0 +1,181 @@
...
+        <div className={styles.flexBox}>{content({ values, isEditing, isBusy, onChange })}</div>
+      )}
+      {(saveButton || editButton || cancelButton) && (
+        <div className={styles.controls}>
+          {isEditing ? (
+            <>
+              {saveButton && (
+                <Button label={saveButton} type="submit" onClick={handleSubmit} disabled={!isDirty || isLoading || (canSave && !canSave(values))} />
+              )}
+              {cancelButton && <Button label={cancelButton} type="reset" variant="text" onClick={cancel} />}
```
AntonLantukh: `onCancel` - I think that we need also to switch to this conception here

AntonLantukh:
Thank you! I still suggest to handle all the submit logic in one place, as it seems to be more of the Form than of FormSection logic. Other things look great except some small issues.

src/components/Form/FormSection.tsx
budzworthy: I have no real preference for one of these snippets over the other, though I tend to try to avoid intermediate variables. However, I refactored to use destructuring more efficiently.

src/components/Form/FormSection.tsx
budzworthy: Done

src/components/Form/FormSection.tsx
budzworthy: Done

budzworthy:
> 

So I would agree about the submit logic, except that different sections submit to different endpoints, so having 1 common submit function forces coupling and input type concatenation that otherwise wouldn't need to exist. I tried to even go so far as to make the error return type fully generic with the return type of the submit method, but couldn't get it working in a clean enough way.

AntonLantukh (approved):

PR #21, opened 2022-03-29
Dbudzinski/fix desktop tests

budzworthy:
## Description

Sorry that this PR is so big, but it seemed better to just fix all of the tests and get it done with.  Most of the relevant changes are in the test-e2e directory, and all of the changes in the src files are fixes for small bugs found during testing. Note that a large amount of churn in the tests came from combining the desktop and mobile versions of e2e tests that were basically the same except for a few small navigational differences, which I moved into logic in the tests.

This PR solves # .

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [x] Unit tests passing
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


src/components/LoginForm/LoginForm.tsx
```diff
@@ -46,6 +46,7 @@ const LoginForm: React.FC<Props> = ({ onSubmit, onChange, values, errors, submit
         name="email"
         type="email"
         required
+        testId="email-input"
```
CarinaDraganJW: Do we have other places in the app where maybe we have `email-input` used as a `testId`? Maybe here we could have something more specific like `login-email-input` 🙂

test-e2e/utils/steps_file.ts
```diff
@@ -90,7 +177,7 @@ module.exports = function() {
       // Since this check executes a script in the browser, it won't use the codecept retries,
       // so we have to manually retry (this is because the video can take time to load and the state will be buffering)
       for (let i = 0; i < tries; i++) {
-        const state = await this.executeScript(() => typeof jwplayer === 'undefined' ? '' : jwplayer().getState());
+        const state = await this.executeScript(() => typeof jwplayer === 'undefined' || typeof jwplayer().getState === 'undefined' ? '' : jwplayer().getState());
```
CarinaDraganJW, line 181: Can we maybe have directly `!jwplayer || !jwplayer().getState` instead of `typeof jwplayer === 'undefined'`? 
If `jwplayer` variable will be `undefined`, it will be a `falsy` value
Maybe you can try to see if it still works with this adjustment 🙂

test-e2e/utils/steps_file.ts
```diff
@@ -113,6 +200,29 @@ module.exports = function() {
       // eslint-disable-next-line no-console
       assert.equal(await this.executeScript(() => typeof jwplayer === 'undefined' ? undefined : jwplayer().getState),
           undefined);
+    },
+    isMobile: async function(this: CodeceptJS.I, isMobile?: boolean) {
+      if (isMobile !== undefined) {
```
CarinaDraganJW: Same here as above, maybe we can have `if(!isMobile) {...}` instead of `isMobile !== undefined` 🙂

CarinaDraganJW (approved):
Looks good to me, nice work 🙂

test-e2e/utils/steps_file.ts
budzworthy: That should work, but something about the way that codeceptjs processes the page.evaluate causes the error below unless I write it exactly like this.  I think it's worth putting a comment for the next person though. Thanks for pointing it out.
```
 page.evaluate: ReferenceError: jwplayer is not defined
    at eval (eval at evaluate (:3:2389), <anonymous>:1:7)
    at t.default.evaluate (<anonymous>:3:2412)
    at t.default.<anonymous> (<anonymous>:1:44)
```

src/components/LoginForm/LoginForm.tsx
budzworthy: There shouldn't be multiple email inputs on the screen, but by making the id more specific it would prevent obscure behavior, so good idea to change it.  However, when I went to make the change, I realized it's not actually used anywhere. I'm not sure when exactly it got added, but it doesn't seem to be needed anymore, so I'll remove the bloat.

test-e2e/utils/steps_file.ts
budzworthy: This code is also unused. I think the intention was to provide a short circuit if the input was defined (true or false) to not repeat the check, so callers could kind of just call it with a potentially existing variable in-line, but it's really confusing.  Thanks for flagging this, I'll just remove the code.

src/components/LoginForm/LoginForm.tsx
budzworthy: Just kidding - April Fool's on me.  It is used in the login tests, but the string is interpolated so ctrl+F didn't find it. I added the login prefix as suggested and I also made the attributes condition only if IS_DEV_BUILD is true.  I tried to fix the spelling (to data-test-id instead of data-testid) but the latter is hard coded into react testing framework.

PR #48, opened 2022-04-26
fix(auth): Make logout into a function

### LINKED ISSUE #37: Make logout into a function, not a route
Right now logout is handled via a route, which can cause some browser pre-fetch issues as well as opening some attack vectors (see https://stackoverflow.com/questions/3521290/logout-get-or-post.) To fix, create a logout function called by the buttons that currently link to /u/logout and remove the route.

AntonLantukh:
## Description

* Fix UserMenu component in the header
* Fix User component in the cabinet
* Update snapshot
* Update e2e-tests

This PR solves: https://github.com/jwplayer/ott-web-app/issues/37

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/screens/User/User.tsx
```diff
@@ -40,6 +39,9 @@ const User = (): JSX.Element => {
 
   const onCardClick = (playlistItem: PlaylistItem) => history.push(cardUrl(playlistItem));
   const onCardHover = (playlistItem: PlaylistItem) => updateBlurImage(playlistItem.image);
+  const onLogout = useCallback(() => {
+    logout();
```
budzworthy, line 44: Did the history.push get dropped here?  Is it possible to just move that to the account store logout function (probably using `window.location.href = '/'` to avoid problems with hooks)

budzworthy (approved):
One small point about the history.push, otherwise looks good

src/screens/User/User.tsx
AntonLantukh: @dbudzins Yeah, right after this hook we have one more with `history.replace`. So it does all the job because the customer (user) gets `null` value in `logout` function.

```
  useEffect(() => {
    if (!loading && !customer) {
      history.replace('/');
    }
  }, [history, customer, loading]);
```

budzworthy: Ah, I see that and it certainly should work as you describe.  Do you think it might be easier to read / follow / update the code if the navigation lived inside the logout function though?  For example if someone wants to change the post-logout page or let users stay on the same page after logout?  Alternatively, if you want to leave the navigation in the hook, can you add a comment that this returns the user to the main page after logout and also remove the history.push in UserMenu?

AntonLantukh: @dbudzins thank you for ideas! I modified the solution a bit and created a new Provider to control page access. And it also does redirect to the base page if the user is not defined. What do you think about this approach?

src/providers/AuthProvider.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { Redirect } from 'react-router-dom';
+
+import { AccountStore } from '../stores/AccountStore';
+
+type Props = {
+  children: JSX.Element;
+};
+
+function AuthProvider({ children }: Props): JSX.Element | null {
```
budzworthy: I think this is a small enough amount of code that it could just be directly in Root.tsx, but this could also eventually expand to have some other user or account related side effects.  If you are going to keep it as a provider, it should probably follow the Context.Provider structure like the other providers.

AntonLantukh: I changed it back to `useEffect` hook and left the comment. When I understand all the possibles corner cased I will switch to the provider logic if possible.

Issue #36, opened 2022-04-20
Upgrade build from snowpack to vite


budzworthy:
Snowpack popularity is down and partners and other teams prefer vite

PR #49, opened 2022-05-02
Dbudzinski/new build

budzworthy:
## Description

To keep sync with partner and industry trends, we want to migrate off of snowpack and to vite.  This PR does just that. Most of the changes are small 1-1 migrations (changing the build constant naming, updating the snapshots with the Vite header, updating test helpers from jest to vitest, etc.)

As mentioned above, I also migrated from jest to vitest.  Jest doesn't work nicely with the vite build pipeline, and would require workarounds and a secondary test-only build setup, while vitest, though relatively new, works nicely with vite out of the box. 

This PR solves #36.

### Steps completed:

- [x] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [x] Unit tests passing
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


CONTRIBUTING.md
```diff
@@ -92,7 +92,7 @@ While far from comprehensive, following these guidelines will get the easy stuff
 
 ##### Style
 
-1. This project uses snowpack + webpack plugin to build and compile. **Code produced by the build must work across all supported browsers and devices - Chrome, FF, Safari, Edge, and IE 9-11**
+1. This project uses vite to build and compile. **Code produced by the build must work across all supported browsers and devices - Chrome, FF, Safari, Edge, and IE 9-11**
```
AntonLantukh: I see that we have a `browserlist` property in `package.json` which doesn't support IE11. Vite should  use it as a reference.

package.json
```diff
@@ -7,28 +7,31 @@
   "private": true,
   "scripts": {
     "prepare": "husky install",
-    "start": "snowpack dev",
-    "build": "snowpack build",
-    "test": "TZ=UTC jest",
-    "test-watch": "TZ=UTC jest --watch",
-    "test-coverage": "TZ=UTC jest --coverage",
+    "start": "vite",
+    "build": "tsc && vite build",
```
AntonLantukh: Hm, do we need `tsc` here? I thought `vite` transpiles everything on its own.

src/screens/User/User.test.tsx
```diff
@@ -1,14 +1,12 @@
 // import React from 'react';
 // import { render } from '@testing-library/react';
 
-import { mockMatchMedia } from '../../testUtils';
-
 // import Home from './Home';
 
-describe('Home Component tests', () => {
-  mockMatchMedia();
+describe.skip('Home Component tests', () => {
```
AntonLantukh:
```suggestion
describe.skip('User Component tests', () => {
```

src/components/Router/Router.tsx
```diff
@@ -3,7 +3,7 @@ import { BrowserRouter, HashRouter } from 'react-router-dom';
 import type { BrowserRouterProps, HashRouterProps } from 'react-router-dom';
 
 export default function Router(props: BrowserRouterProps | HashRouterProps) {
-  if (import.meta.env.SNOWPACK_PUBLIC_GITHUB_PAGES) {
+  if (import.meta.env.APP_PUBLIC_GITHUB_PAGES) {
```
AntonLantukh, line 6: It seems, that if we are not in a build mode, this value equals to `undefined`. Maybe we should better define it in `vite-config` in `define` section?

docs/easy-deployments.md
```diff
@@ -40,8 +40,8 @@ Github pages have some [usage limits](https://docs.github.com/en/pages/getting-s
...
 
-1. Runs `yarn build` with `SNOWPACK_PUBLIC_BASE_URL` envvar.<br/><br />That envvar is used to set the URL location of the project. By default if `SNOWPACK_PUBLIC_BASE_URL` is empty, the value will be based on the `git remote get-url origin` command. You can also pass your own `SNOWPACK_PUBLIC_BASE_URL` envvar by running `SNOWPACK_PUBLIC_BASE_URL=/my-base/ yarn deploy:github`.
-2. Runs `yarn gh-pages -o origin -d build`.<br /><br />You can change the remote from *origin* to *myremote* by running `yarn deploy:github --github-remote=myremote`.  The `yarn deploy:github` command uses the GitHub remote to compute the default value for `SNOWPACK_PUBLIC_BASE_URL`.
+1. Runs `yarn build` with `APP_GITLAB_PUBLIC_BASE_URL` envvar.<br/><br />That envvar is used to set the URL location of the project. By default if `APP_GITLAB_PUBLIC_BASE_URL` is empty, the value will be based on the `git remote get-url origin` command. You can also pass your own `APP_GITLAB_PUBLIC_BASE_URL` envvar by running `APP_GITLAB_PUBLIC_BASE_URL=/my-base/ yarn deploy:github`.
```
AntonLantukh: Hm, why do we use `gitlab` name here? As I see, we use `github-pages` to deploy.

AntonLantukh (changes requested):
Nice one! I also hope to spend some time today or tomorrow to pull the branch and to test the updated build / dev process.

package.json
budzworthy: I copied it from the React-TS project referenced in the docs: https://vitejs.dev/guide/#overview.  I just compared build output with and without it and it seems to be the same... maybe the extra tsc provides better console output in the event of a typescript problem?

CONTRIBUTING.md
budzworthy: I don't think anything is actually still using the browserslist.  We should talk to product about if we need to support ie11... it officially goes to unsupported by MS in 6 weeks, so if we can avoid the extra overhead and polyfills it would be nice.

src/components/Router/Router.tsx
budzworthy: Does it cause a problem if it's undefined?  Both with the dev server and serving the static build files, it just evaluates to falsey for me without any errors or warnings.

docs/easy-deployments.md
budzworthy: Haha, good catch. This is a typo that's been there for who knows how long...

CONTRIBUTING.md
budzworthy: I just double checked and this got added when I merged a bunch of content from the contributing guide of the player. Based on the browserslist, I don't think that was the intent, so I'm going to remove it.

src/components/Router/Router.tsx
AntonLantukh: It does not. Just looks more intentional to have it defined.

AntonLantukh (approved):
Thanks for fixes! I also highlighted some issues with dependencies. It seems that we can get rid of async, jake and minimist, c8(?).

PR #55, opened 2022-05-17
Add Zustand and refactor the AccountStore

ChristiaanScheermeijer:
## Description

Add Zustand dependency and refactor the account store with Zustand. I also moved all business logic to a separate AccountController file.

Relates to #51


src/stores/utils.ts
```diff
@@ -0,0 +1,11 @@
+import type { State, StateCreator } from 'zustand';
+import create from 'zustand';
+import { devtools, subscribeWithSelector } from 'zustand/middleware';
+
+export const createStore = <T extends State>(name: string, storeFn: StateCreator<T>) =>
+  create(
+    devtools(subscribeWithSelector(storeFn), {
```
budzworthy: Should the devtools stuff be wrapped with the `IS_DEV_BUILD` constant so that it's compiled away for prod?

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -28,7 +28,7 @@ const AccountModal = () => {
   const viewParam = useQueryParam('u');
   const [view, setView] = useState(viewParam);
   const message = useQueryParam('message');
-  const { loading, auth } = AccountStore.useState((s) => s);
+  const { loading, auth } = useAccountStore();
```
royschut: Isn't destructuring from the function more performant?
```suggestion
  const [loading, auth] = useAccountStore((s) => [s.loading, s.auth]);
```

src/containers/AccountModal/forms/RenewSubscription.tsx
```diff
@@ -4,14 +4,16 @@ import { useHistory } from 'react-router';
...
 import RenewSubscriptionForm from '../../../components/RenewSubscriptionForm/RenewSubscriptionForm';
 import SubscriptionRenewed from '../../../components/SubscriptionRenewed/SubscriptionRenewed';
 
+import { updateSubscription } from '#src/stores/AccountController';
+
 const RenewSubscription = () => {
   const { t } = useTranslation('account');
   const history = useHistory();
-  const { subscription, user } = AccountStore.useState((s) => s);
+  const { subscription, user } = useAccountStore();
```
royschut:
```suggestion
  const [subscription, user] = useAccountStore((s) => [s.subscription, s.user]);
```

src/containers/AccountModal/AccountModal.tsx
budzworthy: I doubt there's much performance difference with any of these solutions. My vote would be for object destructuring over array destructuring. It's a bit safer in getting the right properties to the right variables. With the array style, if you have 2 properties of the same type, it's easier to mix them up and not notice.

src/stores/utils.ts
ChristiaanScheermeijer: I thought that the middleware already did this, but it doesn't seem so. Thanks, I will change that!

src/containers/AccountModal/AccountModal.tsx
ChristiaanScheermeijer: Yes, agreed!

src/containers/AccountModal/forms/RenewSubscription.tsx
ChristiaanScheermeijer: Yes, agreed!

PR #56, opened 2022-05-17
Refactor ConfigStore to Zustand

ChristiaanScheermeijer:
## Description

This PR refactors the ConfigStore with Zustand.

Relates to #51 

Depends on #55


src/screens/Playlist/Playlist.tsx
```diff
@@ -26,8 +26,8 @@ function Playlist({
   },
 }: RouteComponentProps<PlaylistRouteParams>) {
   const history = useHistory();
-  const config = ConfigStore.useState((state) => state.config);
-  const accessModel = ConfigStore.useState((s) => s.accessModel);
+  const config = useConfigStore((state) => state.config);
```
budzworthy: Any reason not to combine these 2 configStore calls? I doubt there's much difference in efficiency, but the code is a little more concise with a single assignment, and it will scale better if anyone needs to get a 3rd property.

src/screens/Search/Search.tsx
```diff
@@ -28,9 +28,9 @@ const Search: React.FC<RouteComponentProps<SearchRouteParams>> = ({
   },
 }) => {
   const { t } = useTranslation('search');
-  const config = ConfigStore.useState((state) => state.config);
+  const config = useConfigStore((state) => state.config);
```
budzworthy: Combine?

src/screens/Series/Series.tsx
```diff
@@ -41,8 +41,8 @@ const Series = ({ match, location }: RouteComponentProps<SeriesRouteParams>): JS
   const feedId = searchParams.get('l');
 
   // Config
-  const { options, siteName } = ConfigStore.useState((s) => s.config);
-  const accessModel = ConfigStore.useState((s) => s.accessModel);
+  const { options, siteName } = useConfigStore((s) => s.config);
```
budzworthy: Combine?

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -10,16 +10,16 @@ import { getOffer } from '../../../services/checkout.service';
...
 import type { ChooseOfferFormData, OfferPeriodicity } from '#types/account';
 
 const ChooseOffer = () => {
   const history = useHistory();
   const { t } = useTranslation('account');
-  const config = ConfigStore.useState((s) => s.config);
+  const config = useConfigStore((s) => s.config);
   const { cleengSandbox, json } = config;
-  const accessModel = ConfigStore.useState((s) => s.accessModel);
+  const accessModel = useConfigStore((s) => s.accessModel);
```
budzworthy: Combine?

src/screens/Playlist/Playlist.tsx
ChristiaanScheermeijer: Yes, good point. We had a little debate regarding this subject in our team. I think there are 4 similar options for this, but we probably should pick one and use that primarily. We like to use the Array destructuring way.

Looking at https://github.com/pmndrs/zustand#selecting-multiple-state-slices we need to add the `shallow` comparator function so that it only updates when a selected value changes.

PR #63, opened 2022-05-17
Refactor UIStore with Zustand

ChristiaanScheermeijer:
## Description

- Refactor the UIStore

Relates to #51


src/containers/Layout/Layout.tsx
```diff
@@ -48,17 +53,17 @@ const Layout: FC<LayoutProps> = ({ children }) => {
   }, [searchActive]);
 
   const searchButtonClickHandler = () => {
-    UIStore.update((s) => {
-      s.searchActive = true;
-      s.preSearchPage = history.location;
+    useUIStore.setState({
+      searchActive: true,
+      preSearchPage: history.location,
```
royschut, line 58: Minor, but we could create a setter in UIStore which leads to `setSearchActive(true, history.location)`

royschut (approved):

Issue #25, opened 2022-04-05
Cleeng TVOD support


marcovandeveen:
TVOD is the opposite of SVOD. Rather than giving viewers access to a broad content catalog, content owners would give them access to a single piece of content. It's was a way to monetize popular premium content and recent new releases.
TVOD can either be purely transactional (pay for unlimited access), or it can be a rental model. 

## Scope
- TVOD only 
- TVOD titles, in an SVOD model
- TVOD with upgrade to  SVOD 

## Out-of-scope
- TVOD for (instant) live streams

## Entitlement pattern
Aligned with Applicaster approach:
- Video can be viewed using SVOD or TVOD:
  - media field: `productIds`
  - field value: `<vendor-prefix>:<productid>, ...`
  -  example: `cleeng:S916977979_NL, cleeng:S91633379_NL`
- Video can be viewed TVOD only:
  - media field: `productIds` 
  - field value: `!<vendor-prefix>:<productid>, ... ` (exclamation mark)
  - example: `!cleeng:S916977979_NL`
- Video is doesn' require 
  - media field: `free` 
  - field value:  true or empty. Note: The value doesn't matter. 
  - example: `!cleeng:S916977979_NL`
- Cleeng vendor-prefix: `cleeng`

## Details
https://docs.google.com/presentation/d/1zxn-XxFYSaFr9o7YcGdm4iC17DzsmaLkswStda4zRNA/edit#slide=id.g63129a3642_0_0

## Notes
- Align with Applicaster productId
- Support SVOD as well
- Exclamation mark (!) to override global offers for premier access)
- Added: free (instead of 'requiresSubscription')

baukef:
Could you please confirm that TVOD for live streams (including count down to live event and purchasing something that is not available to watch yet, etc) is not a requirement for this sprint?

marcovandeveen:
Marked it as out of scope.

ChristiaanScheermeijer:
Fixed by #66

marcovandeveen:
Note from @ChristiaanScheermeijer 

Rearding the platform access model, we’ve stumbled upon a small issue. This is easily fixed by adding a param to the configuration, but I think it helps to explain this anyway.

We have three access models; AVOD, AUTHVOD and SVOD. The new TVOD access modal can be combined with any existing access model. Since we determine the access model based on the following conditions:
- AVOD: no cleengId/svod offers configured
- AUTHVOD: only cleengId configured
[...]

PR #65, opened 2022-05-18
Feat / entitlement check

royschut:
## TVOD preparations
- Implements the Cleeng entitlement endpoint
- Adds the `useEntitlement()` hook to call this endpoint from the Movie and Series screens. (React-query is used so that the result is cached)
- Updates the `isAllowedToWatch()` utility with the new TVOD accessModel. In regular TVOD cases, both (SVOD + subscription) and (TVOD offer + isEntitled) give access. Whenever a forced offer is applied (`!cleeng: <id>`), this overrules any SVOD subscriptions and the entitlement is always required.

Relates to #25


src/screens/Movie/Movie.tsx
```diff
@@ -49,23 +50,31 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
   const { isLoading, error, data: item } = useMedia(id);
-  const itemRequiresSubscription = item?.requiresSubscription !== 'false';
   useBlurImageUpdater(item);
   const { data: trailerItem } = useMedia(item?.trailerId || '');
   const { data: playlist } = useRecommendedPlaylist(recommendationsPlaylist || '', item);
 
-  const isFavorited = useFavoritesStore((state) => !!item && state.hasItem(item));
+  // AccessModel & entitlement
+  const mediaOffer = useMemo(() => (item?.productIds && filterCleengMediaOffers(item.productIds)) || undefined, [item]);
+  useEntitlement(mediaOffer);
```
ChristiaanScheermeijer: Can this be removed?

royschut: It re-appeared after a merge. Thanks, fixed!

royschut: It re-appeared after a merge. Thanks, fixed!

src/screens/Movie/Movie.tsx
```diff
@@ -49,23 +50,31 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
+  // User, accessModel, entitlement
   const { user, subscription } = useAccountStore(({ user, subscription }) => ({ user, subscription }), shallow);
-  const allowedToWatch = isAllowedToWatch(accessModel, !!user, itemRequiresSubscription, !!subscription);
+  const isItemFree = item?.requiresSubscription === 'false' || !!item?.free;
+
+  const mediaOffers = useMemo(() => filterCleengMediaOffers(item?.productIds), [item]);
+  const hasForcedOffer = mediaOffers?.some((offer) => offer.forced);
+  const skipEntitlement = isItemFree || (subscription && !hasForcedOffer);
+  const { isEntitled } = useEntitlement(mediaOffers, !skipEntitlement);
+  const allowedToWatch = isAllowedToWatch(hasForcedOffer ? 'TVOD' : accessModel, !!user, isItemFree, !!subscription, isEntitled);
```
budzworthy: What's the ROI of expanding the FE validation logic, when this is fairly trivial to bypass? I was assuming that most of this logic would live in the signing service, no?

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,51 @@
...
+
+type QueryResult = {
+  responseData?: GetEntitlementsResponse;
+};
+
+const useEntitlement: UseEntitlement = (mediaOffers = [], enabled = true) => {
+  const sandbox = useConfigStore(({ config }) => config?.cleengSandbox);
+  const jwt = useAccountStore(({ auth }) => auth?.jwt);
+
+  const entitlementQueries = useQueries(
```
budzworthy: Won't this result in caching the entitlement results even after potentially the user has just purchased an entitlement? I feel like there needs to be some invalidation logic somewhere?

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,51 @@
...
+      queryFn: () => getEntitlements({ offerId: offerId || '' }, sandbox, jwt || ''),
+      enabled: enabled && !!jwt && !!offerId,
+    })),
+  );
+
+  const evaluateEntitlement = (queryResult: QueryResult) => !!queryResult?.responseData?.accessGranted;
+
+  return useMemo(
+    () =>
+      entitlementQueries.reduce<UseEntitlementResult>(
```
budzworthy: Not sure I agree with the approach of using reduce here. For example, if their are 3 queries, 1 has an error, 1 is successful, and one hangs for a long time for whatever reason  - this will lead to a result that indicates an error and shows a spinner for a long time.  In reality, we can probably optimize to return isEntitled as soon as we get any 1 query returned without error with accessGranted = true.  Thoughts?

src/hooks/useEntitlement.ts
royschut: Yes. I haven't yet implemented TVOD purchases, but it requires an invalidation mechanism indeed. I'll implement that from another branch!

src/hooks/useEntitlement.ts
royschut: Hm, that's correct. I figured we want to return only one object that represents all queries to keep its implementation simple. So any `isLoading` or `error` would work for the total... But you're right that any accessGranted should overrule it. I could filter out the `accessGranted` first and return it if found, and otherwise return the reduced result.
Thanks!

src/screens/Movie/Movie.tsx
royschut: With "signing service" you mean the useEntitlement() hook? It's currently only used for TVOD, so perhaps I should rename it (`useMediaOffers()` ?) and create another hook that combines all access logics. Or am I getting your question wrong?

src/screens/Movie/Movie.tsx
```diff
@@ -49,23 +50,31 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
   const [hasShared, setHasShared] = useState<boolean>(false);
   const [playTrailer, setPlayTrailer] = useState<boolean>(false);
 
-  // User
+  // User, accessModel, entitlement
   const { user, subscription } = useAccountStore(({ user, subscription }) => ({ user, subscription }), shallow);
-  const allowedToWatch = isAllowedToWatch(accessModel, !!user, itemRequiresSubscription, !!subscription);
+  const isItemFree = item?.requiresSubscription === 'false' || !!item?.free;
+
+  const mediaOffers = useMemo(() => filterCleengMediaOffers(item?.productIds), [item]);
```
budzworthy: Instead of adding another util method, would it make sense to just put all of this logic into `isAllowedToWatch`?  Is there a case where we would use the filter method on its own?

src/utils/cleeng.ts
```diff
@@ -1,11 +1,29 @@
...
+
+  return false;
+};
+
+export const filterCleengMediaOffers = (offerIds: string = ''): MediaOffer[] => {
+  return offerIds
+    .split(',')
+    .reduce<MediaOffer[]>(
+      (offers, offerId) =>
+        offerId?.indexOf('cleeng:') === 0 ? [...offers, { offerId: offerId.slice(offerId.indexOf(':')), forced: offerId[0] === '!' }] : offers,
```
budzworthy: Would be nice to add some comments explaining what this is doing

src/screens/Movie/Movie.tsx
```diff
@@ -49,23 +50,31 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
 
   // General state
   const [hasShared, setHasShared] = useState<boolean>(false);
   const [playTrailer, setPlayTrailer] = useState<boolean>(false);
 
-  // User
+  // User, accessModel, entitlement
   const { user, subscription } = useAccountStore(({ user, subscription }) => ({ user, subscription }), shallow);
-  const allowedToWatch = isAllowedToWatch(accessModel, !!user, itemRequiresSubscription, !!subscription);
+  const isItemFree = item?.requiresSubscription === 'false' || !!item?.free;
```
budzworthy: Personally, I think I would prefer to put this sort of basic data transformation at a lower level (in this case in the hook) so that components don't have to worry about data format, but can just directly use what they need, and it's reusable other places where we might call the hook.

src/screens/Movie/Movie.tsx
budzworthy: Maybe I'm wrong. I thought this was just going to be duplication of whatever validation checks we do on the Backend service, but I think I'm remembering that we decided to let the BE service be dumb, and the FE picks what entitlement it thinks the user can use to view the item.  Let's recap the flow in standup tomorrow.

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const useEntitlement: UseEntitlement = (mediaOffers = [], enabled = true) => {
+  const sandbox = useConfigStore(({ config }) => config?.cleengSandbox);
+  const jwt = useAccountStore(({ auth }) => auth?.jwt);
+
+  const entitlementQueries = useQueries(
+    mediaOffers.map(({ offerId }) => ({
+      queryKey: ['mediaOffer', offerId],
+      queryFn: () => getEntitlements({ offerId: offerId || '' }, sandbox, jwt || ''),
```
budzworthy: Should we just skip blank offer ID's instead of sending empty string?

src/hooks/useEntitlement.ts
budzworthy: Is the entitlement call that slow that the query caching is valuable?  Or are there other benefits that I'm not thinking of?

src/hooks/useEntitlement.ts
budzworthy: I'm not totally sure what to do with the errors. In theory an offerId could be broken and then it should just get ignored. If all of the offers are broken I guess we would just default to saying no access until the queries invalidate and try again. Should we enable retries on the queries?

src/screens/Movie/Movie.tsx
ChristiaanScheermeijer: Partly correct, but let's indeed align this tomorrow. In preparation for that a short explanation.

The media signing service (back-end) won't be dumb, however, the metadata signing service (also back-end) will be. We still need to do these checks client-side to show the correct button state. As well to still support SVOD or even TVOD scenarios without URL signing enabled. The media signing service will contain the same checks as the front-end indeed.

 I can imagine that we will eventually drop support for this and only allow SVOD and/or TVOD when having a signing service to make it more secure. But I want to make sure everyone agrees to make such a change. If we do so, we can probably keep this logic in the signing service exclusively and use that to determine if the user has access.

src/hooks/useEntitlement.ts
ChristiaanScheermeijer: I guess we don't need the caching mechanism for this indeed. @royschut we better just call the entitlement service directly without `useQueries` in this case.

src/screens/Movie/Movie.tsx
ChristiaanScheermeijer: Yes, I agree

src/hooks/useEntitlement.ts
ChristiaanScheermeijer: It is skipped due to the `enabled` property being set to `false`, we are doing this on multiple occasions but due to your comment, I guess this isn't clear.

ChristiaanScheermeijer: In this case, I think that we can ensure that all `mediaOffers` have a valid `offerId`. Then we can prevent skipping these queries entirely. According to the `MediaOffer` type, the `offerId` can only be a string. 

So technically, the or operator is redundant since it will be an empty string already. 

But I think removing invalid `offerIds` from the array is more clear.

```tsx
const useEntitlement: UseEntitlement = (mediaOffers = [], enabled = true) => {
  const sandbox = useConfigStore(({ config }) => config?.cleengSandbox);
  const jwt = useAccountStore(({ auth }) => auth?.jwt);

  // memorize since `.filter` creates a new Array 
  const validMediaOffers = useMemo(() => mediaOffers.filter(({ offerId }) => !!offerId), [mediaOffers]);

  const entitlementQueries = useQueries(
    validMediaOffers.map(({ offerId }) => ({
      queryKey: ['mediaOffer', offerId],
      queryFn: () => getEntitlements({ offerId }, sandbox, jwt || ''),
      enabled: enabled && !!jwt,
    })),
  );
```

src/hooks/useEntitlement.ts
royschut: The call is quite fast actually, so sure, I'll take `useQueries()` out.

src/hooks/useEntitlement.ts
royschut: Agreed that an error could probably be ignored since I guess the most probable cause would be older, broken offers. A retry wouldn't make sense in such cases, but perhaps it would for temporal issues? I'd hope these are edge cases though.
(React-query actually retries by default, but I just agreed with you both to take `useQueries` out. 😅 )

src/screens/Movie/Movie.tsx
royschut: I'm using the result of `filterCleengMediaOffers()` for the `useEntitlement()` hook, so I can't move it to `isAllowedToWatch()`. However, I could move both their logic to useEntitlement. In that case, the hook would handle all accessModels.

src/utils/cleeng.ts
royschut: Sure!

src/screens/Movie/Movie.tsx
royschut: Yes, I agree. It's what I meant before actually with creating a new hook.

src/hooks/useEntitlement.ts
royschut: This will change if I take out `useQueries`. I hope I don't need the extra filtering, to be honest, but I'll keep your request in mind to prevent the empty string assignment.

src/hooks/useEntitlement.ts
budzworthy: Can we usequeries without caching but with retries?

src/screens/Movie/Movie.tsx
budzworthy: The idea is that it would all go into `isAllowedToWatch` which I guess would make it into a hook itself.  Otherwise, `isAllowedToWatch` and `filterCleengMediaOffers` could probably become methods internal to the useEntitlements hook.  Basically these 3 things will always be used together, so they can all live in 1 place and ideally only expose what external callers will need.

src/hooks/useEntitlement.ts
royschut: Yes, or actually just a simple invalidation. I think I should take the user's offers from the CheckoutStore, which updates after a TVOD offer has been purchased. These 'userOffers' (so to say) should be a dependency of useQueries, so I store it within the queryKey, which means it will retrigger after update. It's basically the same as using it as a dependency for a useEffect that would call/recall the entitlement endpoint.

If you're both okay with it, I keep the useQueries, implement all other feedback and showcase what I mean in the next PR.

src/screens/Movie/Movie.tsx
budzworthy: We should suggest to @marcovandeveen dropping the support of SVOD / TVOD or even AuthVod without a signing backend. If we can get buy in there, then the logic for `isAllowedToWatch` can just come from calling the signing endpoint and checking for a 401 or 403 response.  In preparation for that, it would be nice to move all of this current logic into an isolated service that can be swapped out when a customer wants to use an actual backend. So something like `localEntitlementService`.

src/hooks/useEntitlement.ts
budzworthy: OK, the invalidation sounds like it should work. Retries are probably nice to have if it's easy enough.

src/screens/Movie/Movie.tsx
royschut: Yes, I'm implementing that. Only `isAllowedToWatch` is also used for the lock icon in other places, but I'm creating a dedicated `showLock()` util for that now.

src/screens/Movie/Movie.tsx
marcovandeveen: I agree all SVOD/TVOD/AuthVod should all move to a backend entitlement service. 

But let's not do that right now: it makes it more difficult to adopt/try out the web app. A customer/partner doing a POC would need URL signing and a backend service of Videodock to test out these key use cases.

Our current plan
1. Short-term: DYI or Videdock hosting 
2. Mid-term: opensource entitlement service
3. Long-term:  JW platform-based entitlement service

I like the suggestion of having a localEntitlementService. It would also be used as a reference for the implementation of the backend logic. Maybe that's even easier, given that the Videodock entitlement service is written in node.js?

royschut: Great. I've made a note of this, thanks!

src/hooks/useEntitlement.ts
royschut: So as discussed below, we continue with `useQueries` for now and try out a proper invalidation first.

src/hooks/useEntitlement.ts
royschut: Yes okay, let's evaluate the invalidation from the next PR which includes purchasing the media offers.

ChristiaanScheermeijer (approved):

PR #66, opened 2022-05-19
Feat / tvod

royschut:
## TVOD

This PR implements the Transactional Video On Demand feature:
- Implements the Cleeng entitlement endpoint
- Adds the `useEntitlement()` hook to the Movie and Series screens:
  - Checks if the user is entitled within all accessModels
  - For TVOD (if the item has offers) the new endpoint is called
- Replaces the `isAllowedToWatch()` utility for `showLock` to show and hide the lock icon on cards, which takes TVOD in account, but without entitlement, since calling the entitlement endpoint for all items would be very costly.
  - In an AuthVOD model all items are locked until logged in, and then only items with offers have a lock
  - In an SVOD model all items are locked until subscribed, and then only premium items have a lock
- Adds TVOD to the Choose Offer modal
- Adds TVOD to the Checkout modal
  - Within the AuthVOD model TVOD is supported as well, in which case no subscriptions are shown in the Choose Offer model, but only media offers.

Implements #25 

### Some considerations
- Currently the user's transactions are used as a re-trigger for the entitlement after a purchase. This works, but I'm open for alternative suggestions
- AVOD + TVOD is currently not supported, because we define the model to be AVOD when no cleengId is set, but the cleengId is needed for TVOD.
- Do we need to implement additional logics for series? Currently, episodes are checked for entitlement separately, which means the offer has to be added to all episodes in order to sell the series as a whole. We could consider implementing logics to check the offer of the linking item, which represents the full series. (This does require some additional logics in the backend service as well)
- I could extract the pure logics from the `useEntitlement()` hook to a `localEntitlementService`, but I doubt if it can become an identical local copy of the backend version
- If better names for the CheckoutStore property `requestedMediaOffers` come to mind, please let me know!


src/components/ChooseOfferForm/ChooseOfferForm.tsx
```diff
@@ -24,107 +24,156 @@ type Props = {
...
-  const getFreeTrialText = (offer: Offer) => {
-    if (offer.freeDays > 0) {
-      return t('choose_offer.benefits.first_days_free', { count: offer.freeDays });
-    } else if (offer.freePeriods) {
-      // t('periods.day')
-      // t('periods.week')
-      // t('periods.month')
-      // t('periods.year')
-      const period = t(`periods.${offer.period}`, { count: offer.freePeriods });
+  const getFreeTrialText = useCallback(
```
ChristiaanScheermeijer: `useCallback` is redundant here since it isn't passed as prop. When used in JSX directly, it is being called anyways.

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -12,50 +12,98 @@ import { getOffer } from '#src/services/checkout.service';
...
 import useForm, { UseFormOnSubmitHandler } from '#src/hooks/useForm';
-import type { ChooseOfferFormData, OfferPeriodicity } from '#types/account';
+import type { ChooseOfferFormData, OfferPeriodicity, OfferType } from '#types/account';
+import type { Offer } from '#types/checkout';
 
 const ChooseOffer = () => {
   const history = useHistory();
   const { t } = useTranslation('account');
-  const { config, accessModel } = useConfigStore(({ config, accessModel }) => ({ config, accessModel }), shallow);
+  const { config } = useConfigStore(({ config }) => ({ config }), shallow);
```
ChristiaanScheermeijer: I guess destructuring isn't necessary anymore, but not a big deal since `config` never changes.

```suggestion
  const config = useConfigStore(({ config }) => config);
```

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -12,50 +12,98 @@ import { getOffer } from '#src/services/checkout.service';
...
   const { cleengSandbox, json } = config;
-  const hasOffer = accessModel === 'SVOD';
-  const { offer, setOffer } = useCheckoutStore(({ offer, setOffer }) => ({ offer, setOffer }), shallow);
+  const { setOffer, setOfferType } = useCheckoutStore(({ setOffer, setOfferType }) => ({ setOffer, setOfferType }), shallow);
 
   const cleengMonthlyOffer = json?.cleengMonthlyOffer as string;
   const cleengYearlyOffer = json?.cleengYearlyOffer as string;
 
+  const { requestedMediaOffers } = useCheckoutStore(({ requestedMediaOffers }) => ({ requestedMediaOffers: requestedMediaOffers || [] }));
```
ChristiaanScheermeijer: Missing the `shallow` comparator

```suggestion
  const { requestedMediaOffers } = useCheckoutStore(({ requestedMediaOffers }) => ({ requestedMediaOffers: requestedMediaOffers || [] }), shallow);
```

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -12,50 +12,98 @@ import { getOffer } from '#src/services/checkout.service';
...
+  const { data: monthlyOfferData, isLoading: isMonthlyOfferLoading } = useQuery(['offer', cleengMonthlyOffer], () =>
+    getOffer({ offerId: cleengMonthlyOffer }, cleengSandbox),
+  );
+  const { data: yearlyOfferData, isLoading: isYearlyOfferLoading } = useQuery(['offer', cleengYearlyOffer], () =>
+    getOffer({ offerId: cleengYearlyOffer }, cleengSandbox),
+  );
+
+  const tvodOfferQueries = useQueries(
+    requestedMediaOffers?.map(({ offerId }) => ({
+      queryKey: ['requestedMediaOffer', offerId],
```
ChristiaanScheermeijer: I would keep the queryKey predictable to make it easier to invalidate (if we eventually decide to do so). Since it is getting an offer, I suggest to keep it the same e.g.

```suggestion
      queryKey: ['offer', offerId],
```

royschut: Good point!

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import type { MediaOffer } from '#types/media';
+import type { GetEntitlementsResponse } from '#types/checkout';
+import { filterCleengMediaOffers } from '#src/utils/cleeng';
+import type { PlaylistItem } from '#types/playlist';
+import { useConfigStore } from '#src/stores/ConfigStore';
+import { useAccountStore } from '#src/stores/AccountStore';
+import { getEntitlements } from '#src/services/checkout.service';
+
+export type QueriesResult = {
```
ChristiaanScheermeijer: Unused?

src/i18n/locales/en_US/account.json
```diff
@@ -52,7 +52,9 @@
     "monthly": "Monthly",
     "monthly_subscription": "Monthly subscription",
     "offer_not_found": "Offer not found, please try reloading the page",
+    "one_time_only": "One time only",
     "subscription": "Subscription",
+    "tvod_access": "{{ count }} {{ period }} access",
```
ChristiaanScheermeijer:
```suggestion
    "tvod_access": "{{count}} {{period}} access",
```

src/utils/cleeng.ts
```diff
@@ -1,11 +1,49 @@
...
+/**
+ * The appearance of the lock icon, depending on the access model
+ *
+ * @param accessModel Platform AccessModel, excluding TVOD, which can only be applied per item
+ * @param isLoggedIn
+ * @param hasSubscription
+ * @param playlistItem Used to define if the items is 'free' or has mediaOffers
+ * @returns
+ */
+
```
ChristiaanScheermeijer: -newline

types/checkout.d.ts
```diff
@@ -12,7 +12,9 @@ export type Offer = {
   discountedCustomerPriceInclTax: number | null;
   discountedCustomerPriceExclTax: number | null;
   discountPeriods: number | null;
-  offerUrl: string;
+  durationPeriod?: string | null; // Duration isn't sent from Cleeng yet
+  durationAmount?: number | null; //
```
ChristiaanScheermeijer:
```suggestion
  durationAmount?: number | null;
```

src/components/CardGrid/CardGrid.tsx
```diff
@@ -77,7 +77,7 @@ function CardGrid({
             loading={isLoading}
             isCurrent={currentCardItem && currentCardItem.mediaid === mediaid}
             currentLabel={currentCardLabel}
-            isLocked={!isAllowedToWatch(accessModel, isLoggedIn, playlistItem.requiresSubscription !== 'false', hasSubscription)}
+            isLocked={showLock(accessModel, isLoggedIn, hasSubscription, playlistItem)}
```
budzworthy: Can I suggest "isLocked" for the method name? showLock implies that it's a rendering function of some sort instead of a boolean function.

src/components/ChooseOfferForm/ChooseOfferForm.tsx
```diff
@@ -24,15 +24,37 @@ type Props = {
...
+  monthlyOffer,
+  tvodOffers,
+  onBackButtonClickHandler,
+}: Props) => {
   const { siteName } = useContext(ConfigContext);
   const { t } = useTranslation('account');
 
   const getFreeTrialText = (offer: Offer) => {
-    if (offer.freeDays > 0) {
+    if (offer.freeDays && offer.freeDays > 0) {
```
budzworthy: Are these checks because freeDays was negative? Does it make sense to create an extension or util method to check this instead of duplicating the logic?

src/components/CheckoutForm/CheckoutForm.tsx
```diff
@@ -30,6 +32,7 @@ type Props = {
   couponInputValue: string;
   order: Order;
   offer: Offer;
+  offerType: OfferType;
```
budzworthy, line 35: Having both offer and offerType separately is confusing to me. Can we add offerType as a property of `Offer` (realizing that we would probably have to add it to the cleeng response) or better yet is there a way for us to use existing properties of Offer to determine what the offerType is?

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -12,50 +12,98 @@ import { getOffer } from '#src/services/checkout.service';
...
   useEffect(() => {
     // close auth modal when there are no offers defined in the config
-    if (!hasOffer) history.replace(removeQueryParam(history, 'u'));
-  }, [hasOffer, history]);
+    if (!isOffersLoading && !hasOffer) history.replace(removeQueryParam(history, 'u'));
+  }, [isOffersLoading, hasOffer, history]);
 
-  const chooseOfferSubmitHandler: UseFormOnSubmitHandler<ChooseOfferFormData> = async (formData, { setSubmitting, setErrors }) => {
-    const offer = formData.periodicity === 'monthly' ? monthlyOfferData?.responseData : yearlyOfferData?.responseData;
+  const chooseOfferSubmitHandler: UseFormOnSubmitHandler<ChooseOfferFormData> = async (
```
budzworthy: There is a weird mix of periodicity and ID's used here. Is it possible to just use ID's?

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -12,50 +12,98 @@ import { getOffer } from '#src/services/checkout.service';
...
-  const { data: monthlyOfferData } = useQuery(['offer', cleengMonthlyOffer], () => getOffer({ offerId: cleengMonthlyOffer }, cleengSandbox));
-  const { data: yearlyOfferData } = useQuery(['offer', cleengYearlyOffer], () => getOffer({ offerId: cleengYearlyOffer }, cleengSandbox));
+  const { data: monthlyOfferData, isLoading: isMonthlyOfferLoading } = useQuery(['offer', cleengMonthlyOffer], () =>
+    getOffer({ offerId: cleengMonthlyOffer }, cleengSandbox),
+  );
+  const { data: yearlyOfferData, isLoading: isYearlyOfferLoading } = useQuery(['offer', cleengYearlyOffer], () =>
+    getOffer({ offerId: cleengYearlyOffer }, cleengSandbox),
+  );
+
+  const tvodOfferQueries = useQueries(
```
budzworthy: Can we just append the yearly and monthy offer ID's to the tvod offers and do 1 big useQueries call?

src/components/ChooseOfferForm/ChooseOfferForm.tsx
```diff
@@ -47,84 +69,108 @@ const ChooseOfferForm: React.FC<Props> = ({ values, errors, onChange, onSubmit,
...
-              onChange={onChange}
-              type="radio"
-              name="periodicity"
-              value="monthly"
-              id="monthly"
-              checked={values.periodicity === 'monthly'}
-              aria-label={t('choose_offer.monthly_subscription')}
+      {!!tvodOffers.length && (!!yearlyOffer || !!monthlyOffer) && (
+        <div className={styles.offerGroupSwitch}>
+          <input className={styles.radio} onChange={onChange} type="radio" name="offerType" id="svod" value="svod" checked={values.offerType === 'svod'} />
```
budzworthy: I'm confused about what `values` represent. Is it state data of the form? The form flow? Maybe it can be called "checkoutState" or something?  And if all of this is really local state to the checkout flow (e.g. to keep track of what offer, payment method, etc. was seleceted), can it be in a useState hook on the checkout components instead of going all the way up to the store?

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const { isMediaEntitled, isMediaEntitlementLoading } = useMemo(() => {
+    const isEntitled = mediaEntitlementQueries.some((item) => item.isSuccess && (item.data as QueryResult)?.responseData?.accessGranted);
+
+    return { isMediaEntitled: isEntitled, isMediaEntitlementLoading: !isEntitled && mediaEntitlementQueries.some((item) => item.isLoading) };
+  }, [mediaEntitlementQueries]);
+
+  const isEntitled = useMemo(() => {
+    if (isItemFree) return true;
+    if (accessModel === 'AVOD' && (!mediaOffers.length || isMediaEntitled)) return true;
```
budzworthy: A single line comment on each if here would be great

src/utils/cleeng.ts
```diff
@@ -1,11 +1,49 @@
...
+/**
+ * Filters Cleeng MediaOffers from offers string
+ *
+ * @param offerIds String of comma separated key/value pairs, i.e. "cleeng:S916977979_NL, !cleeng:S91633379_NL, other_vendor:xyz123"
+ * Key is vendor, value is the offerId.
+ * Vendor keys starting with an exclamation mark represent a 'Premier Access' offer (TVOD only)
+ *
+ * @returns An array of MediaOffer { offerId, premier }
+ */
+export const filterCleengMediaOffers = (offerIds?: string): MediaOffer[] | null => {
```
budzworthy: It might be nice to add this on PlaylistItem and parse into a dictionary of provider and offers, instead of making it too cleeng specific.

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,64 @@
...
+    })),
+  );
+
+  const { isMediaEntitled, isMediaEntitlementLoading } = useMemo(() => {
+    const isEntitled = mediaEntitlementQueries.some((item) => item.isSuccess && (item.data as QueryResult)?.responseData?.accessGranted);
+
+    return { isMediaEntitled: isEntitled, isMediaEntitlementLoading: !isEntitled && mediaEntitlementQueries.some((item) => item.isLoading) };
+  }, [mediaEntitlementQueries]);
+
+  const isEntitled = useMemo(() => {
```
budzworthy: Is useMemo relevant when you are returning true/false?

src/hooks/useEntitlement.ts
```diff
@@ -0,0 +1,64 @@
...
+      queryFn: () => getEntitlements({ offerId }, sandbox, jwt || ''),
+      enabled: !!playlistItem && !!jwt && !!offerId && !skipMediaEntitlement,
+    })),
+  );
+
+  const { isMediaEntitled, isMediaEntitlementLoading } = useMemo(() => {
+    const isEntitled = mediaEntitlementQueries.some((item) => item.isSuccess && (item.data as QueryResult)?.responseData?.accessGranted);
+
+    return { isMediaEntitled: isEntitled, isMediaEntitlementLoading: !isEntitled && mediaEntitlementQueries.some((item) => item.isLoading) };
+  }, [mediaEntitlementQueries]);
```
budzworthy: Is this going to update when properties of the mediaEntitlementQueries change (i.e. queries finish)?

src/hooks/useEntitlement.ts
budzworthy: I think you can move these if's above and just return `{isEntitled: true}` for the first 3 cases. Then for the fourth case you can start the queries (instead of the duplicated skip logic above)

src/utils/cleeng.ts
```diff
@@ -1,11 +1,49 @@
 import type { AccessModel } from '../../types/Config';
```
budzworthy: This file doesn't really have much cleeng specific logic in it. Maybe it's better renamed as `utils/entitlements.ts`?

src/screens/Series/Series.tsx
```diff
@@ -97,18 +96,21 @@ const Series = ({ match, location }: RouteComponentProps<SeriesRouteParams>): JS
     return nextItem && history.push(episodeURL(seriesPlaylist, nextItem.mediaid, true));
   }, [history, item, seriesPlaylist]);
 
-  const formatStartWatchingLabel = (): string => {
-    if (!allowedToWatch && !user) return t('sign_up_to_start_watching');
-    if (!allowedToWatch && !subscription) return t('complete_your_subscription');
-    return typeof progress === 'number' ? t('continue_watching') : t('start_watching');
-  };
+  const startWatchingLabel = useMemo((): string => {
```
budzworthy: Can we move the label logic to a util since it's reused and getting more complex?

src/utils/cleeng.ts
```diff
@@ -1,11 +1,49 @@
...
+ * @param hasSubscription
+ * @param playlistItem Used to define if the item is 'free' or has mediaOffers
+ * @returns
+ */
+
+export const showLock = (accessModel: AccessModel, isLoggedIn: boolean, hasSubscription: boolean, playlistItem: PlaylistItem): boolean => {
+  const isItemFree = playlistItem?.requiresSubscription === 'false' || !!playlistItem?.free;
+  const mediaOffers = filterCleengMediaOffers(playlistItem?.productIds);
+
+  if (isItemFree) return false;
```
budzworthy: This is very similar logic to the useEntitlement hook. Maybe they can share some logic so they don't risk getting out of sync?

src/utils/cleeng.ts
```diff
@@ -1,11 +1,49 @@
...
+  if (accessModel === 'AVOD' && !mediaOffers) return false;
+  if (accessModel === 'AUTHVOD' && isLoggedIn && !mediaOffers) return false;
+  if (accessModel === 'SVOD' && hasSubscription && !mediaOffers?.some((offer) => offer.premier)) return false;
+
+  return true;
```
budzworthy: Does this default to showing the lock icon for premier and tvod videos all the time? That's probably not a really great user experience. Since we only need to do entitlements checks for premier and tvod videos, would it be that bad of a performance hit if we just used the useEntitlements hook to check this (with caching)? I would imagine in most cases there will only be a handful of videos that we need to check entitlements for and if it happens on the main page in the background, it won't be a huge performance hit and will make the experience faster and more accurate.

src/services/checkout.service.ts
```diff
@@ -1,8 +1,17 @@
-import type { CreateOrder, GetOffer, GetPaymentMethods, PaymentWithAdyen, PaymentWithoutDetails, PaymentWithPayPal, UpdateOrder } from '../../types/checkout';
```
budzworthy, line 1: Would it make sense to rename this to entitlementsService? Orders, Offers, and to some degree payments are all Entitlement related. (I would argue payments should be split off to separate services in the future anyhow.)

src/components/ChooseOfferForm/ChooseOfferForm.tsx
```diff
@@ -47,84 +69,108 @@ const ChooseOfferForm: React.FC<Props> = ({ values, errors, onChange, onSubmit,
...
+          <label className={classNames(styles.label, styles.offerGroupLabel)} htmlFor="svod">
+            {t('choose_offer.subscription')}
+          </label>
+          <input className={styles.radio} onChange={onChange} type="radio" name="offerType" id="tvod" value="tvod" checked={values.offerType === 'tvod'} />
+          <label className={classNames(styles.label, styles.offerGroupLabel)} htmlFor="tvod">
+            {t('choose_offer.one_time_only')}
+          </label>
+        </div>
+      )}
+      {values.offerType === 'svod' ? (
```
budzworthy: It's sort of hard to keep track of the ternary operator here. Maybe it would be easier to read if it was just 2 blocks, the first with `values.offerType === 'svod' && ...` and the second `values.offerType === 'tvod' && ...`

budzworthy (changes requested):
There are a few things I think need a bit more thought before we merge this:

- Can we minimize (or at least avoid adding more) checkout UI state data to the store?
- The useEntitlement hook can be simplified and cleaned up
- It feels like the tvod and svod offers don't 'play nicely'. For example the svod offers are referenced by period while tvod are referenced by ID and the queries to look them up can probably be combined

src/components/CardGrid/CardGrid.tsx
royschut: I agree, thanks!

src/components/ChooseOfferForm/ChooseOfferForm.tsx
royschut: `freeDays` can actually be `null` (I had to update the types), in which case the `> 0` isn't valid. 
I could change it to `Number(offer.freeDays) > 0`, if you find that more elegant, since `Number(null)` casts to 0.

src/components/CheckoutForm/CheckoutForm.tsx
royschut: I could use `offer.period` to define the type actually ('month', 'year', null). That would keep the CheckoutStore cleaner.
I'd still like to use the `OfferType` type within the CheckoutForm though (`const offerType: OfferType = ..`), to keep the JSX readable. What do you think?

src/containers/AccountModal/forms/ChooseOffer.tsx
royschut: I see your point, also about the mix of queries. I'll see if I can merge the two offer types a bit more, without losing the readability/understandability for the regular SVOD situation (which might still be more common).

src/components/ChooseOfferForm/ChooseOfferForm.tsx
royschut: That's indeed the result of the useForm hook, like we use elsewhere as well, which I wouldn't want to replace. I could change the variable name to `checkoutValues` or `checkoutState`, but then it would not match other forms anymore. 🤔

src/hooks/useEntitlement.ts
royschut: Sure, let's make it very clear indeed.

src/components/ChooseOfferForm/ChooseOfferForm.tsx
budzworthy: Why not just `if (offer.freeDays)`?

src/utils/cleeng.ts
royschut: I've added `productIds` to `PlaylistItem`, since it's an optional property now. The MediaOffers (parsed) are not, so I don't think we should add them to PlaylistItem.
I've made it cleeng specific because it's the only key we 'care' about here. Not sure if we should make that more generic. Any more concrete ideas?

src/hooks/useEntitlement.ts
royschut: Good point, perhaps not.

src/hooks/useEntitlement.ts
royschut: In my experience it did, but I'll re-check this.

src/hooks/useEntitlement.ts
royschut: I can try that, but I'm not sure if it makes it more readable. I'll drop a snippet once I get to it.

src/utils/cleeng.ts
royschut: Agreed! (Although I'd prefer to keep the `filterCleengMediaOffers()` as is)

src/screens/Series/Series.tsx
royschut: I could, but generating translations from a util can be hard to find sometimes. Maybe it gets clear when added to a `utils/translation.ts`?

src/utils/cleeng.ts
royschut: Good point, although they are not identical, because the other version takes entitlement in account. I'll see if I can extract the duplicated part.

src/components/CheckoutForm/CheckoutForm.tsx
budzworthy: Sure, maybe via a util method?

src/utils/cleeng.ts
royschut: Hm, that's quite a big one. Other than the potential performance risks, it also requires quite some additional work, which is why I mentioned it during the kick-off of the sprint. 

But let's consider what we truly wish for first. The current situation is as described in the PR description:

> In an AuthVOD model all items are locked until logged in, and then all tvod items have a lock
> In an SVOD model all items are locked until subscribed, and then only **premium** items have a lock

So it's a bit more specific than you described. 
What are your thoughts regarding this @ChristiaanScheermeijer @dbudzins @marcovandeveen?

src/services/checkout.service.ts
royschut: I'm not sure. It might be semantic, but "entitlement" seems to me like the check before accessing content, where "checkout" would define an actual purchase.
The CheckoutStore and CheckoutController are currently not used for the entitlement process, only for checkout purposes.

src/components/ChooseOfferForm/ChooseOfferForm.tsx
budzworthy: Oh, I see. Not sure why I thought this was coming all the way from the store.  This is OK

src/components/ChooseOfferForm/ChooseOfferForm.tsx
royschut: Yes, I have no preference here, so I can do that.

Thanks for all your feedback @dbudzins!

src/utils/cleeng.ts
budzworthy: Sure, if we end up with more providers, we can add the provider name as a param or something. It's probably overkill to abstract to that level now.

src/utils/cleeng.ts
budzworthy: What bothers me here is the repeated parsing for every lock and entitlement check. It's not a huge overhead, but it would be nice to do it once and remember it. The usages in the useEntitlement calls are all memoized but the showLock calls aren't.  We could do any of the following:

1. Make showLock into a hook and add memoization. Slightly more efficient, but still code to sync between showLock and useEntitlement
2. Combine showLock into useEntitlements with a param for whether or not to do the queries. Everything memoized and no code duplication, but params is kind of code smell.
3. Just call useEntitlements for everything and let memo and react query caching store results. Efficient and no duplication, but could be performance issues if lots of videos need querying.

src/services/checkout.service.ts
budzworthy: OK, naming is fine to stay as is. Neither is a perfect fit and naming is hard. Maybe we'll find the right name sometime down the road.

src/utils/cleeng.ts
budzworthy: I think this is ultimately a product decision, but I know if I was a customer and paid for a video, then still saw a lock icon on it, I would think something was broken. What if instead of a lock, we just show a "TVOD" or "premium" icon on the cards and shelf, and then we do the entitlements calls on the detail page as you currently have it?

src/screens/Series/Series.tsx
budzworthy: You could put it into a component then?

src/components/ChooseOfferForm/ChooseOfferForm.tsx
royschut: "Object is possibly 'null'"

Screenshot 2022-05-24 at 11 05 17

src/screens/Series/Series.tsx
royschut: Good idea!

src/components/ChooseOfferForm/ChooseOfferForm.tsx
budzworthy: Without the `> 0`

royschut: Excuse me, of course!

src/components/CheckoutForm/CheckoutForm.tsx
royschut: I've moved it to a new `useOffers()` hook, which hopefully makes things a bit more readable.

src/hooks/useEntitlement.ts
royschut: It actually makes the code more readable, and prevents having to work with `let`. Is there a notable downside?

src/utils/cleeng.ts
royschut: Yes, I see what you mean. In that case, I think your original proposal to add the offers to PlaylistItem is best, parsing it straight after fetching.

src/hooks/useEntitlement.ts
royschut: Yes, confimed

src/utils/cleeng.ts
budzworthy: It's probably good to start decoupling the raw cleeng data formats from the data formats we use in the app logic anyhow.

src/hooks/useEntitlement.ts
royschut: I combined it with the isLocked logics, as you suggested elsewhere, which makes the hook much cleaner. Thanks!

src/screens/Series/Series.tsx
royschut: I've made it into a hook, so I can also add the click handler, which was also reused between `<Movie />` and `<Series />`

src/utils/cleeng.ts
royschut: Yes, I was actually discussing the same with @ChristiaanScheermeijer. I guess this is a good starting point.

src/components/ChooseOfferForm/ChooseOfferForm.tsx
```diff
@@ -69,23 +72,25 @@ const ChooseOfferForm: React.FC<Props> = ({
     return null;
   };
 
-  const OfferBox: OfferBox = ({ offer, radioKey, radioValue, title, ariaLabel, secondBenefit, periodString }) => (
+  console.info(values.offerId);
```
budzworthy: Make sure we take this out before merging

src/hooks/usePlaylist.ts
```diff
@@ -16,9 +17,20 @@ export default function usePlaylist(
...
   limit?: number,
 ): UsePlaylistResult {
-  return useQuery(['playlist', playlistId, relatedMediaId], () => getPlaylistById(playlistId, relatedMediaId, limit), {
-    enabled: !!playlistId && enabled,
-    placeholderData: usePlaceholderData ? placeholderData : undefined,
-    retry: false,
-  });
+  return useQuery(
+    ['playlist', playlistId, relatedMediaId],
+    async () => {
```
budzworthy: You can just add `transformMediaItem` in the `.then` of getPlaylistById

src/hooks/useStartWatchingButton.ts
```diff
@@ -0,0 +1,44 @@
...
+  hasMediaOffers: boolean,
+  isLoggedIn: boolean,
+  videoProgress?: number,
+  videoUrl?: string | null,
+) => {
+  startWatchingLabel: string;
+  handleStartWatchingClick: () => void;
+};
+
+const useStartWatchingButton: UseStartWatchingLabel = (isEntitled, hasMediaOffers, isLoggedIn, videoProgress, videoUrl) => {
```
budzworthy, line 18: Couldn't you have reused the click handler in a component too?  And then you have the benefit of encapsulating the button logic in 1 place and not needing to pass properties.

src/utils/media.ts
```diff
@@ -37,3 +39,12 @@ export const getSeriesIdFromEpisode = (item: PlaylistItem) => {
 
   return null;
 };
+
+// Transfrom media items
+// Parses productId into MediaOffer[] for all cleeng offers
+export const transformMediaItem = (item: PlaylistItem) => {
```
budzworthy: Recommend putting this inside `api.service.ts`

src/hooks/useMedia.ts
```diff
@@ -1,14 +1,24 @@
...
 
 export default function useMedia(mediaId: string, enabled: boolean = true): UseMediaResult {
-  return useQuery(['media', mediaId], () => getMediaById(mediaId), {
-    enabled: !!mediaId && enabled,
-    keepPreviousData: true,
-  });
+  return useQuery(
+    ['media', mediaId],
+    async () => {
+      const media = await getMediaById(mediaId);
```
budzworthy: transformMediaItem can go into the .then inside getMediaById

src/hooks/useOffers.ts
```diff
@@ -0,0 +1,55 @@
...
+      queryKey: ['offer', offerId],
+      queryFn: () => getOffer({ offerId }, cleengSandbox),
+      enabled: !!offerId,
+    })),
+  );
+
+  const isLoadingOffers = useMemo(() => offerQueries.some(({ isLoading }) => isLoading), [offerQueries]);
+
+  const offers = useMemo(
+    () => offerQueries.reduce<Offer[]>((prev, cur) => (cur.isSuccess && cur.data?.responseData ? [...prev, cur.data.responseData] : prev), []),
```
budzworthy: Maybe more efficient to transform offers into a dictionary with ID keys instead of an array.

src/hooks/useOffers.ts
```diff
@@ -0,0 +1,55 @@
...
+    [offerQueries],
+  );
+  const tvodOffers = useMemo(() => offers.filter((offer) => offer.period === null), [offers]);
+  const hasPremierOffer = useMemo(() => !!requestedMediaOffers.some((offer) => offer.premier), [requestedMediaOffers]);
+  const monthlyOffer = useMemo(() => (!hasPremierOffer ? offers.find((offer) => offer.period === 'month') : undefined), [hasPremierOffer, offers]);
+  const yearlyOffer = useMemo(() => (!hasPremierOffer ? offers.find((offer) => offer.period === 'year') : undefined), [hasPremierOffer, offers]);
+
+  return { tvodOffers, monthlyOffer, yearlyOffer, hasPremierOffer, isLoadingOffers };
```
budzworthy: I would return all Offers here instead of by type

src/hooks/useOffers.ts
```diff
@@ -0,0 +1,55 @@
...
+  );
+
+  const isLoadingOffers = useMemo(() => offerQueries.some(({ isLoading }) => isLoading), [offerQueries]);
+
+  const offers = useMemo(
+    () => offerQueries.reduce<Offer[]>((prev, cur) => (cur.isSuccess && cur.data?.responseData ? [...prev, cur.data.responseData] : prev), []),
+    [offerQueries],
+  );
+  const tvodOffers = useMemo(() => offers.filter((offer) => offer.period === null), [offers]);
+  const hasPremierOffer = useMemo(() => !!requestedMediaOffers.some((offer) => offer.premier), [requestedMediaOffers]);
```
budzworthy: If there's a premium offer should we just return that instead of returning any others? (yearly / monthly subscriptions won't allow a user to watch that video anyhow)

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -1,109 +1,71 @@
...
+    offerId: mixed<string>().required(t('choose_offer.field_required')),
+  });
 
-  const tvodOffers = useMemo(
-    () => tvodOfferQueries.reduce<Offer[]>((prev, cur) => (cur.isSuccess && cur.data?.responseData ? [...prev, cur.data.responseData] : prev), []),
-    [tvodOfferQueries],
-  );
-  const premierOfferId = requestedMediaOffers.some((offer) => offer.premier);
+  const initialValues: ChooseOfferFormData = {
+    offerId: yearlyOffer?.offerId || monthlyOffer?.offerId || allOffers?.[0]?.offerId,
```
budzworthy: I think this logic could be better put into the useOffers hook. Return a "defaultOfferId" property. Then we don't need to return as much other detail from the hook.

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -1,109 +1,71 @@
...
-  const tvodOfferQueries = useQueries(
-    requestedMediaOffers?.map(({ offerId }) => ({
-      queryKey: ['offer', offerId],
-      queryFn: () => getOffer({ offerId }, cleengSandbox),
-      enabled: !!offerId,
-    })),
-  );
+  const { setOffer } = useCheckoutStore(({ setOffer }) => ({ setOffer }), shallow);
+  const { tvodOffers, yearlyOffer, monthlyOffer, hasPremierOffer, isLoadingOffers } = useOffers();
+  const allOffers = useMemo(() => [yearlyOffer, monthlyOffer, ...tvodOffers], [yearlyOffer, monthlyOffer, tvodOffers]);
```
budzworthy: Kind of weird how we split the offers in the hook and then immediately recombine them here

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -1,109 +1,71 @@
...
+  const [offerType, updateOfferType] = useState<OfferType>('svod');
```
budzworthy: This and the related offer type logic could also go into the useOffers hook. I think you're running into a bit of extra complexity because you have half the logic in the hook and half still in the component here. Putting it all in one place or the other should make the whole solution come together nicely (I don't have a strong preference which place, since the hook isn't reused)

src/hooks/usePlaylist.ts
ChristiaanScheermeijer: I've moved the transformers to the API service instead

src/components/ChooseOfferForm/ChooseOfferForm.tsx
ChristiaanScheermeijer: Removed, thanks!

src/hooks/useStartWatchingButton.ts
ChristiaanScheermeijer: I've created a StartWatchingButton container which now contains all the logic (not a hook for now).

src/utils/media.ts
ChristiaanScheermeijer: Done

src/hooks/useMedia.ts
ChristiaanScheermeijer: I've moved the transformers to the API service instead

src/hooks/useOffers.ts
ChristiaanScheermeijer: I've refactored this quite a bit, what do you think?

src/hooks/useOffers.ts
ChristiaanScheermeijer: Now the `useOffers` hook returns the offers based on the `offerType`. This is also moved inside the hook to keep it all together.

Though, since `offerType` was initially a form state/value, it does feel a bit strange.

src/containers/AccountModal/forms/ChooseOffer.tsx
ChristiaanScheermeijer: Done

src/containers/AccountModal/forms/ChooseOffer.tsx
ChristiaanScheermeijer: This is refactored as well. This one did throw me back a bit since all logic and components relied on these "separate offers".

src/containers/AccountModal/forms/ChooseOffer.tsx
ChristiaanScheermeijer: Yes, done!

https://github.com/jwplayer/ott-web-app/pull/66#discussion_r883557853

src/hooks/useOffers.ts
ChristiaanScheermeijer: Needed both actually 😄

src/components/Video/Video.tsx
```diff
@@ -127,23 +120,7 @@ const Video: React.FC<Props> = ({
...
-              active={play}
-              fullWidth={breakpoint < Breakpoint.md}
-            >
-              {progress ? (
-                <div className={styles.progressRail}>
-                  <div className={styles.progress} style={{ width: `${progress * 100}%` }} />
-                </div>
-              ) : null}
-            </Button>
+            {startWatchingButton}
```
budzworthy, line 123: Can't you just put the `<StartWatchingButton >` component here directly? What's the point of passing it in as a prop?

budzworthy (approved):
A couple small notes, but nothing blocking. Thanks for the finishing touches!

src/components/Video/Video.tsx
ChristiaanScheermeijer: Because it's a container and `<Video />` is a component

budzworthy: I don't really understand the difference in the convention then I guess. I would vote that the simplicity of just having the button in the jsx outweighs any sort of nuance with trying to limit the amount of logic inside of a component. Or is there something I'm missing?

PR #70, opened 2022-05-23
Feat / add playlist and media entitlement

### LINKED ISSUE #23: VOD DRM Support
Many OTT implementations have licensed content, which comes with the obligation to apply DRM .

## Scope
- Studio DRM with JW Platform for VOD content 
- Update the features docs on video protection located at /docs/features
- JW platform 
- DRM policy 'low' (see [here](https://developer.jwplayer.com/jwplayer/docs/studio-drm-with-jw-platform-policy-reference#default-drm-policies) ), but with 'Max video resolution width': 3840 pixels. 

## Out-of-scope
- Opensource a URL signing solution
- DRM for Vualto Live streams
- DRM for other platforms

## JW Implementation steps
1.  Upload media to a DRM enabled property
2. Implement URL signing
3. Have the mobile and TV applications call the JW Delivery API. 

Detailed information can be found here:
- https://developer.jwplayer.com/jwplayer/docs/studio-drm-getting-started
- https://developer.jwplayer.com/jwplayer/docs/studio-drm-web-protect-your-content-with-studio-drm-with-stream
- https://developer.jwplayer.com/jwplayer/docs/studio-drm-web-protect-your-content-with-studio-drm-with-stream#intermediary-service
 
## Notes
- Testing Apple FairPlay on Safari requires Fairplay credentials. This process should be started at least 2 weeks before the start of implementation. 
- GET /v2/playlists/{playlist_id} doesn't work - Signing 
- Trailer needs to marked as `trailer` as these are freely accessible
- Chromecast doesn't work. Needs custom receiver

ChristiaanScheermeijer:
## Description

- [x] Add `contentSigningService` to config
- [x] Get a (DRM) token for all playlist calls (including search and recommendations)
- [x] Get a (DRM) token for media call on Movie and Series screens
- [x] Get a fresh (DRM) token when starting video playback
- [x] Get tokens for the `getMediaByIds`

Fixes #23


src/hooks/useMedia.ts
```diff
@@ -1,14 +1,18 @@
...
-import type { PlaylistItem } from '#types/playlist';
+export default function useMedia(mediaId: string, enabled: boolean = true) {
+  const { token, signingEnabled, drmEnabled, drmPolicyId, isLoading } = useSignedUrl('media', mediaId, {}, enabled);
 
-export type UseMediaResult<TData = PlaylistItem, TError = unknown> = UseBaseQueryResult<TData, TError>;
-
-export default function useMedia(mediaId: string, enabled: boolean = true): UseMediaResult {
-  return useQuery(['media', mediaId], () => getMediaById(mediaId), {
-    enabled: !!mediaId && enabled,
+  const queryResult = useQuery(['media', mediaId, token], () => getMediaById(mediaId, token, drmEnabled ? drmPolicyId : undefined), {
```
budzworthy: What about putting the `getMediaById` or `getPlaylistById` calls into `useSignedUrl` as callbacks so there isn't as much duplicated code in the boilerplate (the second useQuery) for `useMedia` and `usePlaylist`?

src/hooks/useSignedUrl.ts
```diff
@@ -0,0 +1,32 @@
+import { useQuery } from 'react-query';
```
budzworthy: What about `useContentProtection` as a name for this hook?  It doesn't take or return a url as a parameter so the url naming is a bit misleading.

src/hooks/useSignedUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+      if (!!id && !!signingConfig?.host && signingConfig?.drmEnabled && !!signingConfig?.drmPolicyId) {
+        const { host, drmPolicyId } = signingConfig;
+
+        return getPublicToken(host, type, id, jwt, params, drmPolicyId);
+      }
+    },
+    { enabled: signingEnabled && enabled && !!id, keepPreviousData: false, staleTime: 15 * 60 * 1000 },
+  );
+
+  return { token, signingEnabled, drmEnabled, drmPolicyId, isLoading };
```
budzworthy: Is there a bug with the return properties if `signingEnabled` is true, but the `enabled` param is false?

src/providers/QueryProvider.tsx
```diff
@@ -4,7 +4,7 @@ import { QueryClient, QueryClientProvider } from 'react-query';
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      staleTime: 3600,
+      staleTime: 60 * 60 * 1000,
```
budzworthy, line 7: Good catch with the missing ms multiplier. Do you think an hour is a good default though or potentially long for all but the most basic config values?

ChristiaanScheermeijer: Good question, the default `staleTime` is 5 minutes which is a bit short in my opinion. The data doesn't change that often. 

It does provide a more pleasant browsing experience, especially with URL signing (DRM) since all requests take longer.

src/hooks/useSignedUrl.ts
ChristiaanScheermeijer: I've added the `enabled` param because sometimes we want to defer the request when being dependent on a different useQuery. 

The `usePlaylist` and `useMedia` hooks also have the `enabled` param which syncs the enabled state between them. Although not concrete...

src/hooks/useSignedUrl.ts
ChristiaanScheermeijer: Yes, sounds good. I have no preference here

src/hooks/usePlaylist.ts
```diff
@@ -1,24 +1,42 @@
...
+export default function usePlaylist(playlistId: string, params: GetPlaylistParams = {}, enabled: boolean = true, usePlaceholderData: boolean = true) {
+  const { token, signingEnabled, drmEnabled, drmPolicyId, isLoading } = useSignedUrl('playlist', playlistId, params, enabled);
+
+  const queryResult = useQuery(
+    ['playlist', playlistId, params, token],
+    async () => {
+      const playlist = await getPlaylistById(playlistId, { ...params, token }, drmEnabled ? drmPolicyId : undefined);
+
+      return filterMediaItem(playlist, params.related_media_id);
+    },
```
ChristiaanScheermeijer: This will be cleaned up in the TVOD merge

src/stores/AccountController.ts
```diff
@@ -389,6 +391,34 @@ export async function reloadActiveSubscription({ delay }: { delay: number } = {
...
+ *
+ * @param mediaIds
+ */
+export async function getMediaItems(mediaIds: string[]) {
+  const jwt = useAccountStore.getState().auth?.jwt;
+  const signingConfig = useConfigStore.getState().config.contentSigningService;
+  const host = signingConfig?.host;
+  const drmPolicyId = signingConfig?.drmPolicyId;
+  const drmEnabled = !!drmPolicyId && !!signingConfig?.drmEnabled;
+  const signingEnabled = !!host && drmEnabled && !!drmPolicyId;
```
budzworthy: These property checks are done in 3 different places and they are all a little different (`AccountController`, `useSignedUrl`, and `usePlaylistItemCallback`). Can we just make host required on `contentSigningService` (but the entire block still optional) and drop the `drmEnabled`, so we just assume drm is enabled if `drmPolicyId` is populated?

src/stores/AccountController.ts
```diff
@@ -389,6 +391,34 @@ export async function reloadActiveSubscription({ delay }: { delay: number } = {
...
+
+  if (signingEnabled) {
```
budzworthy: Is it not possible to have signing enabled without DRM?

src/providers/QueryProvider.tsx
budzworthy: I thought the signed urls / drm urls only worked for a short time

src/hooks/useMedia.ts
ChristiaanScheermeijer: Updated this, but realized that this probably wasn't needed because it will be removed in the near future anyway.

src/providers/QueryProvider.tsx
ChristiaanScheermeijer: The video DRM tokens are, but that doesn't count for the media/playlist metadata requests as long it's cached. We still do need to invalidate the token after 15 minutes so that we don't risk using invalidated tokens.

src/stores/AccountController.ts
ChristiaanScheermeijer: Yes, sounds good. I've consciously kept the `drmEnabled` as config so this can be controlled by the property settings later. But for now, it will make it more simple.

There are differences between the metadata and video calls. E.g. all media/playlist calls only need signing when DRM is enabled in the property whilst the media item call for video playback needs to be signed for URL signing and DRM.

src/stores/AccountController.ts
ChristiaanScheermeijer: See the comment above, this call is only used for UI, not for video playback.

src/providers/QueryProvider.tsx
budzworthy: OK, so the strategy is to use an hour for the default for things that can be cached indefinitely, and then control invalidation for other things manually?

budzworthy (approved):

ChristiaanScheermeijer:
Thanks @dbudzins, I've pushed one more commit which removes a redundant playlist filter (which is moved to the API service).

budzworthy (approved):

PR #77, opened 2022-06-01
Fix / entitlement bug fixes

ChristiaanScheermeijer:
## Description

Here are a few bug fixes which had a high priority and seemed safe enough to fix. Most of the bugs reported by ourselves were due to the issues yesterday in the sandbox environment. 

**Fixes:**

- No offer modal after "buying" a TVOD item in the AUTHVOD access model
- Improve the entitlement mechanism by:
  - leveraging the `invalidateQueries` method from react-query
  - enable re-fetch on mount for the entitlement queries
  - clean entitlement caches after logout/reload subscriptions calls
  - wait for the reload subscriptions delay to be finished before continuing (loading state before confirming purchase)
- Fix the start watching button using the wrong URL for series
- Enable the Payments button in all menus and hide the subscription section in the AUTHVOD access model
- Clear any existing order after closing the checkout modal causing customers to see outdated order data

Next to this I've also fixed a metadata error in the Primitive Animals series causing it to navigate to the wrong media (different property).

### DIFF
--- a/src/hooks/useOffers.ts
+++ b/src/hooks/useOffers.ts
@@ -20,7 +20,6 @@ const useOffers = () => {
   const { cleengSandbox, json } = config;
   const { requestedMediaOffers } = useCheckoutStore(({ requestedMediaOffers }) => ({ requestedMediaOffers }), shallow);
   const hasPremierOffer = (requestedMediaOffers || []).some((offer) => offer.premier);
-  const tvodOfferIds = (requestedMediaOffers || []).map(({ offerId }) => offerId);
   const [offerType, setOfferType] = useState<OfferType>(accessModel === 'SVOD' ? 'svod' : 'tvod');
 
   const monthlyOfferId = json?.cleengMonthlyOffer ? (json.cleengMonthlyOffer as string) : '';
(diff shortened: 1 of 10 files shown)


budzworthy (approved):
Looks good. I really appreciate all the comments in the code!

src/hooks/useOffers.ts
```diff
@@ -50,7 +49,9 @@ const useOffers = () => {
     const allOffers = offerQueries.reduce<Offer[]>((prev, cur) => (cur.isSuccess && cur.data?.responseData ? [...prev, cur.data.responseData] : prev), []);
     const offers = allOffers.filter((offer) => (offerType === 'tvod' ? !isSVODOffer(offer) : isSVODOffer(offer)));
     const offersDict = Object.fromEntries(offers.map((offer) => [offer.offerId, offer]));
-    const defaultOfferId = hasPremierOffer ? tvodOfferIds[0] : yearlyOfferId || monthlyOfferId;
+    // we need to get the offerIds from the offer responses since it contains different offerIds based on the customers
+    // location. E.g. if an offer is configured as `S12345678` it becomes `S12345678_US` in the US.
+    const defaultOfferId = offers[offers.length - 1]?.offerId;
```
royschut, line 54: Doesn't this undo that a premier TVOD offer gets selected by default?

royschut: Never mind, you've just filtered out the other offers based on the `offerType` :)

royschut (approved):

PR #81, opened 2022-06-08
feat(project): support watchlists

### LINKED ISSUE #31: Handle Cleeng externalData maxsize
The [externalData attribute of Cleeng](https://developers.cleeng.com/reference/update-customers-data) can contain max 5000 characters. This is 50-75 objects.
So we maximize the number of history and favorite objects to 25 and rotate the oldest one out based on the 'updated' attribute

An optimized format is under discussion to ensure more items can be stored.  See [docs/features/user-watchlist.md](https://github.com/jwplayer/ott-web-app/blob/develop/docs/features/user-watchlists.md). This is possible now we have watchlist.

### LINKED ISSUE #43: Add support for watchlists in Web OTT app
Watchlist The new OTT Web App includes features for "Continue Watching" and "Your Favorites". These are manually compiled and tracked in the frontend.

Our new Watch Lists are developed to improve the user and editorial experience of this feature:

The app will use the watch list to "bulk fetch" metadata. Currently metadata is fetched one video at a time, reducing Lighthouse scores.

The app will return the watch list ID as part of analytics, so editors can track the usage and impact of watch lists on plays / watchtime.

This epic covers the OTT Web App front end, not the back end (covered as part of the OTT Config Updates).

AntonLantukh:
## Description

- New endpoint /apps/watchlists is now used to get favorites and continue_watching playlists (no drm is needed)
- Max number of favorites and continue_watching videos added (30 for each)
- Alert component to be shown in case of exceeding for favorites- The size of externalData string stored at the back-end side reduced- Local storage watchlists info save irrespective of the user account presence
- New validation to check that both content array item present and the corresponding watchlist are set
- Favorites row can be hidden now
- e2e tests add for favorites and continue_watching
- Fixed zustand devtools warnings when launching tests
- Changed tests on commit run just for touched files
- Changed Alert, Modal and Dialog snapshots

This PR solves #43.

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/App.tsx
```diff
@@ -45,7 +47,7 @@ class App extends Component {
 
   configErrorHandler = (error: Error) => {
     this.setState({ error });
-    console.info('Error while loading the config.json:', error);
+    console.error('Error while loading the config.json:', error);
```
budzworthy: Suggest using logDev util here and considering something else in production

docs/configuration.md
```diff
@@ -278,6 +280,18 @@ The eight-character Playlist ID of the Search playlist that you want to use to e
 
 ---
 
+**features.favorites_list** (optional)
```
budzworthy: I recommend we keep these the same format as search and recommendations (i.e. `favoritesPlaylist`.) If this isn't what it is on the backend, let's fix it there before any of this goes out.

docs/configuration.md
```diff
@@ -278,6 +280,18 @@ The eight-character Playlist ID of the Search playlist that you want to use to e
...
+
+The eight-character Playlist ID of the Watchlist playlist that you want to use to populate the "Favorites" shelf in your site.
+
+---
+
+**features.continue_watching_list** (optional)
```
budzworthy: Use `continueWatchingPlaylist`

public/config.json
```diff
@@ -9,14 +9,6 @@
       "type": "playlist",
       "contentId": "JSKF03bk"
     },
-    {
-      "enableText": true,
```
budzworthy: Can we keep these in the default example config (we can add playlists to the Blender property)

src/App.tsx
```diff
@@ -28,11 +27,14 @@ class App extends Component {
   }
 
   async initializeServices(config: Config) {
-    if (config.content.some((el) => el.type === PersonalShelf.ContinueWatching)) {
+    // We only request favorites and continue_watching data if these features are enabled
+    if (config?.features?.continue_watching_list) {
```
budzworthy: Should we include both checks here?  First if the playlist is set in features, but then also if it's used?

src/components/Alert/__snapshots__/Alert.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Vitest Snapshot v1
+
+exports[`<Alert> > renders and matches snapshot 1`] = `<div />`;
```
budzworthy: Am I missing something here? Why is this just an empty div?

src/containers/Cinema/Cinema.tsx
```diff
@@ -31,25 +28,27 @@ type Props = {
 };
 
 const Cinema: React.FC<Props> = ({ item, onPlay, onPause, onComplete, onUserActive, onUserInActive, feedId, isTrailer = false }: Props) => {
-  const config: Config = useContext(ConfigContext);
```
budzworthy, line 34: These should be the same. Any particular reason for the change?

docs/configuration.md
AntonLantukh: Good catch Danny! I am still confused about differences in case notation. I will fix it on Monday after coming back from PTO.

src/containers/Cinema/Cinema.tsx
```diff
@@ -31,25 +28,27 @@ type Props = {
...
   const setPlayer = useOttAnalytics(item, feedId);
   const handlePlaylistItemCallback = usePlaylistItemCallback();
 
-  const getProgress = useCallback((): VideoProgress | null => {
+  const getProgress = useCallback((): number | null => {
     if (!playerRef.current) return null;
 
-    const duration = playerRef.current.getDuration();
-    const progress = playerRef.current.getPosition() / duration;
+    const progress = playerRef.current.getPosition() / item.duration;
```
budzworthy, line 49: Is there any chance that item.duration will be wrong, less precise, or missing?

src/containers/Layout/Layout.module.scss
```diff
@@ -4,7 +4,7 @@
 .layout {
   display: flex;
   flex-direction: column;
-  min-height: calc(100vh - calc(100vh - 100%));
```
budzworthy, line 7: Can you check with @ChristiaanScheermeijer and @royschut do see if they remember why this calculation is the way it is and why it's min height?  It seems that they did a lot of iteration to get this right last summer, and I worry there is some subtle thing we're missing.

src/screens/Movie/Movie.tsx
```diff
@@ -55,13 +60,40 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
+  const onFavoritesWarningToggle = useCallback(() => {
+    setIsFavoritesWarningShown(!isFavoritesWarningShown);
+  }, [setIsFavoritesWarningShown, isFavoritesWarningShown]);
+
+  const onFavoriteButtonClick = useCallback(() => {
+    if (!item) {
+      return;
+    }
+
+    if (isFavorited) {
```
budzworthy: Did you try clicking the favorite button really fast to make sure this isn't a race condition? It seems like it should be ok, but I can't quite mentally map the flow with the callback, the store access, and the local variable.

src/screens/Movie/Movie.tsx
```diff
@@ -55,13 +60,40 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
   // User, entitlement
   const { user, subscription } = useAccountStore(({ user, subscription }) => ({ user, subscription }), shallow);
   const { isEntitled } = useEntitlement(item);
 
   // Handlers
+  const onFavoritesWarningToggle = useCallback(() => {
+    setIsFavoritesWarningShown(!isFavoritesWarningShown);
+  }, [setIsFavoritesWarningShown, isFavoritesWarningShown]);
+
+  const onFavoriteButtonClick = useCallback(() => {
```
budzworthy, line 69: Would it make sense to put all of this logic into the favoritesController and just expose a `toggleFavorite` method?

src/services/config.service.ts
```diff
@@ -28,6 +47,16 @@ const featuresSchema: SchemaOf<Features> = object({
   enableSharing: boolean().notRequired(),
   recommendationsPlaylist: string().nullable(),
   searchPlaylist: string().nullable(),
+  continue_watching_list: string().test('has-continue_watching-list-element', 'errorMessage', (value, context) => {
```
budzworthy: imho, this is more validation than we want a schema check to do. Instead of just checking formatting, it's starting to enforce business logic.

src/services/config.service.ts
```diff
@@ -7,6 +7,25 @@ import { PersonalShelf } from '#src/enum/PersonalShelf';
  * Set config setup changes in both config.services.ts and config.d.ts
  * */
 
+/**
+ * We check here that we:
+ * 1. Added favorites_list / continue_watching_list feature
+ * 2. Included a corresponding element (with favorites or continue_watching type) in the content array
+ */
+const checkAdditionalFeatures = (content: Content[], playlistId: string | undefined | null, type: PersonalShelf) => {
```
budzworthy: This check seems a bit heavy handed and I think it will cause us more trouble than good. We've talked before about potentially having favorites or continue watching show up as libraries instead of in the content. It's possible that other use cases will arise where we want to set the playlist ID's but not have a content item.

src/services/config.service.ts
```diff
@@ -85,30 +114,30 @@ const loadConfig = async (configLocation: string) => {
...
-  if (!data.content.some(({ type }) => type === PersonalShelf.Favorites)) {
-    data.content.push({ type: PersonalShelf.Favorites });
-  }
+const enrichConfig = (config: Config): Config => {
+  const { content, siteName } = config;
+  const updatedContent = content.map((content) => Object.assign({ enableText: true, featured: false }, content));
 
-  data.content = data.content.map((content) => Object.assign({ enableText: true, featured: false }, content));
-  data.siteName = data.siteName || 'My OTT Application';
+  return { ...config, siteName: siteName || 'My OTT Application', content: updatedContent };
```
budzworthy: Might be nice to put this into the translations and also expose the build version

src/stores/FavoritesController.ts
```diff
@@ -2,36 +2,43 @@ import { useAccountStore } from '#src/stores/AccountStore';
...
+export const serializeFavorites = (favorites: Favorite[]): SerializedFavorite[] => {
+  return favorites.map(({ mediaid }) => ({ mediaid }));
 };
 
 export const persistFavorites = () => {
   const { favorites } = useFavoritesStore.getState();
   const { user } = useAccountStore.getState();
 
   if (user) {
-    return updatePersonalShelves();
```
budzworthy, line 31: I think these are intentionally split. If your favorites save locally when you are logged in, it could leak personal data (i.e. if you favorite a bunch of embarrassingly bad movies, your partner would see them and make fun of you if they shared your browser.)

src/stores/WatchHistoryController.ts
```diff
@@ -51,38 +48,52 @@ export const persistWatchHistory = () => {
...
   const { watchHistory } = useWatchHistoryStore.getState();
 
   if (!videoProgress) return;
 
   const watchHistoryItem = createWatchHistoryItem(item, videoProgress);
   const index = watchHistory.findIndex(({ mediaid }) => mediaid === watchHistoryItem.mediaid);
 
+  let updatedHistory = watchHistory;
+
   if (index > -1) {
```
budzworthy: Can these 2 paths be combined now and we can skip checking index beforehand?

src/stores/WatchHistoryController.ts
```diff
@@ -51,38 +48,52 @@ export const persistWatchHistory = () => {
...
-    watchHistory[index] = watchHistoryItem;
-    useWatchHistoryStore.setState({ watchHistory });
+    updatedHistory = [watchHistoryItem, ...watchHistory.filter(({ mediaid }) => mediaid !== watchHistoryItem.mediaid)];
   } else {
-    useWatchHistoryStore.setState({ watchHistory: [watchHistoryItem].concat(watchHistory) });
+    updatedHistory = [watchHistoryItem, ...watchHistory];
+    if (watchHistory.length > MAX_WATCHLIST_ITEMS_COUNT) {
+      updatedHistory = updatedHistory.slice(0, watchHistory.length - 1);
```
budzworthy: Should this be `updatedHistory = updatedHistory.slice(0, MAX_WATCHLIST_ITEMS_COUNT - 1);`
Otherwise if the watchlist gets more items in it somehow it won't be restricted down to the intended length.

src/stores/utils.ts
```diff
@@ -2,12 +2,11 @@ import type { State, StateCreator } from 'zustand';
...
 import { devtools, subscribeWithSelector } from 'zustand/middleware';
 
-import { IS_DEV_BUILD } from '#src/utils/common';
-
 export const createStore = <T extends State>(name: string, storeFn: StateCreator<T>) => {
   const store = subscribeWithSelector(storeFn);
 
-  if (IS_DEV_BUILD) {
+  // https://github.com/pmndrs/zustand/issues/852#issuecomment-1059783350
+  if (import.meta.env.MODE === 'development') {
```
budzworthy, line 9: How is this different than using `.DEV`?

src/components/Alert/__snapshots__/Alert.test.tsx.snap
AntonLantukh: I mocked the outer container to get the correct snapshot.

src/App.tsx
AntonLantukh: I added this check to the validation function. It requires both content item and a playlist in features.

```
  continueWatchingList: string().test('has-continue_watching-list-element', 'errorMessage', (value, context) => {
    // @ts-expect-error https://github.com/jquense/yup/issues/1631
    const { content, features } = context.from[1].value as Config;
    return checkAdditionalFeatures(content, value, PersonalShelf.ContinueWatching);
  }),
```

So it should not reach this stage if there is only content item or a playlist in the features section.

src/containers/Cinema/Cinema.tsx
AntonLantukh: There are only 3 use cases left with the config in the context. I thought we are now mostly using config as a store value.

src/containers/Cinema/Cinema.tsx
AntonLantukh: I didn't notice any issues if we don't store the duration in externalData but just retrieve it from the media object.

src/containers/Layout/Layout.module.scss
AntonLantukh: I did this one for the footer to be on the bottom of page. Right now it just sticks to the content.

Now:

Screenshot 2022-06-14 at 11 11 22

After changes:

Screenshot 2022-06-14 at 11 10 12

src/screens/Movie/Movie.tsx
AntonLantukh: I tired it know. Works fine.

src/screens/Movie/Movie.tsx
AntonLantukh: We should toggle the Alert component state in the callback to show something better than just a native alert message:

Screenshot 2022-06-14 at 11 19 20

That is why I decided to use it as a component's callback.

src/stores/utils.ts
AntonLantukh: We get a plenty of warning messages while launching the tests:

Screenshot 2022-06-14 at 11 30 29

IS_DEV_BUILD is true for both 'test' and 'dev' modes, which means that we initialize dev-tools for tests.

src/App.tsx
AntonLantukh: `logDev` uses info method to show the message. Do we need an alternative function to show the error messages? We can also have one and just pass a message type as a param.

src/services/config.service.ts
AntonLantukh: The problem is, that it won't and it also shouldn't work just with a `content` array item or with a playlist in the `features` section. We need both a row place and a playlist to retrieve media items. We can possibly only check the presence of the feature, and if it exists - insert the content item automatically. However this solution may be not so obvious for the customer.

I think that when possible we should let the customer know what went wrong. And the earlier we do it, the easier it is to track the issue.

We could possibly avoid this problem having `favorites` and `continue_watching` just in the content section (as other rows), but as far as I understood it is not possible to do at the back-end side.

src/services/config.service.ts
AntonLantukh: I have answered in the previous thread 😄 Do you think that we should add content items with `favorites` and `continue_watching` on our own if the customer only set the playlist id?

src/services/config.service.ts
AntonLantukh: I added the default name to translations. What do you mean by "build version"? Is it the one we use for tagging our versions on GitHub now?

src/stores/FavoritesController.ts
AntonLantukh: Makes sense. I changed it back.

src/containers/Layout/Layout.module.scss
royschut: I remember that we added this calculation to compensate for native browser bars which subtract from vh (particularly on Android if I remember well). However, I think the main purpose was to keep the footer at the bottom indeed, so if your first screenshot was the situation before PR, then there must've been an error in it.

src/App.tsx
budzworthy: I've always shied away from showing debug messages in production beyond the generic "something went wrong". Giving all users too much detail is only going to confuse most people and potentially expose too much "behind the scenes"

src/containers/Cinema/Cinema.tsx
budzworthy: That makes sense... Maybe it's worth a separate PR just to remove the duplication entirely then?

src/containers/Cinema/Cinema.tsx
budzworthy: OK, then nice optimization!

src/containers/Layout/Layout.module.scss
budzworthy: Let's try to do some cross browser testing at some point.

src/screens/Movie/Movie.tsx
budzworthy: What I mean is to decouple the favorites logic from the Movie component. So the favorite logic would all be in the controller and an error property on the store would trigger the warning.

src/App.tsx
budzworthy: It seems cleanest to me to only enforce that there is a playlist id set if there's a content item that needs it, but not the other way around. But that being said, this would mean we initialize the service if the playlist id is set, so this should be fine either way.

src/services/config.service.ts
budzworthy: But what does it hurt if they set a playlist id but then don't use it in the content items? It may be a mistake but it also may be a use case we haven't planned for yet (like they want to put favorites on a new page or in a library for example.) I understand we need to enforce that the playlist id is set if the content item is used, but don't think the opposite is necessarily a hard fail. Secondly, even if we decide it is, I'd prefer to do this validation in the controller or service and let the scheme validation just be formatting checks.

src/services/config.service.ts
budzworthy: See above, but I think it's fine to just allow it

src/services/config.service.ts
budzworthy: Yes, but that can be a separate feature request

src/stores/utils.ts
budzworthy: Ok makes sense

src/App.tsx
AntonLantukh: I agree, I just wanted to suggest to differentiate info and error messages for dev environment, for example:

```
  const logDev = (message, type = 'info') => {
      console[type](message);
  }
```

src/containers/Cinema/Cinema.tsx
AntonLantukh: I created an issue for this one and will fix it in a separate PR

src/App.tsx
AntonLantukh: Changed it back to content item presence

src/services/config.service.ts
AntonLantukh: I agree that nothing prevents us from using just playlistIds in features with no corresponding content items. So I moved it to `loadConfig` function and added a log message instead of error. Please have a look.

src/screens/Movie/Movie.tsx
AntonLantukh: I did my best to decouple it and to store it in the controller.

src/stores/FavoritesController.ts
```diff
@@ -63,6 +64,30 @@ export const removeItem = (item: PlaylistItem) => {
...
+
+  if (isFavorited) {
+    removeItem(item);
+
+    return;
+  }
+
+  // If we exceed the max available number of favorites, we show a warning
+  if (favorites?.length >= MAX_WATCHLIST_ITEMS_COUNT) {
+    toggleWarning();
```
budzworthy: I would suggest using an optional string property for the warning, and then you can explicitly set the warning message here. It allows for better scalability when/if more error messages get incorporated.  It also means that the translation substitution of `MAX_WATCHLIST_ITEMS_COUNT` can happen directly here and is more closely coupled to the place where we check that limit.

src/stores/FavoritesStore.ts
```diff
@@ -6,12 +6,16 @@ import type { Playlist, PlaylistItem } from '#types/playlist';
...
   hasItem: (item: PlaylistItem) => boolean;
   getPlaylist: () => Playlist;
+  toggleWarning: () => void;
 };
 
-export const useFavoritesStore = createStore<FavoritesState>('FavoritesState', (_, get) => ({
+export const useFavoritesStore = createStore<FavoritesState>('FavoritesState', (set, get) => ({
   favorites: [],
+  isWarningShown: false,
+  toggleWarning: () => set({ isWarningShown: !get().isWarningShown }),
```
budzworthy: Toggling here is a little confusing. I would expect the places it's called to either want to set it or clear it, and that there isn't really a case for toggling. (see also the suggestion about using a string for the warning instead of a boolean)

src/screens/Movie/Movie.tsx
```diff
@@ -204,10 +188,10 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
           ) : undefined}
 
           <Alert
-            open={isFavoritesWarningShown}
+            open={isWarningShown}
             title={t('video:favorites_warning.title')}
             body={t('video:favorites_warning.body', { count: MAX_WATCHLIST_ITEMS_COUNT })}
-            onClose={onFavoritesWarningToggle}
+            onClose={onToggleWarning}
```
budzworthy: Toggle is not really the right operation here.  Closing the alert should clear the warning regardless of the current state.  It probably would be hard for this to get into a buggy state with the way it's programmed now, but it is a possibility, especially as code evolves, and either way, having it clear the warning makes the code more readable.  Note, there are some other comments about the warningShown property below.

src/screens/Movie/Movie.tsx
```diff
@@ -60,39 +59,24 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
-    if (favoritesCount >= MAX_WATCHLIST_ITEMS_COUNT) {
-      onFavoritesWarningToggle();
-      return;
-    }
+    toggleFavorite(item);
+  }, [item]);
 
-    saveItem(item);
-  }, [item, isFavorited, favoritesCount, onFavoritesWarningToggle]);
+  const onToggleWarning = useCallback(() => {
```
budzworthy: Do you need useCallback here? It keeps the code a little cleaner to remove it.  It looks to me like the function will be the same reference every call, and even if it wasn't, the reference changing would cause the memoized callback to change anyhow.

src/screens/Movie/Movie.tsx
```diff
@@ -60,39 +59,24 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
...
   // User, entitlement
   const { user, subscription } = useAccountStore(({ user, subscription }) => ({ user, subscription }), shallow);
   const { isEntitled } = useEntitlement(item);
 
   // Handlers
-  const onFavoritesWarningToggle = useCallback(() => {
-    setIsFavoritesWarningShown(!isFavoritesWarningShown);
-  }, [setIsFavoritesWarningShown, isFavoritesWarningShown]);
-
   const onFavoriteButtonClick = useCallback(() => {
```
budzworthy, line 69: This is sort of outside of the scope of this PR, but all of the favorite, share, etc. functions can probably go directly into the VideoComponent component, because I don't think there are any differences between Movie and Series and it's just creating duplicated code and unnecessary property passing.  I would maybe even go so far as to making separate components for the Favorite Button, Share Button, etc. inside of the VideoComponent.

src/screens/Series/Series.tsx
```diff
@@ -60,14 +63,27 @@ const Series = ({ match, location }: RouteComponentProps<SeriesRouteParams>): JS
...
   // User, entitlement
   const { user, subscription } = useAccountStore(({ user, subscription }) => ({ user, subscription }), shallow);
   const { isEntitled } = useEntitlement(item);
 
   // Handlers
+  const onFavoriteButtonClick = useCallback(() => {
+    toggleFavorite(item);
+  }, [item]);
+
+  const onToggleWarning = useCallback(() => {
```
budzworthy: Same comment about if we can remove the useCallback and just use the function directly.

src/i18n/locales/en_US/video.json
```diff
@@ -21,5 +21,9 @@
   "trailer": "Trailer",
   "video_not_found": "Video not found",
   "watch_trailer": "Watch the trailer",
-  "share_video": "Share this video"
+  "share_video": "Share this video",
+  "favorites_warning": {
+    "title": "Maximum amount of favorite videos exceeded",
```
budzworthy: In line with the other comments about changing the warning to a string, I would suggest keeping this title generic. Maybe even just something as generic as "An Error Occurred" and then let the body communicate the full message.  It's kind of hard to make a succinct title that communicates the max exceeded anyhow.

src/screens/Series/Series.tsx
```diff
@@ -190,6 +207,12 @@ const Series = ({ match, location }: RouteComponentProps<SeriesRouteParams>): JS
             isLoggedIn={!!user}
             hasSubscription={!!subscription}
           />
+          <Alert
```
budzworthy: This can probably go inside of the VideoComponent too

src/stores/WatchHistoryController.ts
```diff
@@ -79,17 +79,13 @@ export const saveItem = (item: PlaylistItem, videoProgress: number | null) => {
...
-  } else {
-    updatedHistory = [watchHistoryItem, ...watchHistory];
-    if (watchHistory.length > MAX_WATCHLIST_ITEMS_COUNT) {
-      updatedHistory = updatedHistory.slice(0, watchHistory.length - 1);
-    }
+  updatedHistory = [watchHistoryItem, ...watchHistory.filter(({ mediaid }) => mediaid !== watchHistoryItem.mediaid)];
+
+  if (watchHistory.length > MAX_WATCHLIST_ITEMS_COUNT) {
+    updatedHistory = updatedHistory.slice(0, MAX_WATCHLIST_ITEMS_COUNT - 1);
   }
```
budzworthy:
```suggestion
  const updatedHistory = watchHistory.filter(({ mediaid }) => mediaid !== watchHistoryItem.mediaid);
  updatedHistory.unshift(watchHistoryItem);
  updatedHistory.splice(MAX_WATCHLIST_ITEMS_COUNT);
```

src/screens/Movie/Movie.tsx
AntonLantukh: I will have a look in Series PR.

src/screens/Movie/Movie.tsx
AntonLantukh: Well, we mostly try to use it for the cases we pass callback to other components for them not to rerender unnecessarily. But in this case we don't show the alert fro most cases at all, so I removed it.

src/stores/WatchHistoryController.ts
AntonLantukh: Hm, strange, because I already updated this part of logic before.

AntonLantukh:
Thank you Danny! Really valuable comments to improve the code.

src/utils/formatting.ts
```diff
@@ -40,7 +40,9 @@ export const addQueryParams = (url: string, queryParams: { [key: string]: string
     const value = queryParams[key];
 
     // null or undefined
-    if (value == null || !value?.length) return;
+    if (value == null) return;
```
budzworthy, line 42: Was the double equals on purpose instead of triple?  Any reason not to just check truthy?

src/components/Video/Video.tsx
```diff
@@ -81,6 +83,12 @@ const Video: React.FC<Props> = ({
   const handlePause = useCallback(() => setIsPlaying(false), []);
   const handleComplete = useCallback(() => onComplete && onComplete(), [onComplete]);
 
+  const { clearWarning, warning } = useFavoritesStore((state) => ({
+    isFavorited: !!item && state.hasItem(item),
```
budzworthy: Unused

budzworthy (approved):
A couple small comments if you have a chance to address them, otherwise looks good

src/utils/formatting.ts
AntonLantukh: It should check both `undefined` and `null`

budzworthy:
```suggestion
    if (!value) return;
```

AntonLantukh: Then we will skip empty strings :) I will check, maybe we needed them for some cases.

AntonLantukh: Yeah, for search playlists we get error for empty string:

`Search feeds require a search phrase`

PR #87, opened 2022-06-16
refactor(e2e): e2e-report

### LINKED ISSUE #79: Fix e2e test
After the last sprint, there are some failing e2e tests. Fix these and also make sure that tests are added for the tvod scenarios.

AntonLantukh:
## Description

- Fixed types in e2e tests (now strict types are used in each test file)
- Fixed tests
- Workers added for CodeceptJS to run tests faster (~6 min for each device now) 
- Node version updated
- Added allure report
- Added artifacts for allure reports
- File structure changed to support several features for one suite
- Readme added

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


test-e2e/tests/account/index_test.ts
```diff
@@ -1,35 +1,27 @@
-import passwordUtils, {LoginContext} from "../utils/password_utils";
```
budzworthy: I don't know if it's going to be necessary to create directories for each test category. I'd probably prefer having a mixed hierarchy with the single files keeping their names and being in /tests and only having subfolders for categories that have multiple files.

budzworthy (commented):
Really good improvements and documentation. My only feedback is maybe consider not adding all the single file sub-folders in /test-e2e/test.

Also would it be good to just run lint fix on all of the files so we stop getting so many format change noise?

test-e2e/tests/account/index_test.ts
AntonLantukh: I agree, I fix this one

AntonLantukh:
> Really good improvements and documentation. My only feedback is maybe consider not adding all the single file sub-folders in /test-e2e/test.
> 
> Also would it be good to just run lint fix on all of the files so we stop getting so many format change noise?

No problem. I think I should better merge linting changes first and then just update this branch.

budzworthy (approved):

PR #98, opened 2022-06-28
feat(series): add new series-api

### LINKED ISSUE #59: Support New Series Management
- Series have a new data structure
- Support the new delivery endpoint
- Detect media placeholder that represents a series and then activate the series page in that case
- Keep support for the current implementation that uses playlists
- Support bonus content

Out-of-scope
- Functional web app enhancements like support for continue watching on series level

image

[Series Overview for Partners](https://docs.google.com/presentation/d/1G-gymaZCSf6Wr_eSpCRIrQgLaTPuTZP5gyJgT_WBeew/edit#slide=id.g11a81b300da_0_23)

image

AntonLantukh:
## Description

- we still support old approach with series as media items
- new Switcher component added to check whether new api can be used
- SeriesNew component now uses new api to show series
- dev host content-portal added to test demo features
- fixed player to show the next episode in case a current one is finished
- docs updates
- **(!)** we use features.favoritesList to get media items (temporary)

This PR solves #59.

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/hooks/usePlaylist.ts
```diff
@@ -3,16 +3,27 @@ import { generatePlaylistPlaceholder } from '#src/utils/collection';
...
 const placeholderData = generatePlaylistPlaceholder(30);
 
 export default function usePlaylist(playlistId?: string, params: GetPlaylistParams = {}, enabled: boolean = true, usePlaceholderData: boolean = true) {
   const callback = async (token?: string, drmPolicyId?: string) => {
     const playlist = await getPlaylistById(playlistId, { token, ...params }, drmPolicyId);
 
-    // This pre-caches all playlist items and makes navigating a lot faster. This doesn't work when DRM is enabled
-    // because of the token mechanism.
+    const mediaIds = (playlist?.playlist || []).map((el) => el.mediaid);
+    const series = await getSeriesByMediaIds({ media_ids: mediaIds });
```
ChristiaanScheermeijer: What does this do specifically? I think we don't want to use this for all playlist items, right?

AntonLantukh: This one get `series` information about media items. That way we can get `episodeNumber` and `seasonNumber` to show a corresponding tag.

ChristiaanScheermeijer: If this is only needed for the new Series screen, we better make it conditional. Now, this logic is also used on the homepage and playlist screens.

AntonLantukh: For Home screen we usually show episode and season number of media items in a non-series playlist. According to the previous approach, we had to add them manually as custom params to each episode (media item). Now there is no such need as we already have predefined series.

AntonLantukh: I thought about additional requests, but we load playlists only fitting the viewport size, so this will limit the amount of them.

AntonLantukh: Agreed no to use `getSeriesByMediaIds` endpoint at all.

src/components/Root/Root.tsx
```diff
@@ -34,7 +34,7 @@ const Root: FC<Props> = ({ error }: Props) => {
         <Route path="/" component={Home} exact />
         <Route path="/p/:id" component={Playlist} exact />
         <Route path="/m/:id/:slug?" component={Movie} exact />
-        <Route path="/s/:id/:slug?" component={Series} />
+        <Route path="/s/:id/:slug?" component={SeriesSwitcher} />
```
budzworthy: What about `SeriesRouter` as a name?

package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "prepare": "husky install",
     "start": "vite",
+    "start:dev-api": "APP_API_ENV='dev' vite",
```
budzworthy:
```suggestion
    "start:dev-api": "APP_API_ENV='dev' yarn start",
```

src/containers/Cinema/Cinema.tsx
```diff
@@ -130,9 +138,10 @@ const Cinema: React.FC<Props> = ({ item, onPlay, onPause, onComplete, onUserActi
       if (currentItem && currentItem.mediaid === item.mediaid) {
         return;
       }
-
       // load new item
-      playerRef.current.setConfig({ playlist: [deepCopy(item)], autostart: true });
+      playerRef.current.load([deepCopy(item)]);
```
budzworthy, line 141: What are the reasons for the changes here?  Is this a bug fix, code cleanup, or logic change?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+        const ids = season.episodes.map((el) => el.media_id);
+        return [...acc, ...ids];
+      }, []);
+
+      const mediaItems = await getMediaByWatchlist(watchlistId, mediaIds);
+
+      return { series, mediaItems: enrichMediaItems(series, mediaItems) };
+    },
+    {
+      staleTime: Infinity,
```
budzworthy: Series data probably won't change often, but it may change. Maybe a long-ish stale time less than infinity would be good?  Like an hour or 8 hours?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+      retry: 0,
```
budzworthy: Any reason not to allow 2-3 retries?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+    async () => {
+      if (!watchlistId) {
+        throw Error('Please set features.favoritesList property');
+      }
+
+      const series = await getSeries(seriesId);
+      const mediaIds = (series?.seasons || []).reduce((acc: string[], season: Season) => {
+        const ids = season.episodes.map((el) => el.media_id);
+        return [...acc, ...ids];
+      }, []);
```
budzworthy:
```suggestion
      const mediaIds = (series?.seasons || []).flatMap((season: Season) => season.episodes.map((episode) => episode.media_id));
```

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+import { getSeries } from '#src/services/series.service';
+import { getMediaByWatchlist } from '#src/services/api.service';
+import { enrichMediaItems } from '#src/utils/series';
+import type { Season, Series } from '#types/series';
+import type { PlaylistItem } from '#types/playlist';
+import type { ApiError } from '#src/utils/api';
+
+export const useSeriesData = (id: string, season?: number): UseQueryResult<Series | undefined, ApiError> =>
+  useQuery(`series-${id}`, async () => getSeries(id, { season }), {
+    staleTime: Infinity,
```
budzworthy: Series data probably won't change often, but it may change. Maybe a long-ish stale time less than infinity would be good? Like 1 or 8 hours?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+    retry: 0,
```
budzworthy: Any reason not to allow 2-3 retries?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+export const useSeriesMediaItems = (
```
budzworthy: Is it cleaner to keep 1 custom hook definition per file?

src/services/series.service.ts
```diff
@@ -0,0 +1,23 @@
...
+import { API_BASE_URL } from '../config';
+
+import type { Series, GetSeriesParams } from '#types/series';
+
+/**
+ * Get series by id
+ * @param {string} id
+ * @param params
+ */
+export const getSeries = async (id?: string, params: GetSeriesParams = {}): Promise<Series | undefined> => {
```
budzworthy: Are there going to be any / many more series related service calls?  If not, does it make sense to just add this to the existing `api.service.ts`?

src/screens/SeriesNew/SeriesNew.tsx
```diff
@@ -0,0 +1,200 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
budzworthy: There's a lot of duplicated code in this file and the existing Series file. Do you think that the old and new methods will existing for a short time or a long time?  If short, then duplication is fine, but let's make sure the `Series` file stays the primary path (for git history reasons, etc.) and create a copy called `SeriesLegacy` or something.  Also if we can avoid adding framework like the SeriesSwitcher.  Then when we get rid of the old series, there's less cleanup to do and fewer artifacts.

However, if we think the 2 methods of doing series will stay around for a long time, I think it's worth the effort to try to get both to share the same Series component code and I would try to hide all of the other logic (checking for the 404 error response, falling back to the old series method, etc.) in the hook and/or a controller.

src/containers/StartWatchingButton/StartWatchingButton.tsx
```diff
@@ -49,16 +50,16 @@ const StartWatchingButton: React.VFC<Props> = ({ item }) => {
   }, [isEntitled, isLoggedIn, hasMediaOffers, videoProgress, t]);
 
   const handleStartWatchingClick = useCallback(() => {
-    const seriesId = getSeriesIdFromEpisode(item);
+    const parsedSeriesId = seriesId || getSeriesIdFromEpisode(item);
```
budzworthy: It would keep the button component simpler if you moved this out to the Series component to pass either value through the prop

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+  watchlistId: string | null | undefined,
+): UseQueryResult<{ mediaItems: PlaylistItem[]; series: Series }, ApiError> =>
+  useQuery(
+    `series-watchlist-${seriesId}`,
+    async () => {
+      if (!watchlistId) {
+        throw Error('Please set features.favoritesList property');
+      }
+
+      const series = await getSeries(seriesId);
```
budzworthy, line 13: Should this use the `useSeriesData` hook so it doesn't do another api call?  Would it make sense to combine these 2 hooks?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+      const mediaIds = (series?.seasons || []).reduce((acc: string[], season: Season) => {
+        const ids = season.episodes.map((el) => el.media_id);
+        return [...acc, ...ids];
+      }, []);
+
+      const mediaItems = await getMediaByWatchlist(watchlistId, mediaIds);
+
+      return { series, mediaItems: enrichMediaItems(series, mediaItems) };
```
budzworthy: Will `enrichMediaItems` be used anywhere else or should it be a non-exported function in this file?

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+): UseQueryResult<{ mediaItems: PlaylistItem[]; series: Series }, ApiError> =>
+  useQuery(
+    `series-watchlist-${seriesId}`,
+    async () => {
+      if (!watchlistId) {
+        throw Error('Please set features.favoritesList property');
+      }
+
+      const series = await getSeries(seriesId);
+      const mediaIds = (series?.seasons || []).reduce((acc: string[], season: Season) => {
```
budzworthy: This would be a good place to hide all of the old and new series query methods (i.e. if the series call above fails, fallback to the old series retrieval.)

src/screens/SeriesNew/SeriesNew.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const play = searchParams.get('play') === '1';
+  const feedId = searchParams.get('l');
+
+  // Config
+  const { config, accessModel } = useConfigStore(({ config, accessModel }) => ({ config, accessModel }), shallow);
+  const { styling, features, siteName } = config;
+  const posterFading: boolean = styling?.posterFading === true;
+  const enableSharing: boolean = features?.enableSharing === true;
+
+  // We use favoritesList watchlist here as a temporary solution.
```
budzworthy: Maybe I missed this, but what will we use as the long term solution?  Does the new series logic also have to move to using a watchlist or can we leave it as the individual queries for now, since reusing a different watchlist seems error prone?

src/containers/SeriesSwitcher/SeriesSwitcher.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const SeriesSwitcher = (params: RouteComponentProps<Params>): JSX.Element => {
+  const seriesId = params.match.params.id;
+
+  const { isLoading, isFetching, error } = useSeriesData(seriesId);
+
+  if (isLoading || isFetching) return <LoadingOverlay />;
+
+  // In case we have not found series using id, we assume it is a v2.0 seriesId
+  if (error?.code === 404) return <Series {...params} />;
```
budzworthy: What happens when other errors that aren't API Errors get thrown inside the hook?

src/services/series.service.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (!id) {
+    return undefined;
+  }
+
+  const pathname = `/apps/series/${id}`;
+  const url = addQueryParams(`${API_BASE_URL}${pathname}`, params);
+  const response = await fetch(url);
+  const data = await getDataOrThrow(response);
```
budzworthy: Instead of using getDataOrThrow and adding all the ApiError and response type, why not just check `response.status` here and return undefined if it equals 404?  This function already potentially returns undefined and it prevents using error throwing and relying on error types and inspecting return codes in hooks to control business logic.

budzworthy (commented):
I think it would be worth another try to handle the old and new series splits without the switcher and duplicated screens, unless we know for sure that it will be very temporary.

src/hooks/useSeries.ts
AntonLantukh: That will take much time to load the page (7-8 sec). If we set retryDelay to 200ms, that will still take 2-3 sec.

src/hooks/useSeries.ts
AntonLantukh: Why not. I set it to 8 hours.

src/services/series.service.ts
AntonLantukh: We had two endpoints for series. Now I use only one of them. Good suggestion for now!

src/hooks/useSeries.ts
AntonLantukh: I answered in the upper thread.

src/hooks/useSeries.ts
AntonLantukh: In case we set it to very low value, there is no sense in such retry.

src/services/series.service.ts
AntonLantukh: Hm, returning undefined for such cases doesn't look to me as a good solution. Having an error type provides additional typings for hooks and can promote further customisations. In case we want enrich error object / add code, we can do it in the constructor. Right now we can't even get the code from api response in the hooks. We can move the business logic to the hook but I am not sure that service in the right place.

src/hooks/useSeries.ts
AntonLantukh: Nope, just here. I will use it in the hook. I most cases I just prefer storing utils in a separate file

src/containers/StartWatchingButton/StartWatchingButton.tsx
AntonLantukh: You are right, I will move it to the outer component.

src/containers/SeriesSwitcher/SeriesSwitcher.tsx
AntonLantukh: We will use SeriesNew component for such cases.

src/screens/SeriesNew/SeriesNew.tsx
AntonLantukh: As the long term solution we wanted to create and use watchlists at the back-end side and not to expose this functionality to the front-end. I am not sure when such functionality may be ready :( First I wanted to use a separate watchlist for this purpose, but then I was advised to reuse a current one.

We also have `Bulk retrieve series by media ids` endpoint to retrieve media items by their ids. However the object we get doesn't include all the expected properties we have for media items (like images, duration, feedid + preoperties we use for subscription models):

```
{
    "series_id": "abcd1234",
    "title": "Ted Lasso",
    "description": "An american football coach starts coaching a Premier League team.",
    "season_number": 1,
    "episode_number": 1
}
```

src/hooks/useSeries.ts
AntonLantukh: That may be a problem to wait for the useSeriesData hook as it just returns the object with `isLoading` state (not promise)

src/containers/SeriesSwitcher/SeriesSwitcher.tsx
budzworthy: I just mean if an error is thrown where error.code doesn't exist. I haven't used useQuery with the error typing before.

src/utils/series.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  return itemsWithEpisodes;
+};
+
+/** Next episode to show after video is complete */
+export const getNextEpisode = (series: Series, seasonNumber: number, media: PlaylistItem) => {
+  const episodes = series.seasons[seasonNumber - 1].episodes;
+
+  const episodeIndex = episodes.findIndex((el) => el.media_id === media.mediaid);
```
ChristiaanScheermeijer: The season could potentially be removed in the dashboard resulting in this throwing a type error

```suggestion
  const episodes = series.seasons[seasonNumber - 1]?.episodes;

  const episodeIndex = episodes?.findIndex((el) => el.media_id === media.mediaid);
```

src/utils/series.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  // If we have the last episode in the last season => we leave the same video
+  if (series.seasons.length === seasonNumber && episodeIndex === episodes.length - 1) {
+    return media.mediaid;
+  }
+
+  // If we have the last episode in the current season => we change the season
+  if (episodes.length - 1 === episodeIndex) {
+    return series.seasons[seasonNumber + 1].episodes[0].media_id;
```
ChristiaanScheermeijer: Potentially no episodes in a season?

```suggestion
    return series.seasons[seasonNumber + 1].episodes[0]?.media_id;
```

src/hooks/useSeries.ts
budzworthy: I guess that leaves the question as to what's better, a slow loading time where it loads successfully or a fast loading time where the user gets an error. I would vote for the former, but if the ux of the latter is nice enough I could be convinced to go that direction too.

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,41 @@
...
+        const ids = season.episodes.map((el) => el.media_id);
+        return [...acc, ...ids];
+      }, []);
+
+      const mediaItems = await getMediaByWatchlist(watchlistId, mediaIds);
```
budzworthy:
```suggestion
      const mediaItems = await getMediaByIds(mediaIds);
```

budzworthy: I think it's cleaner to use the existing multi-item call until the series functionality is ready. That way no one will run into weird behavior with reusing another watchlist.

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: I guess this is an attempt to fix an issue where the event listeners are using a stale state. 

We've also been experiencing this issue and created a different fix for this. 

The current fix will only work when a new playlist is loaded in the player but doesn't work when a different state value changes that are used in the event callbacks.

I.M.O. we should try to prevent (un)binding event listeners on every state update (we've found other issues using this method). For this reason, I suggest using the `useEventCallback` hook instead. By using this, you'll only need to bind the event listeners on player initialization and unbind on unmount (probably redundant since the player instance is destroyed on unmount).

e.g.

```ts
  const handlePlay = useEventCallback(onPlay);

  useEffect(() => {
    playerRef.current.setup({
      playlist: [deepCopy(item)],
      ...
      repeat: false,
    });

    playerRef.current.on('play', handlePlay);
    // rest
  }, []);

  useEffect(() => {
    return () => {
      if (!playerRef.current) return;
      
      playerRef.current.off('play', handlePlay);
      // rest
    };
  }, []);
```

src/hooks/useSeries.ts
AntonLantukh: Yeah, I think you are right. Let's just request some one by one for some time

src/containers/SeriesSwitcher/SeriesSwitcher.tsx
AntonLantukh: I changed it now to a general error message in case we failed to get series

src/containers/Cinema/Cinema.tsx
AntonLantukh: Thanks! This one is also to fix autoplay after the current series item is over. Right now it doesn't work

src/hooks/useSeries.ts
```diff
@@ -1,41 +1,46 @@
...
-      }, []);
+export default (seriesId: string): UseQueryResult<{ series: Series; playlist: Playlist }, ApiError> => {
+  return useQuery(
+    `series-${seriesId}`,
+    async () => {
+      const data = await getSeries(seriesId).then(async (series) => {
+        const mediaIds = (series?.seasons || []).flatMap((season: Season) => season.episodes.map((episode) => episode.media_id));
+        const mediaItems = await getMediaByIds(mediaIds);
 
-      const mediaItems = await getMediaByWatchlist(watchlistId, mediaIds);
```
budzworthy: Why the change to use promise then chaining? I think it was fine as 2 awaits before.

src/hooks/useSeriesData.ts
```diff
@@ -0,0 +1,74 @@
...
+  episodeId: string,
+): {
+  data: Data;
+  isItemError: boolean;
+  isPlaylistError: boolean;
+  isLoading: boolean;
+} => {
+  const { data: seriesData, isLoading: isSeriesLoading, error: seriesError } = useSeries(entityId);
+  // We enable it only after new series api unsuccessful load
+  const { data: playlistData, isLoading: isPlaylistLoading, error: playlistError } = usePlaylist(entityId, {}, !isSeriesLoading && Boolean(seriesError), false);
```
budzworthy:
```suggestion
  const { data: seriesData, isLoading: isSeriesLoading, error: seriesError, isError: isSeriesError } = useSeries(entityId);
  
  const usePlaylistFallback = isSeriesError && seriesError?.code === 404;
  
  // We enable it only after new series api unsuccessful load
  const { data: playlistData, isLoading: isPlaylistLoading, error: playlistError } = usePlaylist(entityId, {}, usePlaylistFallback, false);
```

src/hooks/useSeriesData.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const getData = (seriesData: { playlist: Playlist; series: Series } | undefined, playlistData: Playlist | undefined, seriesError: ApiError | null) => {
+  if (seriesError?.code === 404) {
+    return playlistData || DEFAULT_DATA;
+  }
+
+  return seriesData?.playlist || DEFAULT_DATA;
+};
+
+/** Current item we show on the Series page (we need to add episodeId and seasonId to be displayed properly) */
```
budzworthy: Since this is really related to the Series page, maybe it should go in Series.tsx

src/hooks/useSeriesData.ts
```diff
@@ -0,0 +1,74 @@
...
+  isPlaylistError: boolean;
+  isLoading: boolean;
+} => {
+  const { data: seriesData, isLoading: isSeriesLoading, error: seriesError } = useSeries(entityId);
+  // We enable it only after new series api unsuccessful load
+  const { data: playlistData, isLoading: isPlaylistLoading, error: playlistError } = usePlaylist(entityId, {}, !isSeriesLoading && Boolean(seriesError), false);
+
+  const data = getData(seriesData, playlistData, seriesError);
+
+  const { data: rawItem, isLoading: isEpisodeLoading, isError: isItemError } = useMedia(episodeId);
```
budzworthy: Moving the media call here is making the logic and return data of this hook too complicated. It was probably OK on the series page where it was.

src/hooks/useSeries.ts
```diff
@@ -0,0 +1,46 @@
...
+  return useQuery(
+    `series-${seriesId}`,
+    async () => {
+      const data = await getSeries(seriesId).then(async (series) => {
+        const mediaIds = (series?.seasons || []).flatMap((season: Season) => season.episodes.map((episode) => episode.media_id));
+        const mediaItems = await getMediaByIds(mediaIds);
+
+        return {
+          series,
+          playlist: { title: series?.title, description: series?.description, feedid: series?.series_id, playlist: enrichMediaItems(series, mediaItems) },
```
budzworthy: I think it's probably better moving forward to try to keep with the Series data structure instead of the playlist.

src/hooks/useSeriesData.ts
```diff
@@ -0,0 +1,74 @@
...
+import useMedia from '#src/hooks/useMedia';
+import type { ApiError } from '#src/utils/api';
+import type { Playlist, PlaylistItem } from '#types/playlist';
+import type { Series } from '#types/series';
+import { getSeriesIdFromEpisode } from '#src/utils/media';
+import { getNextItemId } from '#src/utils/series';
+
+const DEFAULT_DATA = { title: '', playlist: [] };
+
+const getData = (seriesData: { playlist: Playlist; series: Series } | undefined, playlistData: Playlist | undefined, seriesError: ApiError | null) => {
```
budzworthy: I don't quite get what the logic here is, especially with the default data.

src/hooks/useSeriesData.ts
```diff
@@ -0,0 +1,74 @@
+import usePlaylist from '#src/hooks/usePlaylist';
```
budzworthy, line 1: I think you could move the media item stuff out of this and back to the Series component and then combine the remaining logic into the useSeries hook instead of having 2 hooks.

src/hooks/useSeriesData.ts
```diff
@@ -0,0 +1,74 @@
...
+const getData = (seriesData: { playlist: Playlist; series: Series } | undefined, playlistData: Playlist | undefined, seriesError: ApiError | null) => {
+  if (seriesError?.code === 404) {
+    return playlistData || DEFAULT_DATA;
+  }
+
+  return seriesData?.playlist || DEFAULT_DATA;
+};
+
+/** Current item we show on the Series page (we need to add episodeId and seasonId to be displayed properly) */
+const enrichMediaItem = (mediaItems: PlaylistItem[], item: PlaylistItem | undefined): PlaylistItem | undefined => {
```
budzworthy: This is more or less duplicated from the other hook, no?

src/services/api.service.ts
```diff
@@ -120,3 +106,21 @@ export const getMediaByIds = async (ids: string[], tokens?: Record<string, strin
...
 };
+
+/**
+ * Get series by id
+ * @param {string} id
+ * @param params
+ */
+export const getSeries = async (id?: string, params: GetSeriesParams = {}): Promise<Series | undefined> => {
+  if (!id) {
+    return undefined;
```
budzworthy:
```suggestion
    throw "Series ID is required";
```

src/services/api.service.ts
```diff
@@ -120,3 +106,21 @@ export const getMediaByIds = async (ids: string[], tokens?: Record<string, strin
 
   return responses.map((result) => (result.status === 'fulfilled' ? result.value : null)).filter(notEmpty);
 };
+
+/**
+ * Get series by id
+ * @param {string} id
+ * @param params
+ */
+export const getSeries = async (id?: string, params: GetSeriesParams = {}): Promise<Series | undefined> => {
```
budzworthy:
```suggestion
export const getSeries = async (id: string, params: GetSeriesParams = {}): Promise<Series | undefined> => {
```

src/screens/SeriesNew/SeriesNew.tsx
AntonLantukh: Switched to requesting media items one by one

src/hooks/useSeriesData.ts
AntonLantukh: The idea was to use a default playlist value. I changed the logic a bit and got rid of the getData function

src/hooks/useSeriesData.ts
AntonLantukh: I reused the present `enrichMediaItems` function and deleted this one. We can just pass an array with only one element.

src/hooks/useSeries.ts
AntonLantukh: I am not sure that it is possible now, as Series don't have all the fields we need for `CardGrid` which is also used for ordinary playlists.

AntonLantukh:
@dbudzins Danny, I moved media item hook and connected utils to Series component so that `useSeriesData` is responsible now only for Series retrieval. I liked the fact that we could isolate all the 'if conditions' and series logic in one place, so that it could be easier to update it or to clean the old approach.

src/hooks/useSeries.ts
budzworthy: I was assuming that the Series component would use the seasons and/or the episodes properties of the series object, taking advantage of the hierarchy that the series service gives us.  This would also simplify things like enumerating the seasons for the seasons filters, but I see that it's possible to leave those parts the way they are now using a flat playlist. It would be nice in the long run if we mutate the fallback playlist to look like a series and update everything to use the series structure, but we can postpone that as a separate issue so that we can get this merged.

src/config.ts
```diff
@@ -1,5 +1,3 @@
-export const API_BASE_URL = 'https://content.jwplatform.com';
```
budzworthy, line 1: If these are the same, let's drop the CONFIG from the naming and just call it `APP_API_HOST` for now (the only reason I made it config specific was that I wasn't sure if cdn.jwplayer.com and content.jwplatform.com were the same)

budzworthy (approved):
Some comments and some suggestions, but I think it's good

Issue #34, opened 2022-04-20
Release OTT App for App Config Previews


budzworthy:
On the OTT-App Config page, there is a link that allows you to preview your app config. This link essentially points to a hosted web app that can dynamically load the specific app config.

AC

Create deployment with latest version of the ott-web-app.
hosted domain app-config-preview.jwplayer.com
Make sure the app-config can be dynamically loaded based on a query param.

marcovandeveen:
Done. See docs: https://github.com/jwplayer/ott-web-app/blob/develop/docs/configuration.md

meriseprince:
hi, please help my app is giving this error after deployment. 

Invalid or missing settings
There was a problem loading the default settings. Try again later. If the problem persists contact technical support.

chrome console printed this error too :

Error: No usable config sources
    at Object.B1 [as queryFn] (index-306f2f52.js:1:205274)

budzworthy:
You need to provide a default config to load: https://github.com/jwplayer/ott-web-app/blob/develop/docs/initialization-file.md

This usually comes from an App Config that you create with your JW Player account: https://docs.jwplayer.com/platform/docs/apps-app-config-overview

meriseprince:
can i have an example files? cause i followed all these steps.
in the build/public/ files i get the .webapp.ini with theses two lines 

defaultConfigSource=xxxxxxxx
playerLicenseKey=xxxxxxxx


just like that

budzworthy:
The config won't be created automatically, you need to do that through the JWP dashboard. Do you have a JWP account? If not, you can look at https://jwplayer.com/broadcasters/ for more information.

meriseprince:
Hi have my account at jwplayer dashboard. i created App config and player too. account informations is set to .webapp.ini config.json url and the player ID is set inside but seem to no work

budzworthy:
Hm, if you are using the full url for the config.json, instead try using just the ID for the config source? i.e:
`defaultConfigSource = gnnuzabk`

Also make sure the the right .ini file is being copied to the build output that you are hosting on your actual server.

meriseprince:
yes from .webapp.prod.ini

i also tried  defaultConfigSource = gnnuzabk

defaultConfigSource=https://cdn.jwplayer.com/apps/configs/XXXXXX.json
playerLicenseKey=xxxxxxxx

or like that

defaultConfigSource = [xxxxxxxx](https://cdn.jwplayer.com/apps/configs/XXXXXX.json)
[...]

PR #105, opened 2022-07-05
Dbudzinski/preview mode

budzworthy:
## Description

This PR solves #34.

### Steps completed:

- [x] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [x] Unit tests passing
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


docs/configuration.md
```diff
@@ -1,45 +1,34 @@
 # Configuration
 
-JW OTT Webapp uses a JSON configuration file to store all configuration parameters. This file can be located at the following location: `./public/config.json`.
+JW OTT Webapp uses a JSON configuration file to store all configuration parameters. This file can be located at the following location: [`./public/config.json`](public/config.json).
 
-## Dynamic Configuration
+## Dynamic Configuration File Sources
 
-In the `public/index.html` file, a small script is added to allow switching configurations based on a URL search parameter.
+Using environment variables for build parameters, you can adjust what config file the application loads at startup and which if any it will allow to be set using the `c=<config source>` query param. The location can be specified using either the 8-character ID of the config from the dashboard (i.e. `gnnuzabk`), in which case the file will be loaded from the JW Player App Config delivery endpoint, or a relative (i.e. `/config.json`) or absolute (i.e. `https://cdn.jwplayer.com/apps/configs/gnnuzabk.json`) path, in which case the file will be loaded using fetch to make a 'get' request.
```
marcovandeveen: Suggestions:  
- Add a few commas for readability: "...loads at startup and which, if any, it..."
- Mention JW Player dashboard for clarity: "of the config from the JW Player dashboard"

marcovandeveen (commented):
A powerful and well thought-through solution, supporting
- Preview from JW Dashboard
- Test/production setup (in a secure way)
- Preparing major config changes (in a secure way)

I understand that supporting configs from a complete site or property is not feasible.  Which would be nice to have.

docs/configuration.md
marcovandeveen: Cool that we also support absolute paths from another, non-jw endpoint.

types/env.d.ts
```diff
@@ -1,8 +1,11 @@
 /// <reference types="vite/client" />
 
 interface ImportMetaEnv {
-  readonly VITE_APP_TITLE: string;
+  readonly APP_TITLE: string;
   readonly APP_GITHUB_PUBLIC_BASE_URL: string;
+  readonly APP_CONFIG_DEFAULT_SOURCE: string;
+  readonly APP_CONFIG_ALLOWED_SOURCES: string | undefined;
+  readonly APP_UNSAFE_ALLOW_DYNAMIC_CONFIG: boolean | undefined;
```
ChristiaanScheermeijer: Env properties are always typed `string | undefined`

```suggestion
  readonly APP_UNSAFE_ALLOW_DYNAMIC_CONFIG: string | undefined;
```

src/utils/configOverride.ts
```diff
@@ -1,34 +1,84 @@
...
+  return allowedSources.indexOf(source) >= 0;
+}
+
+function formatSourceLocation(source?: string) {
+  if (!source) {
+    return undefined;
+  }
+
+  if (source.match(/^[a-z,\d]{8}$/)) {
+    return `https://cdn.jwplayer.com/apps/configs/${source}.json`;
```
AntonLantukh: For dev environment the url will look like: 

```
https://content-portal.jwplatform.com/apps/configs/_ID_.json
```
Maybe we will be able just to change it in case we have APP_DEV_API env on.

.env
```diff
@@ -0,0 +1,3 @@
+APP_CONFIG_DEFAULT_SOURCE=/config.json
+APP_CONFIG_ALLOWED_SOURCES=gnnuzabk
+APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=false
```
AntonLantukh: Will we also use this value for the preview button on the App Config dashboard page?

.github/workflows/firebase-preview.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build
-        run: yarn && yarn build
+        run: yarn && APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1 yarn build
```
AntonLantukh: Hm, I didn't manage to use `test--no-cleeng` config for the preview build :(

Screenshot 2022-07-06 at 17 30 09

budzworthy:
> A powerful and well thought-through solution, supporting
> 
> * Preview from JW Dashboard
> * Test/production setup (in a secure way)
> * Preparing major config changes (in a secure way)
> 
> I understand that supporting configs from a complete site or property is not feasible. Which would be nice to have.

@marcovandeveen allowing all of a property is not possible at the moment because property ID isn't part of the api route or response data (though those can certainly be added.) We'd have to require specifying if the given ID is a property or individual config.

As far as allowing an entire domain, that's doable. In fact, there's no reason we can't allow regex, which is probably the easiest and most flexible way to solve this.  So you could add something like this to the list: `/^http://www.test.com/.*\.json$/` and it would allow `http://www.test.com/myconfig.json`, `http://www.test.com/someotherconfig.json`, etc.

.github/workflows/firebase-preview.yml
budzworthy: I think this is because the test data is only bundled with dev/test builds and the preview / github builds are production environment

.env
budzworthy: Ya, that was my plan. It's our owned deployment and it's clearly billed as a demo, so even if someone puts a bunch of crazy content on it, it's no different than them putting that content on our media infrastructure as well.

.github/workflows/firebase-preview.yml
AntonLantukh: Should we maybe use dynamic configs (only ones we already have) for previews in the PR? That would help much to test different cases.

src/utils/configOverride.ts
budzworthy: This is a function that I hadn't considered. Makes sense for it to be a param, but I will probably default it to prod everywhere, since that's generally what we've used for dev and tests

.github/workflows/firebase-preview.yml
budzworthy: Good thought. I will add it as another param that we can control

AntonLantukh (approved):
🔥

PR #107, opened 2022-07-07
feat(project): sanitize html content

### LINKED ISSUE #47: sanitize publisher consent label to prevent xss
AC:

resolve following TODOs:
https://github.com/jwplayer/ott-web-app/blob/develop/src/components/Account/Account.tsx#L61
https://github.com/jwplayer/ott-web-app/blob/develop/src/components/RegistrationForm/RegistrationForm.tsx#L57

AntonLantukh:
## Description

- dompurify package added to sanitze content

Solves #47 

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/src/components/RegistrationForm/RegistrationForm.tsx
+++ b/src/components/RegistrationForm/RegistrationForm.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useHistory } from 'react-router';
 import { useTranslation } from 'react-i18next';
+import DOMPurify from 'dompurify';
 
 import useToggle from '../../hooks/useToggle';
 import { addQueryParam } from '../../utils/history';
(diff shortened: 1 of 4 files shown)


budzworthy (approved):

src/components/RegistrationForm/RegistrationForm.tsx
```diff
@@ -56,9 +57,9 @@ const RegistrationForm: React.FC<Props> = ({
   const history = useHistory();
 
   const formatConsentLabel = (label: string): string | JSX.Element => {
-    // @todo sanitize consent label to prevent XSS
-    const hasHrefOpenTag = /<a(.|\n)*?>/.test(label);
-    const hasHrefCloseTag = /<\/a(.|\n)*?>/.test(label);
+    const sanitizedLabel = DOMPurify.sanitize(label);
```
ChristiaanScheermeijer, line 60: Only `<a />` tags are allowed, perhaps limit the sanitizer to only accept that?

```suggestion
    const sanitizedLabel = DOMPurify.sanitize(label, { ALLOWED_TAGS: ['a'] });
```

AntonLantukh: Hm, we get this value from the back-end side and it seems that other tags are also available to use.

budzworthy: Tbh, I almost think we can skip sanitizing this at all... It's coming directly from what a company has put into their cleeng configurations. If they want to put a script or an iframe or something, that's up to them.

ChristiaanScheermeijer: Hmm, true. I guess the only way to misuse this is by doing a man-in-the-middle attack and somehow altering the response.

AntonLantukh: Let's have some basic sanitizing here. In case we get requests from customers to remove it or understand that it is not the best solution, we will remove it.

PR #108, opened 2022-07-08
chore(project): change provider to store

### LINKED ISSUE #88: Use store to retrieve config data
Right now there are two approaches we apply to deal with config retrieval: zustand store and context (`useContext(ConfigContext)`).
We need to use only one and to get a config from the store.

AntonLantukh:
## Description

- use only store approach to retrieve information from config

This PR solves #88.

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/providers/ConfigProvider.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { createContext, FunctionComponent, ReactNode, useEffect, useState } from 'react';
```
budzworthy, line 1: I think this file can be moved out of a component and just into .ts code (maybe like `utils/config.ts`), returning an "isLoading", which App.tsx can use to either show the loader or show the Router.  Having a provider to do this (that's not really a React provider at all anymore) might confuse anyone new to the project.

budzworthy (commented):
Thanks for doing this! 1 small point to consider with the remnants of the provider component, but otherwise, nice to have this done.

AntonLantukh:
@dbudzins Danny, thanks for the feedback. I moved it to utils and also added config retrieval to it.

budzworthy (approved):

PR #112, opened 2022-07-13
fix(auth): check the free value

### LINKED ISSUE #103: Ambiguous interpretation of 'free'
The access attribute`free` has ambiguous behavior
- `""` --> blocked
- `false` --> open to anonymous users 
- `true` --> open to anonymous users
- `False`  --> open to anonymous users 

NOT TESTED:
- 'FALSE'
- 'True'
- 'TRUE"

Note: this also impact the entitlement service of videodock

AntonLantukh:
## Description

- for the params we expect to be true, we accept following values:
  - boolean `true`
  - `'true'` / `'yes'` strings in lower / upper case
  - `1` number

- for the params we expect to be false, we accept following values:
  - boolean `false`
  - `'false'` / `'no'` strings in lower / upper case
  - `0` number

`requiresSubscription ` is a deprecated property now.

This PR solves #103 .

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/utils/common.ts
```diff
@@ -83,6 +83,67 @@ export function getOverrideIP() {
...
 }
 
+/**
+ * Here we check that the value in the custom param can be considered to be True:
+ * 1. Boolean true value
+ * 2. Any number !== 0
+ * 3. 'yes' or 'true' string in lower / upper case
+ */
+export const hasTrueValue = (value: undefined | null | string | number | boolean): boolean => {
+  // null, undefined or empty string
```
budzworthy: I think you could follow a similar pattern to this and it's a little more readable and maintainable:
https://github.com/jwplayer/ott-web-app/blob/6098e0f094ffa770c7a224b902bbf2d4d0ba8e8b/src/services/config.service.ts#L118

src/utils/common.ts
```diff
@@ -83,6 +83,67 @@ export function getOverrideIP() {
     .trim();
 }
 
+/**
+ * Here we check that the value in the custom param can be considered to be True:
+ * 1. Boolean true value
+ * 2. Any number !== 0
+ * 3. 'yes' or 'true' string in lower / upper case
+ */
+export const hasTrueValue = (value: undefined | null | string | number | boolean): boolean => {
```
budzworthy: Maybe value should just be `unknown`?

src/utils/common.ts
```diff
@@ -83,6 +83,67 @@ export function getOverrideIP() {
...
+    return Number(value) > 0;
+  }
+
+  const strValue = value.toLowerCase();
+
+  return strValue === 'true' || strValue === 'yes';
+};
+
+/**
+ * @deprecated We should use true and hasTrueValue check for new properties
```
budzworthy: I wouldn't mark the code for this as deprecated. The _usage_ of this for hasSubscription is deprecated, but the code itself is fine.

src/utils/common.ts
budzworthy: I would suggest a more verbose name like `isTruthyCustomParamValue` or something like that.

src/utils/common.ts
AntonLantukh: It depends on the purpose of the function. My intention was to make it more universal to use not only for custom params. Custom params accept only strings now (just checked it), so the approach with `indexOf` will work.
But maybe it is better just to solve custom params problem and not make it more complicated.

budzworthy: If you take in `unknown` and use `String(value)` it should work

src/utils/common.ts
AntonLantukh: My suggestion was to use only one approach: either `isTruthyCustomParamValue` or `isFalsyCustomParamValue`.
I think it may be confusing if we use an opposite (false) value for some cases. It seems to me that 'True' could solve all our cases, so instead of `requiresSubscription` we use `free` and instead of `useRealService` we use `useSandbox`.

budzworthy: Based off of the check in today, it sounds like we can get rid of requires subscription so also can get rid of this method. Though I will say, there is something to be said for having both. Because we are being careful about what means truthy or falsey, they aren't exact opposites.  Example 'Danny' is neither truthy nor falsey. So if we have a case like requires subscription where it's a false value, it won't be a good solution to just put !isTruthy.

budzworthy (approved):
Looks great! Nice use of includes instead of indexOf.

PR #118, opened 2022-07-19
Feat / add EPG service and data fixtures

ChristiaanScheermeijer:
## Description

This is the first part of the EpgService with unit tests. It accepts media items as input and fetches all EPG schedules when defined.

The external data schema looks like (not needed to be defined in typings):

```ts
type ChapterPointCustomProperty = {
  key: string;
  value: string;
};

type IncomingEpgProgram = {
  id: string;
  title: string;
  startTime: string;
  endTime: string;
  chapterPointCustomProperties: ChapterPointCustomProperty[];
};
```

The above data is validated and transformed to the following schema:

```ts
type EpgProgram = {
  id: string;
  title: string;
  startTime: string;
  endTime: string;
  description?: string;
  image?: string;
};
```

There will be changes here when starting working on other features.

### DIFF
--- a/src/fixtures/schedule.json
+++ b/src/fixtures/schedule.json
@@ -146,6 +146,57 @@
       {
         "key": "image",
         "value": "https://www.themoviedb.org/t/p/w1066_and_h600_bestv2/akuD37ySZGUkXR7LNb1oOHCboy8.jpg"
+      },
+      {
+        "key": "other-property",
+        "value": "This property should be ignored"
+      }
+    ]
+  },
+  {
+    "id": "cbec3682-eae8-44d4-9cc1-a97b3e95413c",
+    "title": "Empty properties",
+    "startTime": "2022-07-15T07:00:00",
+    "endTime": "2022-07-15T07:40:00",
+    "chapterPointCustomProperties": []
+  },
+  {
+    "id": "c909c1ba-ba8f-485f-880e-12974b82819d",
+    "title": "No image",
+    "startTime": "2022-07-15T07:00:00",
+    "endTime": "2022-07-15T07:40:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "description",
+        "value": "Item description"
+      }
+    ]
+  },
+  {
+    "id": "6ab2c86e-d435-4311-a3bf-12a96c5f6603",
+    "title": "No description",
+    "startTime": "2022-07-15T07:00:00",
+    "endTime": "2022-07-15T07:40:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "image",
+        "value": "https://cdn.jwplayer.com/image.jpg"
+      }
+    ]
+  },
+  {
+    "id": "9507d674-506f-415f-85b6-cf32baaa2c2f",
+    "title": "Empty image and description",
+    "startTime": "2022-07-15T07:00:00",
+    "endTime": "2022-07-15T07:40:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "description",
+        "value": ""
+      },
+      {
+        "key": "image",
+        "value": ""
       }
     ]
   },
--- a/src/fixtures/schedule.json
+++ b/src/fixtures/schedule.json
@@ -1,0 +1,173 @@
+[
+  {
+    "id": "00990617-c229-4d1e-b341-7fe100b36b3c",
+    "title": "Peaky Blinders",
+    "startTime": "2022-07-14T23:50:00",
+    "endTime": "2022-07-15T00:55:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "description",
+        "value": "A gangster family epic set in 1900s England, centering on a gang who sew razor blades in the peaks of their caps, and their fierce boss Tommy Shelby."
+      },
+      {
+        "key": "image",
+        "value": "https://www.themoviedb.org/t/p/w1066_and_h600_bestv2/wiE9doxiLwq3WCGamDIOb2PqBqc.jpg"
+      }
+    ]
+  },
+  {
+    "id": "45a70c27-1681-4da3-ad3b-73fa7855ee6a",
+    "title": "Friends",
+    "startTime": "2022-07-15T00:55:00",
+    "endTime": "2022-07-15T02:35:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "description",
+        "value": "Follows the personal and professional lives of six twenty to thirty-something-year-old friends living in Manhattan."
+      },
+      {
+        "key": "image",
+        "value": "https://www.themoviedb.org/t/p/w1066_and_h600_bestv2/l0qVZIpXtIo7km9u5Yqh0nKPOr5.jpg"
+      }
+    ]
+  },
+  {
+    "id": "d7846aa7-cd76-49eb-be7c-2e183a920d9e",
+    "title": "Malcolm in the Middle",
+    "startTime": "2022-07-15T02:35:00",
+    "endTime": "2022-07-15T03:30:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "description",
+        "value": "A gifted young teen tries to survive life with his dimwitted, dysfunctional family."
+      },
+      {
+        "key": "image",
+        "value": "https://www.themoviedb.org/t/p/w1066_and_h600_bestv2/ysaA0BInz4071p3LKqAQnWKZCsK.jpg"
+      }
+    ]
+  },
+  {
+    "id": "03c2932e-8590-4ab1-bff5-8894558c34d1",
+    "title": "The Simpsons",
+    "startTime": "2022-07-15T03:30:00",
+    "endTime": "2022-07-15T04:15:00",
+    "chapterPointCustomProperties": [
+      {
+        "key": "description",
+        "value": "The satiric adventures of a working-class family in the misfit city of Springfield."
+      },
+      {
+        "key": "image",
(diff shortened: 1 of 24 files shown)


src/fixtures/schedule.json
```diff
@@ -0,0 +1,173 @@
...
+    "title": "Euphoria",
+    "startTime": "2022-07-15T06:00:00",
+    "endTime": "2022-07-15T07:00:00"
+  },
+  {
+    "id": "0c3405b6-6e5d-4800-8dc1-2ed75ec410d8",
+    "title": "The Flash",
+    "startTime": "2022-07-15T07:00:00",
+    "endTime": "2022-07-15T07:40:00",
+    "chapterPointCustomProperties": [
```
budzworthy, line 141: Might be nice to add an entry with `chapterPointCustomProperties` as an empty array and with only `description`, only `image`, blank `description` and `image`, and maybe some extra custom properties to cover all the permutations we're expecting.

src/services/epg.service.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    const customProperties = incomingProgram.chapterPointCustomProperties;
+
+    // map customProperties
+    if (customProperties) {
+      transformed.image = customProperties.find((item) => item.key === 'image')?.value || '';
+      transformed.description = customProperties.find((item) => item.key === 'description')?.value || '';
+    }
+
+    return transformed;
```
dbudzins, line 60:
```suggestion
    return {
      id: incomingProgram.id,
      title: incomingProgram.title,
      startTime: incomingProgram.startTime,
      endTime: incomingProgram.endTime,
      image: incomingProgram.chapterPointCustomProperties?.find((item) => item.key === 'image')?.value || undefined,
      description: incomingProgram.chapterPointCustomProperties?.find((item) => item.key === 'description')?.value || undefined;
    };
```

budzworthy:
```suggestion
    const program = await epgProgramSchema.validate(data);
    return {
      id: program.id,
      title: program.title,
      startTime: program.startTime,
      endTime: program.endTime,
      image: program.chapterPointCustomProperties?.find((item) => item.key === 'image')?.value || undefined,
      description: program.chapterPointCustomProperties?.find((item) => item.key === 'description')?.value || undefined
    };
```

budzworthy: Simplifying a little to avoid having to get creative with local variable naming

src/services/epg.service.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Ensure the given data validates to the EpgProgram schema
+   */
+  async parseSchedule(data: unknown) {
+    if (!Array.isArray(data)) return [];
+
+    const transformResults = await Promise.allSettled(data.map((program) => this.transformProgram(program)));
+
+    return transformResults.filter(isFulfilled).map((result) => result.value);
```
budzworthy, line 91: Looks like this will swallow any errors from invalid parsing. Should we rethrow an error if any are rejected?

src/services/epg.service.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Fetch the schedule data for the given PlaylistItem
+   */
+  async fetchSchedule(item: PlaylistItem) {
+    if (!item.scheduleUrl) return [];
```
budzworthy: I would suggest returning undefined here so we can tell the difference between something that doesn't have a schedule defined (no schedule url) vs something that may have made a successful schedule query, but it returned an empty schedule.

src/services/epg.service.ts
```diff
@@ -0,0 +1,114 @@
...
+   * Fetch and parse the EPG schedule for the given PlaylistItem.
+   * When there is no program (empty schedule) or the request fails, it returns a static program.
+   */
+  async getSchedule(item: PlaylistItem) {
+    let programs: EpgProgram[] = [];
+
+    try {
+      const schedule = await this.fetchSchedule(item);
+      programs = await this.parseSchedule(schedule);
+    } catch (error: unknown) {
```
budzworthy: I assume this try-catch is temporary? If so can we just put a todo on it?

src/services/epg.service.ts
```diff
@@ -0,0 +1,114 @@
...
+      logDev(error);
+    }
+
+    return {
+      title: item.title,
+      programs,
+    } as EpgChannel;
```
budzworthy:
```suggestion
    const channel: EpgChannel = {
        title: item.title,
        programs: [],
    };

    try {
      const schedule = await this.fetchSchedule(item);
      channel.programs = await this.parseSchedule(schedule);
    } catch (error: unknown) {
      logDev(error);
    }

    return channel;
```

budzworthy (commented):
Looking good. Some small suggestions and questions, but I think you've made a good start of the project.

src/fixtures/schedule.json
ChristiaanScheermeijer: Good points, I will add those 👍

src/services/epg.service.ts
ChristiaanScheermeijer: That's much better indeed!

src/services/epg.service.ts
AntonLantukh: Good point

ChristiaanScheermeijer: I did this on purpose so that we only remove invalid programs from the schedule. The problem is that there are no indicators that there might be a problem. Do you have a suggestion?

src/services/epg.service.ts
ChristiaanScheermeijer: I did this in preparation for the "single program" fallback schedule when the schedule fails to load or is empty. But we can detect that with `undefined` as well.

src/services/epg.service.ts
ChristiaanScheermeijer: `fetchSchedule` could throw an error since it is returning the `getDataOrThrow` function. Do you suggest moving the error catching to the `fetchSchedule` function?

src/services/epg.service.ts
ChristiaanScheermeijer: Good suggestion, thanks!

src/services/epg.service.ts
ChristiaanScheermeijer: I've moved the try/catch to the `fetchSchedule` function.

src/services/epg.service.ts
budzworthy: I guess we should check with @marcovandeveen do decide how strict or forgiving we want to be with the parsing. If a program is invalid we have all the following options:
1. Silently drop it (better UX, but hard to debug)
2. Try to cast it as best we can (maybe most properties are valid and we can still use it, but might lead to unexpected results)
3. Show an error for the whole schedule (bad UX, but easier to debug)
4. Just show an error for the specific program (might be hard to display if we don't have valid start and end times for example.)

I think my votes would be for number 3 in dev and number 2 in prod.

src/services/epg.service.ts
```diff
@@ -0,0 +1,108 @@
...
+   */
+  async fetchSchedule(item: PlaylistItem) {
+    if (!item.scheduleUrl) return undefined;
+
+    try {
+      const response = await fetch(item.scheduleUrl);
+
+      // await needed to ensure the error is caught here
+      return await getDataOrThrow(response);
+    } catch (error: unknown) {
```
budzworthy, line 117: This is still going to cause problems I think.  First, if we end up calling this via a useQuery hook, the retries won't work, so if the fetch fails for a networking hiccup we don't recover gracefully.

Secondly, I think I would prefer to show an error in place of the schedule when it fails to load instead of showing a solid bar of the current program, which probably is misleading.  Something simple like 'Failed to load program guide. Please refresh to try again.'

src/services/epg.service.ts
ChristiaanScheermeijer: Ok, we can still change this behavior if @marcovandeveen agrees

src/services/epg.service.ts
ChristiaanScheermeijer: That's a valid point. Then we might only do the static program (#120) when the schedule successfully loads, but the schedule is an empty array. 

When something is wrong, we render this in the EPG instead.

@marcovandeveen what do you think?

src/services/epg.service.ts
budzworthy: I'd also be fine with number 1 as is as long as we throw a logDev in for debugging

src/services/epg.service.ts
budzworthy: Maybe this is still the same static schedule, just with a different message?

ChristiaanScheermeijer: Yes, I like that idea as well. Something like:

Empty schedule: "No program information"
Failed request: "Failed to load schedule"
Network error: "Failed to load schedule" (can even be more concrete for users when the network drops)

src/services/epg.service.ts
ChristiaanScheermeijer: I actually had that coded, but removed it later 😅

budzworthy: Cool, let's go with that then unless Marco says otherwise

src/services/epg.service.ts
budzworthy: Looks good to me

ChristiaanScheermeijer: Cool, I will process this feedback in #120

marcovandeveen: Agreed:
Empty schedule: "No program information"
Failed request: "Failed to load schedule"
Network error: "Failed to load schedule" (can even be more concrete for users when the network drops)

budzworthy (approved):

PR #120, opened 2022-07-20
Feat / fill schedule with static program when empty

ChristiaanScheermeijer:
## Description

This PR adds a static program in one of the following scenarios:

When the schedule is empty it creates a static program with the title: "No program information"
When the schedule request fails it creates a static program with the title: "Failed to load schedule"

I've also fixed an issue in the translation mocking and a few minor changes after I ran `$ yarn i18next` which scans for translation keys.

### DIFF
--- a/src/services/epg.service.test.ts
+++ b/src/services/epg.service.test.ts
@@ -5,6 +5,7 @@ import epgService, { EpgProgram } from '#src/services/epg.service';
 import scheduleFixture from '#src/fixtures/schedule.json';
 import livePlaylistFixture from '#src/fixtures/livePlaylist.json';
 import type { Playlist } from '#types/playlist';
+import { addDays, endOfDay, startOfDay, subDays } from '#src/utils/datetime';
 
 const livePlaylist = livePlaylistFixture as Playlist;
 const scheduleData = scheduleFixture as EpgProgram[];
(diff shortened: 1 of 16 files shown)


src/services/epg.service.test.ts
```diff
@@ -35,33 +36,88 @@ describe('epgService', () => {
...
+
+    expect(channel3Mock).toHaveFetched();
+    expect(schedule3.title).toEqual('Channel 3');
+    expect(schedule3.programs.length).toEqual(1);
+    expect(schedule3.programs[0]).toMatchObject({
+      id: 'no-program',
+      title: 'No program',
+      description: 'There is no information available for this program.',
+      startTime: subDays(startOfDay(), 1).toJSON(),
+      endTime: addDays(endOfDay(), 1).toJSON(),
```
budzworthy: A small point, but reusing the day functions here means that any bugs in the day functions won't be caught. Maybe it's more thorough to [mock the date](https://vitest.dev/guide/mocking.html#dates) and hard code the expected values here?

budzworthy (approved):
One thing about the tests, but not a blocker

src/services/epg.service.test.ts
ChristiaanScheermeijer: Yes, that's a really good point. I will try date mocking real quick

ChristiaanScheermeijer: Done! ✅

budzworthy (approved):

PR #124, opened 2022-07-21
Feat / add live channels layout

ChristiaanScheermeijer:
This is a proposal refactor which makes it possible to re-use most of the existing components for the live channels layout. 

**Changes:**

- Defer app loading state until services are initialized
  - This fixes some issues where the watch history wasn't fully loaded before loading the movie page
- Moved `containers/Playlist/PlaylistContainer.tsx` to `containers/PlaylistContainer/PlaylistContainer.tsx` to prevent confusion
- Extracted player logic and rendering from the `<Cinema />` container to a separate `<Player />` component
  - This is needed to make the player re-usable for other components
- Extracted all video logic and rendering from the `<Video />` component to the `<Cinema />` container
- Created `<TrailerModal />` container which renders a player inside a modal and uses the `<Player />` component
- Simplified the `<Video />` component to work with composed and generated props to make the component more flexible
- Moved `<Cinema />` and `<TrailerModal />` containers to screen level (containers shouldn't be rendered in a component)
- Created formatting utils for video and series metadata strings
- Use JWPlayers `playlistItem#starttime` property to continue watching instead of a manual seek


src/App.tsx
```diff
@@ -60,8 +60,8 @@ class App extends Component {
   };
 
   configValidationCompletedHandler = async (config: Config) => {
-    this.setState({ isLoading: false });
     await this.initializeServices(config);
+    this.setState({ isLoading: false });
```
ChristiaanScheermeijer, line 64: This is causing flaky continue watching tests. Sometimes the watchlist is still loading while navigating to a video page.

src/components/Player/Player.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import Player from './Player';
+
+import type { PlaylistItem } from '#types/playlist';
+
+describe('<Cinema>', () => {
```
ChristiaanScheermeijer: rename

src/components/Video/Video.tsx
```diff
@@ -1,117 +1,76 @@
...
   const breakpoint: Breakpoint = useBreakpoint();
   const { t } = useTranslation(['video', 'common']);
 
-  const handleUserActive = useCallback(() => setUserActive(true), []);
-  const handleUserInactive = useCallback(() => setUserActive(false), []);
-  const handlePlay = useCallback(() => setIsPlaying(true), []);
-  const handlePause = useCallback(() => setIsPlaying(false), []);
-  const handleComplete = useCallback(() => onComplete && onComplete(), [onComplete]);
-
   const { clearWarning, warning } = useFavoritesStore((state) => ({
```
ChristiaanScheermeijer: TODO: uses store, should be moved to parent container

src/containers/Cinema/Cinema.tsx
```diff
@@ -1,192 +1,139 @@
...
-      if (playerRef.current) {
-        // Detaching events before component unmount
-        detachEvents();
-        playerRef.current.remove();
-      }
-    };
-  }, [detachEvents]);
+    if (open) setUserActive(true);
+    if (!open) saveItem(item, getProgress());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ChristiaanScheermeijer, line 102: This is needed since we only want this effect to run when the `open` property updates

src/containers/Cinema/Cinema.tsx
```diff
@@ -1,192 +1,139 @@
...
+              <div>
+                <h2 className={styles.title}>{title}</h2>
+                <div className={styles.metaContainer}>
+                  {isSeries && (
+                    <div className={styles.seriesMeta}>
+                      <strong>{seriesMeta}</strong> - {episodeTitle}
+                    </div>
+                  )}
+                  <div className={styles.meta}>{videoMeta}</div>
+                </div>
```
ChristiaanScheermeijer: This could also be more generic e.g. `firstMetaString` and `secondMetaString`

src/utils/configLoad.ts
```diff
@@ -102,8 +102,5 @@ export const loadAndValidateConfig = async (
     })
     .catch((error: Error) => {
       onValidationError(error);
-    })
-    .finally(() => {
-      onLoading(false);
```
ChristiaanScheermeijer, line 107: This was causing the loading state to be set to `false` before the services are initialized.

src/App.tsx
budzworthy: Good find

AntonLantukh: Thank you for investigating!

src/components/Video/Video.tsx
```diff
@@ -1,117 +1,76 @@
...
 
 type Props = {
   title: string;
-  item: PlaylistItem;
-  feedId?: string;
+  description: string;
+  videoMeta: string;
+  seriesMeta?: string;
+  episodeTitle?: string;
   trailerItem?: PlaylistItem;
```
ChristiaanScheermeijer: Can be removed

src/components/Player/Player.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const handleBeforePlay = useEventCallback(onBeforePlay);
+  const handlePlay = useEventCallback(onPlay);
+  const handlePause = useEventCallback(onPause);
+  const handleComplete = useEventCallback(onComplete);
+  const handleUserActive = useEventCallback(onUserActive);
+  const handleUserInactive = useEventCallback(onUserInActive);
+  const handlePlaylistItemCallback = useEventCallback(onPlaylistItemCallback);
+  const handleReady = useEventCallback(() => onReady && onReady(playerRef.current));
+
+  const attachEvents = useCallback(() => {
```
budzworthy, line 59: Do playerRef or playerRef.current need to be a dependency to these 2 useCallback calls?

src/components/Player/Player.tsx
```diff
@@ -0,0 +1,157 @@
...
+  onComplete,
+  onUserActive,
+  onUserInActive,
+  onBeforePlay,
+  onPlaylistItemCallback,
+  feedId,
+  startTime = 0,
+}: Props) => {
+  const playerElementRef = useRef<HTMLDivElement>(null);
+  const playerRef = useRef<JWPlayer>();
```
budzworthy, line 43:
```suggestion
  const playerRef = useRef<JWPlayer>();
  const player = playerRef.current; // Use this everywhere instead of playerRef.current (except for the assignment call) and code inspections work better to pick up missing dependencies for callbacks, etc.
```

src/components/Player/Player.tsx
ChristiaanScheermeijer: Refs don't have an effect on the effect's lifecycle. Perhaps it's better for the `attachEvents` function to accept the playerInstance instead of relying on the ref so that we get typescript errors when we change this later.

The `attachEvents` and `detachEvents` functions are only used on setup and unmount.

src/components/Player/__snapshots__/Cinema.test.tsx.snap
```diff
@@ -0,0 +1,11 @@
+// Vitest Snapshot v1
+
+exports[`<Cinema> > renders and matches snapshot 1`] = `
```
budzworthy: Where is this used? Can it be deleted?

src/components/Player/Player.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { addScript } from '#src/utils/dom';
+import { deepCopy } from '#src/utils/collection';
+import type { JWPlayer } from '#types/jwplayer';
+import type { PlaylistItem } from '#types/playlist';
+import useEventCallback from '#src/hooks/useEventCallback';
+import useOttAnalytics from '#src/hooks/useOttAnalytics';
+import { logDev } from '#src/utils/common';
+
+type Props = {
+  playerId?: string | null;
```
budzworthy: Shouldn't playerId be required?  Otherwise the script url ends up being funky and the page will probably error.

src/components/Video/Video.tsx
```diff
@@ -1,117 +1,61 @@
...
-  isFavorited,
+  posterMode,
+  isFavorite,
   isFavoritesEnabled,
   onFavoriteButtonClick,
   children,
+  hasTrailer,
   playTrailer,
   onTrailerClick,
-  onTrailerClose,
```
budzworthy, line 70: onTrailerClose is unused now, so it can be removed from props

src/components/Video/Video.tsx
```diff
@@ -1,117 +1,61 @@
...
+  seriesMeta,
+  episodeTitle,
   poster,
-  enableSharing,
-  hasShared,
-  onShareClick,
-  isFavorited,
+  posterMode,
+  isFavorite,
   isFavoritesEnabled,
```
budzworthy: Suggest we pull out Favorites button the same as the Share and Start Watching buttons

src/components/Video/Video.tsx
```diff
@@ -123,14 +67,18 @@ const Video: React.FC<Props> = ({
...
+              </div>
+            )}
           </div>
-          <CollapsibleText text={item.description} className={styles.description} maxHeight={isMobile ? 60 : 'none'} />
+          <CollapsibleText text={description} className={styles.description} maxHeight={isMobile ? 60 : 'none'} />
 
           <div className={styles.buttonBar}>
             {startWatchingButton}
-            {trailerItem && (
+            {hasTrailer && (
```
budzworthy: Suggest we pull the trailer stuff out as a button as well

src/components/Video/Video.tsx
```diff
@@ -144,77 +92,19 @@ const Video: React.FC<Props> = ({
...
-            />
-            <div className={classNames(styles.playerOverlay, { [styles.hidden]: isPlaying && !userActive })} />
-            <div className={classNames(styles.trailerMeta, styles.title, { [styles.hidden]: isPlaying && !userActive })}>{`${title} - Trailer`}</div>
-            <ModalCloseButton onClick={onTrailerClose} visible={!isPlaying || userActive} />
-          </div>
-        </Modal>
-      )}
     </div>
   );
 };
```
budzworthy: Would it make sense to rename `Video` to something like `MediaDetails` / `VideoDetails` / `VideoOverview` / etc., since there is no video or player in it now?

src/containers/Cinema/Cinema.tsx
budzworthy: Will `getProgress` or `item` change though?  I think they shouldn't.  If you end up leaving the `disable-next-line`, just add the comment ("This is needed since we only want this effect to run when the open property updates") with it

src/containers/Cinema/Cinema.tsx
```diff
@@ -1,192 +1,147 @@
...
+                <ArrowLeft />
+              </IconButton>
+              <div>
+                <h2 className={styles.title}>{title}</h2>
+                <div className={styles.metaContainer}>
+                  {isSeries && (
+                    <div className={styles.seriesMeta}>
+                      <strong>{seriesMeta}</strong> - {episodeTitle}
+                    </div>
+                  )}
```
budzworthy: Do you need the isSeries property or can you just use the truthiness of seriesMeta to decide to show it?

```suggestion
                  {seriesMeta && (
                    <div className={styles.seriesMeta}>
                      <strong>{seriesMeta}</strong> - {episodeTitle}
                    </div>
                  )}
```

src/containers/Cinema/Cinema.tsx
budzworthy: What about `primaryMetadata` and `secondaryMetadata`?

src/containers/FavoritesWarningDialog/FavoritesWarningDialog.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
budzworthy: If you pull the favorites stuff out into its own button, this can be part of the button instead of its own component. As is, it's a bit awkward because you have to remember to put it with the button in order for the warnings to show up.

src/containers/Cinema/Cinema.tsx
```diff
@@ -1,192 +1,147 @@
...
+  // state
+  const [playerInstance, setPlayerInstance] = useState<JWPlayer>();
+  const [isPlaying, setIsPlaying] = useState<boolean>(false);
+  const [userActive, setUserActive] = useState(true);
 
-  const getProgress = useCallback((): number | null => {
-    if (!playerRef.current) return null;
+  // watch history
+  const watchHistoryItem = useWatchHistoryStore((state) => item && state.getItem(item));
+  const videoProgress = watchHistoryItem?.progress;
```
budzworthy: Should we still do this if `enableWatchHistory` is false?

src/containers/TrailerModal/TrailerModal.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useCallback, useState } from 'react';
```
budzworthy, line 1: It looks like there's a lot of overlap between `Cinema` and this component.  I played around with it a bit to see if I could get them into 1 and I think it's possible. Want me to submit a PR to your PR?  Or are they separate for a reason?

src/utils/configLoad.ts
budzworthy: Nice find

budzworthy (commented):
Good progress! There are a few questions, below, but feel free to push back on any of them if the additional refactoring will slow you down too much and Anton or I can get to them afterwards.

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: `item` will change when the next video starts.
`item` is a dependency of `getProgress`, so that will change as well.

I will add a comment 👍

src/components/Video/Video.tsx
ChristiaanScheermeijer: Agree! I've renamed it to `<VideoDetails />`

src/components/Player/__snapshots__/Cinema.test.tsx.snap
ChristiaanScheermeijer: It should be removed indeed

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: Are you suggesting something like this?

```tsx
  // watch history
  const watchHistoryItem = useWatchHistoryStore((state) => !!item && enableWatchHistory ? state.getItem(item) : undefined);

  const startTime = useMemo(() => {
    const videoProgress = watchHistoryItem?.progress;

    if (videoProgress && videoProgress > VideoProgressMinMax.Min && videoProgress < VideoProgressMinMax.Max) {
      return videoProgress * item.duration;
    }

    return 0;
  }, [item.duration, watchHistoryItem?.progress]);
```

src/containers/TrailerModal/TrailerModal.tsx
ChristiaanScheermeijer: I kept those separate to make the differences clear. IMHO it gets more complicated when combining functionally different components like this (we had some issues before).

The biggest problem is when changing something for the trailer player, you accidentally change it for the movie as well. 

However, I never mind looking at PRs with suggestions. Please do submit it if you have it ready.

src/containers/FavoritesWarningDialog/FavoritesWarningDialog.tsx
ChristiaanScheermeijer: Done

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: Done (using the `secondaryMetadata` prop)

src/components/Video/Video.tsx
ChristiaanScheermeijer: Done!

src/components/Video/Video.tsx
ChristiaanScheermeijer: Also done!

src/components/Player/Player.tsx
ChristiaanScheermeijer: I've updated it, but it went all the way down to the Config type and validation. Not 100% sure if this has other effects. I guess the `player` property should always be required...

ChristiaanScheermeijer:
@dbudzins @AntonLantukh I've updated the PR and processed most feedback. Could you look at these changes? I will rebase the EPG branch when approved.

Have a great weekend!

src/components/VideoDetails/VideoDetails.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import styles from './VideoDetails.module.scss';
+
+import CollapsibleText from '#src/components/CollapsibleText/CollapsibleText';
```
AntonLantukh, line 6: Is there any reason why we always repeat the folder's name in the component's file? I mean why not to use `import CollapsibleText from '#src/components/CollapsibleText';` in the feature with the index.tsx file inside the folder.

AntonLantukh: I don't suggest to change everything now :) We can do refactoring later.

ChristiaanScheermeijer: Hehe, that would mean changing every file indeed 😅  But it is a valid question. In our boilerplate project, we add an index.ts file for every component folder which re-exports the component. This is even done automatically via the generate component/container/screen command.

We can do that for sure, I don't really have an opinion on this.

budzworthy: Using an index file can sometimes make it hard to navigate since you end up with 100 files in your project named index.  If it's just boilerplate that re-exports the main component that's fine, but it's another thing to maintain, so I would recommend we setup something like [plop](https://plopjs.com/) to help us if we want to go that route.

src/components/Player/Player.tsx
budzworthy: Indeed it looks like on the dashboard we allow creation of configs without a player ID. These configs wouldn't work with the web app though (unless they forked it and hard coded a player.) @marcovandeveen do you have any idea why player ID is optional?

src/containers/Cinema/Cinema.tsx
budzworthy: Yes, nice solution without breaking the hooks!

src/containers/TrailerModal/TrailerModal.tsx
budzworthy: OK, I'll submit the PR separate from this.

src/screens/Series/Series.tsx
```diff
@@ -141,6 +122,13 @@ const Series = ({ match, location }: RouteComponentProps<SeriesRouteParams>): JS
   const pageTitle = `${item.title} - ${siteName}`;
   const canonicalUrl = seriesPlaylist && item ? `${window.location.origin}${episodeURL(seriesPlaylist, item.mediaid)}` : window.location.href;
 
+  const primaryMetadata = formatVideoMetaString(t, item, true, seriesPlaylist.playlist.length);
+  const secondaryMetadata = (
+    <>
+      <strong>{formatSeriesMetaString(item)}</strong> - {item.title}
```
dbudzins, line 128: What was the reason for not including the `strong` tag in the `VideoDetail` component?  Seems like unnecessary flexibility that might break the formatting if not used carefully.

src/components/Player/Player.tsx
budzworthy: Imho, I say we enforce it to be not null, since the web app won't work without one

src/components/Player/Player.tsx
budzworthy: OK, what I was worried about is that this seems like a place we might get memory leaks if the playerRef changes and the events don't attach or detach properly, but it seems to be working ok.

budzworthy (approved):
Let's move ahead with the player id enforced non-null until we hear otherwise

PR #125, opened 2022-07-22
chore(project): config select changes

### LINKED ISSUE #127: Demo / Dev Mode Options Menu
For the demo / dev mode we want to have easy access to switching configs and potentially other settings like the jw host (dev vs prod)

From Slack (@marcovandeveen):
Functionally I would like to see the following:

- Opening [app-preview.jwplayer.com](http://app-preview.jwplayer.com/) has a switcher
  - Use case: Demo
    - app config "Blender Subscription Demo"  (includes TVOD, Premium Free, 24x7) -- Default
    - app config "Blender Advertisement Demo"
    - app config "Blender Authentication Only Demo"
   - Suggestion: get the combo box name from the app display name
- Opening [app-preview.jwplayer.com?c=abcdef](http://app-preview.jwplayer.com/?c=abcdef) does not have a switcher
   - Use case: Preview config in JW Environment
   - Use case: Announcement/Updates (be careful as these configs need to be maintained)
- Maybe we can have another URL for dev/test? E.g. [app-dev.jwplayer.com](http://app-dev.jwplayer.com/)?
- Can we also have the config selection like this?
or what about something like [this](https://www.cssscript.com/demo/sidebar-navigation-bootstrap/). Gives us plenty of real estate for future demo / debug options.
image

AntonLantukh:
## Description

1. Dev environment supported via a query `env=dev` value, now accessible from the dashboard. When setting the query we change the `API_HOST` and use the configs from 'dev' folder.

2. Changed configs location to store them all in one place. Now all the configs are stored in a `configs` folder. We use `prod` for a live version and local development (for local we can switch to `test`). We use `dev` for our test api. We use `test` for e2e-tests and PR-previews.

3. Debug section added. Section is shown only when opening the page without `c` query. I decided to locate the icon on the bottom of the page for it not to be an eye-catcher. Section reuses `Sidebar` component with different content.

4. Configs in a select control are formatted to include several words and also a given `siteName`. The format of the file should comply with several rules to be shown in a right way.
 
5. Docs updated

6. README preview url updated

7. Env variables are stored now in one place (env file).

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


AntonLantukh:
Suggestion: move 'internal' (dev api) configs to `configs/internal` folder. Use `dev` for real development mode.

types/env.d.ts
```diff
@@ -6,6 +6,7 @@ interface ImportMetaEnv {
   readonly APP_CONFIG_DEFAULT_SOURCE: string | undefined;
   readonly APP_CONFIG_ALLOWED_SOURCES: string | undefined;
   readonly APP_UNSAFE_ALLOW_DYNAMIC_CONFIG: string | undefined;
+  readonly APP_INCLUDE_CONFIGS: 'dev' | 'test' | 'prod';
```
budzworthy, line 9: We probably don't need this included in the type definition, since it's only a build time param

.env.test
```diff
@@ -1,4 +1,5 @@
-APP_CONFIG_DEFAULT_SOURCE=test--blender
-APP_CONFIG_ALLOWED_SOURCES=
+APP_CONFIG_DEFAULT_SOURCE=default--config
 APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1
+APP_CONFIG_ALLOWED_SOURCES=0
```
budzworthy, line 3: Why the `0` here?

vite.config.ts
```diff
@@ -36,6 +36,7 @@ export default ({ mode }: { mode: 'production' | 'development' | 'test' }) => {
 
   return defineConfig({
     plugins: plugins,
+    mode: process.env.APP_MODE || mode,
```
budzworthy, line 39:
```suggestion
```

configs/test/default--config.json
```diff
@@ -0,0 +1,72 @@
+{
```
budzworthy, line 1: Does this have any functional differences from `blender--config.json`?  If not, let's drop the blender one and just keep this.

package.json
```diff
@@ -10,8 +10,9 @@
   },
   "scripts": {
     "prepare": "husky install",
-    "start": "APP_INCLUDE_TEST_CONFIGS=1 vite",
-    "start:dev-api": "APP_API_BASE_URL=https://content-portal.jwplatform.com yarn start",
+    "start": "vite",
+    "start:e2e-test": "APP_MODE=test APP_INCLUDE_CONFIGS=test vite",
```
budzworthy, line 14: Instead of making more start package scripts, I suggest we create a .env file for each mode/environment and let the .env files capture all of the variables.

.github/workflows/codeceptjs.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           yarn
           npm install wait-on -g
       - name: Start dev server
-        run: yarn start &
+        run: yarn start:e2e-test &
```
budzworthy, line 26:
```suggestion
        run: yarn start --mode test-e2e &
```

.github/workflows/firebase-live.yml
```diff
@@ -10,9 +10,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build
-        run: yarn && APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1 APP_INCLUDE_TEST_CONFIGS=1 yarn build
+        run: yarn && APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1 APP_INCLUDE_CONFIGS=prod yarn build
```
budzworthy, line 13:
```suggestion
        run: yarn && yarn build --mode demo
```

.github/workflows/firebase-preview.yml
```diff
@@ -10,9 +10,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build
-        run: yarn && APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1 APP_INCLUDE_TEST_CONFIGS=1 yarn build
+        run: yarn && APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1 APP_INCLUDE_CONFIGS=test yarn build
```
budzworthy, line 13:
```suggestion
        run: yarn && yarn build --mode demo
```

docs/configuration.md
```diff
@@ -30,6 +30,15 @@ A space separated list of 8-character IDs and/or url paths for config files that
...
 
+---
+
+**APP_INCLUDE_CONFIGS** (dev | test | prod) - We can specify which config folder to use to get local configs from. It can be useful to differentiate having several environments. Some configs can be used for testing, some for production, some for the dev environment. We differentiate them this way.
+    - `dev` these configs are used internally by JW Player developers to test new features
+    - `test` these configs are used for end-to-end testing and local development.
+    - `prod` configs in this folders are used both for production and local manual testing. They are also available on our preview page: https://app-preview.jwplayer.com/
+  
+(!) Configs have `${NAME}--config.json` format, where words are separated by `-` sign. We also add `siteName` to the beginning of each config. You can also notice that every folder has its default config which is used in case a dynamically retrieved value can't resolved (default value is set in `APP_CONFIG_DEFAULT_SOURCE` env).
```
budzworthy, line 40: You can omit the discussion about naming, since it's only really used for dev and demo with the config selector. We shouldn't be encouraging customers to try to use this.

docs/configuration.md
```diff
@@ -30,6 +30,15 @@ A space separated list of 8-character IDs and/or url paths for config files that
...
+
```
budzworthy, line 41: The description of the default behavior is a bit confusing / misleading. The APP_CONFIG_DEFAULT_SOURCE property should be the one and only determiner of a fallback.

docs/e2e.md
```diff
@@ -56,4 +56,5 @@ To serve allure reports locally `allure-commandline` package should be installed
 1. Install Java 8 (for Mac homebrew `adoptopenjdk8` package can be used)
 2. `yarn install`
 3. Install `allure-commandline` globally (can help in the future to serve downloaded artifacts)
-4. Run `yarn codecept-serve:desktop` 
+4. Run `yarn start:e2e-test`
```
budzworthy, line 59:
```suggestion
4. Run `yarn start --mode test-e2e`
```

src/containers/Layout/Layout.tsx
```diff
@@ -89,7 +92,7 @@ const Layout: FC<LayoutProps> = ({ children }) => {
     if (!cleengId) return null;
 
     return isLoggedIn ? (
-      <UserMenu showPaymentsItem={accessModel !== 'AVOD'} />
+      <UserMenu showPaymentsItem={accessModel !== 'AVOD'} onClick={loginButtonClickHandler} />
```
budzworthy, line 95: What does this onClick handler do (why was it added)?

src/containers/Layout/Layout.tsx
```diff
@@ -33,12 +33,13 @@ const Layout: FC<LayoutProps> = ({ children }) => {
   const { searchPlaylist } = features || {};
   const { footerText, dynamicBlur } = styling || {};
 
-  const { blurImage, searchQuery, searchActive, userMenuOpen } = useUIStore(
-    ({ blurImage, searchQuery, searchActive, userMenuOpen }) => ({
+  const { blurImage, searchQuery, searchActive, userMenuOpen, showDebugTools } = useUIStore(
+    ({ blurImage, searchQuery, searchActive, userMenuOpen, showDebugTools }) => ({
```
budzworthy, line 37: I would try to keep the behavior of this pop up out of the main application, so that when it's disabled it can totally disappear and it doesn't complicate anything, since it's 99% mostly used for our own dev and demo purposes.

src/env/index.ts
```diff
@@ -0,0 +1,25 @@
+import { getQueryParam } from '#src/utils/history';
```
budzworthy, line 1: You can probably put this in `utils/env` instead of creating a whole subdirectory for `env`

src/env/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const IS_TEST_BUILD = import.meta.env.MODE === 'test';
+
+export const IS_DEV_OR_TEST_BUILD = import.meta.env.DEV;
+
+// We can use dev or prod host (dev host is used only internally by JW Player developers to test new features)
+export const API_HOST = getQueryParam('env') !== 'dev' ? import.meta.env.APP_API_BASE_URL : 'https://content-portal.jwplatform.com';
+
+// Small hack to toggle on demo environment and use its own configs
+export const INCLUDE_CONFIGS = getQueryParam('env') === 'dev' ? 'dev' : import.meta.env.APP_INCLUDE_CONFIGS;
```
budzworthy, line 17: Will this work? Aren't the configs included / excluded during build time?

src/utils/configOverride.ts
```diff
@@ -1,28 +1,23 @@
...
-  return formatSourceLocation(getConfigOverride() || DEFAULT_SOURCE);
+  return formatSourceLocation(getConfigOverride() || CONFIG_DEFAULT_SOURCE);
 }
 
 export const setStoredConfig = (value: string) => {
   storage.setItem(configFileStorageKey, value);
 };
 
 export const getStoredConfig = () => {
-  return storage.getItem(configFileStorageKey)?.toLowerCase();
```
budzworthy, line 25: It's safest to keep the lowercasing here in case the stored value is not lowercase somehow

src/utils/configOverride.ts
```diff
@@ -66,29 +67,25 @@ function getConfigOverride() {
 }
 
 function isValidConfigSource(source: string) {
-  // Dynamic values are valid as long as they are defined
-  if (UNSAFE_ALLOW_DYNAMIC_CONFIG) {
+  // Dynamic values are valid as long as they are defined, value can be 0 | 1
+  if (Number(UNSAFE_ALLOW_DYNAMIC_CONFIG)) {
```
budzworthy, line 71: This doesn't match the .env files, either add a `0` there or allow falsey here

src/utils/configOverride.ts
```diff
@@ -66,29 +67,25 @@ function getConfigOverride() {
...
     return !!source;
   }
 
-  if (INCLUDE_TEST_CONFIGS && source.startsWith('test--')) {
```
budzworthy, line 74: How will this check for allowed demo/test configs? Or I guess allow_unsafe_dynamic has to be true to use any of the demo/test configs?

src/stores/UIStore.ts
```diff
@@ -8,11 +8,13 @@ type UIState = {
   searchActive: boolean;
   userMenuOpen: boolean;
   preSearchPage?: LocationDescriptor<unknown>;
+  showDebugTools: boolean;
```
budzworthy, line 11: I would keep this out of the store so the dev component doesn't pollute the business logic of the main app

src/utils/history.ts
```diff
@@ -19,3 +19,10 @@ export function removeQueryParam(history: History, key: string): string {
 
   return `${history.location.pathname}${searchParams ? `?${searchParams}` : ''}`;
 }
+
+export function getQueryParam(value: string): string {
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const params = Object.fromEntries(urlSearchParams.entries());
+
+  return params[value];
```
budzworthy, line 27:
```suggestion
  const urlSearchParams = new URLSearchParams(window.location.search);

  return urlSearchParams.get(value);
```

src/env/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+export const IS_DEV_BUILD = import.meta.env.MODE === 'development';
+
+export const IS_TEST_BUILD = import.meta.env.MODE === 'test';
+
+export const IS_DEV_OR_TEST_BUILD = import.meta.env.DEV;
+
+// We can use dev or prod host (dev host is used only internally by JW Player developers to test new features)
+export const API_HOST = getQueryParam('env') !== 'dev' ? import.meta.env.APP_API_BASE_URL : 'https://content-portal.jwplatform.com';
```
budzworthy, line 14: I recommend we do this query param check in the same place as `configOverride` so we can do the same thing to remove it from the url and probably use storage to maintain it between postbacks

src/utils/configOverride.ts
```diff
@@ -66,29 +67,25 @@ function getConfigOverride() {
...
 
 function formatSourceLocation(source?: string) {
   if (!source) {
     return undefined;
   }
 
-  if (source.match(/^[a-z,\d]{8}$/)) {
-    return `${CONFIG_HOST}/apps/configs/${source}.json`;
+  if (['prod', 'test', 'dev'].includes(INCLUDE_CONFIGS) && source.includes('--config')) {
+    return `/configs/${INCLUDE_CONFIGS}/${source}.json`;
```
budzworthy, line 84: Aren't the files included without the `prod`, `test`, etc. directories? Does this work as intended?

src/utils/configOverride.ts
```diff
@@ -66,29 +67,25 @@ function getConfigOverride() {
...
 }
 
 function formatSourceLocation(source?: string) {
   if (!source) {
     return undefined;
   }
 
-  if (source.match(/^[a-z,\d]{8}$/)) {
-    return `${CONFIG_HOST}/apps/configs/${source}.json`;
+  if (['prod', 'test', 'dev'].includes(INCLUDE_CONFIGS) && source.includes('--config')) {
```
budzworthy, line 83: What if we use a convention that won't overlap with any legit local/remote file names, like `local:<config>`

src/env/index.ts
```diff
@@ -0,0 +1,25 @@
+import { getQueryParam } from '#src/utils/history';
+
+/**
+ * Read more in /docs/environment.md
+ */
+
+export const IS_DEV_BUILD = import.meta.env.MODE === 'development';
```
budzworthy, line 7: I think we can get rid of `IS_DEV_BUILD` and just use `IS_DEV_OR_TEST_BUILD` everywhere, otherwise it can get confusing to tell the difference between these 3 constants

src/components/DevTools/DevTools.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Blender Avod Subs  -> avod-subs--config.json
+ *
+ * 1. We remove siteName from the config
+ * 2. We combine words with '-' looking at word breaks
+ * 3. We add '--config' part
+ *
+ */
+const getSystemConfigName = (siteName: string | undefined, configName: string | undefined) =>
```
budzworthy, line 57: It might be nice to expand the drop down component to accept values and labels so that you don't have to convert names back and forth.  Also, this would break if names don't follow the exact naming convention, which we don't have to be so rigid about.

src/App.tsx
```diff
@@ -56,12 +56,12 @@ class App extends Component {
     this.setState({ error });
     this.setState({ isLoading: false });
     clearStoredConfig();
-    logDev('Error while loading the config.json:', error);
+    logDev('Error while loading the config file:', error);
```
budzworthy, line 59: It would be nice to add a second param for the config location to the loading and error handlers to include in the log messages.

src/components/DevTools/DevTools.tsx
```diff
@@ -0,0 +1,117 @@
...
+    return '';
+  }
+
+  const name = configName
+    ?.split('--')[0]
+    .split('-')
+    .map((str) => str.charAt(0).toUpperCase() + str.slice(1))
+    .join(' ');
+
+  return siteName ? `${siteName} ${name}` : name;
```
budzworthy, line 46: Why do we add site name to the configs? It would make sense to show the site name defined in each config, but that would require loading all of them which is probably overkill.

src/components/DevTools/DevTools.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { getStoredConfig, setStoredConfig } from '#src/utils/configOverride';
+import { CONFIG_DEFAULT_SOURCE, INCLUDE_CONFIGS } from '#src/env';
+import { logDev } from '#src/utils/common';
+import { useConfigStore } from '#src/stores/ConfigStore';
+
+// Dynamic value doesn't work for some reason
+const getConfigs = () => {
+  switch (INCLUDE_CONFIGS) {
+    case 'test':
+      return import.meta.glob(`../../../configs/test/*.json`);
```
budzworthy, line 18: Does it work to add an alias for the configs directory (i.e. `#config`) in vite.config.ts to avoid all the dots?

src/utils/configOverride.ts
```diff
@@ -33,7 +28,9 @@ function getConfigOverride() {
   const url = new URL(window.location.href);
 
   if (url.searchParams.has(configFileQueryKey)) {
-    const configQuery = url.searchParams.get(configFileQueryKey)?.toLowerCase();
```
budzworthy, line 36: Also nice to keep  the lowercase here so that it's easier to manually enter something

src/utils/configOverride.ts
```diff
@@ -53,6 +50,10 @@ function getConfigOverride() {
     }
 
     // Yes this falls through to look up the stored value if the query string is invalid and that's OK
+  } else {
+    if (INCLUDE_CONFIGS !== 'dev') {
```
budzworthy, line 54: Why `!== dev`?  Shouldn't we always show it in dev?

test-e2e/tests/account_test.ts
```diff
@@ -16,7 +16,7 @@ const lastName = 'Tester';
 Feature('account').retry(3);
 
 Before(({ I }) => {
-  I.useConfig('test--subscription');
+  I.useConfig('subscription--config');
```
budzworthy, line 19: If you're renaming all of the files, might as well make them as succinct as possible by dropping the `--config` along with the `test--`

src/components/DevTools/DevTools.tsx
```diff
@@ -0,0 +1,117 @@
+import { useMemo, useState } from 'react';
```
budzworthy, line 1: It probably makes sense to combine this file with the logic in overrideConfig, since they overlap almost entirely in function and purpose

budzworthy (changes requested):
Sorry, I know there are a lot of comments, but in my defense, there are a lot of changes.

I propose we use the [modes mechanism](https://vitejs.dev/guide/env-and-mode.html#modes) of vite to specify the following different modes/environments with all of the settings defined in .env files:

1. `production` - Deployed to a real, customer site, shouldn't include any additional configs, locked down, config selector not shown
2. `demo` - the firebase demo. Connects to prod api and uses live configs from the api, not hard coded in the app. Perhaps we have a dictionary of ID's and descriptions defined in an env variable or a separate config file to populate the demo selector
3. `development` - the current dev setup for running locally (not called local because vite makes assumptions about .local files.) Connects to prod api by default and includes all configs (/dev and /test-e2e) for ease of testing. Config selector always shown.
4. `internal` - defaults to use the dev api (config files can all point to dev api data.) Not sure we need this tbh, but nice to distinguish internal from development.
5. `test` - used for unit/snapshot testing, does funny things with the css for snapshots to work. Doesn't need configs
6. `test-e2e` - only include the test configs, talk to the prod api, hide the selector so it's more representative of production

As suggested in the details above, I would only have test config files for the e2e tests that can also be used in dev.  Having too many example files is going to be confusing, hard to maintain, and give customers the impression not to use the nice dashboard system we just built. Connecting the demo to the dashboard is also going to make it easier for product people to maintain it and less likely there's conflict between test data and demo data.

Other than that, I would like to see all of the demo / selector code grouped together and isolated from the rest of the application as much as possible so it doesn't complicate the real business logic.

AntonLantukh:
Thanks Danny. I will have a look a bit later when having free time during our spotlight project.

src/components/Sidebar/Sidebar.module.scss
```diff
@@ -11,11 +11,16 @@
...
+  display: inline-block;
   width: 100%;
   height: 100%;
   background: rgba(variables.$black, 0.4);
+  visibility: hidden;
   transition: all 0.3s ease;
+
+  &.visible {
+    visibility: visible;
+  }
```
ChristiaanScheermeijer, line 23: Does this still work well for keyboard tab navigation? It should not be possible to reach the menu items using the tab (shift+tab) key when the sidebar is not open.

PR #126, opened 2022-07-22
Feat / add live channels layout and EPG

ChristiaanScheermeijer:
This PR implements the following:

- Playlist router mechanics that uses the `contentType` to switch between layouts (MVP implementation for now)
- LiveChannels container with Video component and EPG viewer
- EPG hook
- Add planby module
- Planby hook (for data transformation)
- EPG components and styling according to designs


ChristiaanScheermeijer:
No deploy preview 😢

AntonLantukh:
> No deploy preview 😢

Seems to be an action limitation:

```
on:
  pull_request:
    branches: ['develop']

```

Maybe we can just remove the branch limitation and leave:

```
on: [pull_request]

```

@dbudzins what do you think?

PS codeql has same restriction.

src/screens/Playlist/Playlist.tsx
```diff
@@ -1,21 +1,12 @@
...
-import { useAccountStore } from '../../stores/AccountStore';
-import { useConfigStore } from '../../stores/ConfigStore';
+import usePlaylist from '#src/hooks/usePlaylist';
+import ErrorPage from '#src/components/ErrorPage/ErrorPage';
+import LoadingOverlay from '#src/components/LoadingOverlay/LoadingOverlay';
 
-import styles from './Playlist.module.scss';
-
-import type { PlaylistItem } from '#types/playlist';
+const GridLayout = React.lazy(() => import('#src/containers/PlaylistGrid/PlaylistGrid'));
```
AntonLantukh, line 8: 👍

public/config.json
```diff
@@ -40,6 +40,11 @@
       "contentId": "xQaFzykq",
       "type": "playlist",
       "filterTags": "Beginner,Advanced"
+    },
+    {
+      "label": "Live",
+      "contentId": "WppSIFKF",
+      "type": "playlist"
```
AntonLantukh, line 52: I remember we discussed that it could be useful to define it as a 'live' type. @dbudzins can we support it at the back-end side? Right now we have `playlist`, `favorites` and `continue_watching` types.

ChristiaanScheermeijer: If that's the case, we don't need the playlist routing mechanism but use a dedicated URL instead. Although, we also decided to follow the upcoming layout mechanics using the `contentType` property.

ChristiaanScheermeijer:
@AntonLantukh perhaps we can use [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch), but I'm not sure what the deploy function does with the URL

budzworthy:
> > No deploy preview cry
> 
> Seems to be an action limitation:
> 
> ```
> on:
>   pull_request:
>     branches: ['develop']
> ```
> 
> Maybe we can just remove the branch limitation and leave:
> 
> ```
> on: [pull_request]
> ```
> 
> @dbudzins what do you think?
> 
> PS codeql has same restriction.

Maybe we can turn it on for `feat/*` and/or `feature/*` branches?  Doing it for everything might create unwanted noise.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

AntonLantukh:
> Maybe we can turn it on for `feat/*` and/or `feature/*` branches? Doing it for everything might create unwanted noise. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Great idea. @ChristiaanScheermeijer could you please change the workflow file (firebase-preview.yml)?

```
on:
  pull_request:
    branches: ['develop', 'feat/*', 'feature/*']
```

public/config.json
AntonLantukh: The thing is that we may want to add new live playlists on our app config page. In this case it won't be possible to do it without additional configuration (custom param) on the playlist page. And it may be also confusing when in some cases we set the type in the config, and in other cases as a custom param

AntonLantukh:
@ChristiaanScheermeijer I noticed that when first opening the ott-app and navigating to the Live section, there are no programs available. You have to visit the page one more time for it to work.

src/containers/PlaylistGrid/PlaylistGrid.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { cardUrl } from '#src/utils/formatting';
+import { filterPlaylist, getFiltersFromConfig } from '#src/utils/collection';
+import CardGrid from '#src/components/CardGrid/CardGrid';
+import Filter from '#src/components/Filter/Filter';
+import useBlurImageUpdater from '#src/hooks/useBlurImageUpdater';
+import { useAccountStore } from '#src/stores/AccountStore';
+import { useConfigStore } from '#src/stores/ConfigStore';
+import type { Playlist, PlaylistItem } from '#types/playlist';
+
+function PlaylistGrid({ playlist: { feedid: id, title, playlist } }: { playlist: Playlist }) {
```
AntonLantukh, line 17: Maybe we should better place it in `components` folders if no children are used?

src/hooks/useLiveChannels.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import type { PlaylistItem } from '#types/playlist';
+import epgService, { EpgChannel, EpgProgram } from '#src/services/epg.service';
+
+const isLiveProgram = (program: EpgProgram) => isBefore(new Date(program.startTime), new Date()) && isAfter(new Date(program.endTime), new Date());
+
+const useLiveChannels = (playlist: PlaylistItem[]) => {
+  const { data: channels = [] } = useQuery(['live-channels', ...playlist.map(({ mediaid }) => mediaid)], async () => {
+    return await epgService.getSchedules(playlist);
+  });
```
AntonLantukh:
```suggestion
  const { data: channels = [] } = useQuery(['live-channels', ...playlist.map(({ mediaid }) => mediaid)], async () => epgService.getSchedules(playlist));
```

ChristiaanScheermeijer:
> @ChristiaanScheermeijer I noticed that when first opening the ott-app and navigating to the Live section, there are no programs available. You have to visit the page one more time for it to work.

Yes, I've also seen that issue (see the known issues above). I don't know yet why this happens. The data is loaded and the state is updated correctly.

src/containers/PlaylistGrid/PlaylistGrid.tsx
ChristiaanScheermeijer: It is using multiple stores, so IMO this shouldn't be a component.

public/config.json
ChristiaanScheermeijer: @marcovandeveen @dbudzins what do you think?

budzworthy: We also need to add custom params to the playlist for the epg source, right? My way of thinking about it is that a 'live' playlist can only be shown on a PlaylistLiveChannel page, so as part of setting up the playlist, we should make sure that it has the right params in custom params, then there's nothing to adjust in the config itself. It probably ties into the meta data management project that the OTT team is looking to tackle later this year.

ChristiaanScheermeijer:
> @ChristiaanScheermeijer I noticed that when first opening the ott-app and navigating to the Live section, there are no programs available. You have to visit the page one more time for it to work.

@AntonLantukh this is fixed now 🎉  Although, we're not yet 100% sure why it does work now but it worked after moving the EPG to a dedicated component.

src/components/Epg/ChannelItem.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const { position, logo } = channel;
+
+  return (
+    <ChannelBox {...position} onClick={() => onClick && onClick(channel)} style={{ cursor: 'pointer' }}>
+      <ChannelLogo src={logo} alt="Logo" />
+    </ChannelBox>
+  );
+};
+
+export default ChannelItem;
```
budzworthy: This file appears to be unused?

src/components/EpgProgramItem/EpgProgramItem.module.scss
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  &:not(.disabled):hover {
+    outline: 2px solid theme.$primary-color;
+    transform: scale(1.01);
+  }
+
+  @include responsive.mobile-only {
+    padding: variables.$base-spacing / 2;
```
budzworthy:
```suggestion
    padding: calc(variables.$base-spacing / 2);
```

src/components/Epg/Epg.tsx
```diff
@@ -0,0 +1,90 @@
...
+  channel: EpgChannel | undefined;
+  program: EpgProgram | undefined;
+  config: Config;
+};
+
+export default function Epg({ channels, setActiveChannel, channel, program, config }: Props) {
+  const breakpoint = useBreakpoint();
+  const { t } = useTranslation('common');
+
+  const isSmall = breakpoint < Breakpoint.sm;
```
budzworthy:
```suggestion
  const isSmall = breakpoint <= Breakpoint.sm;
```
Otherwise, rename the const to `isExtraSmall`.  We've now used this same pattern in several places, would it make sense to add it to the hook?

budzworthy (approved):

src/components/Epg/Epg.tsx
ChristiaanScheermeijer: Yes, I agree. The could look something like:

```ts
const { breakpoint, isMobile, isTablet, isDesktop } = useBreakpoint();

if (isTablet) {

// or
const { breakpoint, size } = useBreakpoint();

if (size === 'tablet')
```

src/components/Epg/ChannelItem.tsx
ChristiaanScheermeijer: Hmm yes, will remove it!

src/components/Epg/Epg.tsx
ChristiaanScheermeijer: Renamed to `isMobile` for now.

PR #129, opened 2022-07-25
Feat / live channels start from the beginning

ChristiaanScheermeijer:
This PR implements the "Start from beginning" button. This button is enabled when the current program is live. When the program is finished, the "Start watching" button selects the program of the live stream and plays this item VOD.

### DIFF
--- a/src/containers/Cinema/Cinema.tsx
+++ b/src/containers/Cinema/Cinema.tsx
@@ -29,7 +29,7 @@ type Props = {
   secondaryMetadata?: React.ReactNode;
   liveStartDateTime?: string | null;
   liveEndDateTime?: string | null;
-  liveCatchup?: boolean;
+  liveFromBeginning?: boolean;
 };
 
 const Cinema: React.FC<Props> = ({
@@ -45,7 +45,7 @@ const Cinema: React.FC<Props> = ({
   feedId,
   liveStartDateTime,
   liveEndDateTime,
-  liveCatchup,
+  liveFromBeginning,
 }: Props) => {
   const { t } = useTranslation();
   const { player, features } = useConfigStore((s) => s.config);
@@ -67,7 +67,8 @@ const Cinema: React.FC<Props> = ({
       return videoProgress * item.duration;
     }
 
-    return -1;
+    // start at the beginning of the video (only for VOD content)
+    return 0;
   }, [item.duration, watchHistoryItem?.progress]);
 
   const getProgress = useCallback((): number | null => {
--- a/src/containers/Cinema/Cinema.tsx
+++ b/src/containers/Cinema/Cinema.tsx
@@ -71,7 +71,9 @@ const Cinema: React.FC<Props> = ({
   }, [item.duration, watchHistoryItem?.progress]);
 
   const getProgress = useCallback((): number | null => {
-    if (!playerInstance) return null;
+    if (!playerInstance) {
+      return null;
+    }
 
     return playerInstance.getPosition() / item.duration;
   }, [playerInstance, item.duration]);
@@ -84,8 +86,12 @@ const Cinema: React.FC<Props> = ({
   }, []);
 
   const handleFirstFrame = useCallback(() => {
-    if (liveCatchup) playerInstance?.seek(0);
-  }, [liveCatchup, playerInstance]);
+    // when playing a livestream, the first moment we can seek to the beginning of the DVR range is after the
+    // firstFrame event.
+    if (liveFromBeginning) {
+      playerInstance?.seek(0);
+    }
+  }, [liveFromBeginning, playerInstance]);
 
   const handlePlay = useCallback(() => {
     setIsPlaying(true);
--- a/src/containers/Cinema/Cinema.tsx
+++ b/src/containers/Cinema/Cinema.tsx
@@ -27,9 +27,26 @@ type Props = {
   title: string;
   primaryMetadata: React.ReactNode;
   secondaryMetadata?: React.ReactNode;
+  liveStartDateTime?: string | null;
+  liveEndDateTime?: string | null;
+  liveCatchup?: boolean;
 };
 
-const Cinema: React.FC<Props> = ({ open, item, title, primaryMetadata, secondaryMetadata, onPlay, onPause, onComplete, onClose, feedId }: Props) => {
+const Cinema: React.FC<Props> = ({
+  open,
+  item,
+  title,
+  primaryMetadata,
+  secondaryMetadata,
+  onPlay,
+  onPause,
+  onComplete,
+  onClose,
+  feedId,
+  liveStartDateTime,
+  liveEndDateTime,
+  liveCatchup,
+}: Props) => {
   const { t } = useTranslation();
   const { player, features } = useConfigStore((s) => s.config);
   const continueWatchingList = features?.continueWatchingList;
@@ -50,7 +67,7 @@ const Cinema: React.FC<Props> = ({ open, item, title, primaryMetadata, secondary
       return videoProgress * item.duration;
     }
 
-    return 0;
+    return -1;
   }, [item.duration, watchHistoryItem?.progress]);
 
   const getProgress = useCallback((): number | null => {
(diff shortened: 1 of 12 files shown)


src/containers/Cinema/Cinema.tsx
```diff
@@ -66,6 +83,10 @@ const Cinema: React.FC<Props> = ({ open, item, title, primaryMetadata, secondary
     setPlayerInstance(player);
   }, []);
 
+  const handleFirstFrame = useCallback(() => {
+    if (liveCatchup) playerInstance?.seek(0);
```
budzworthy: Let's avoid brace-less `if` statements: https://www.imperialviolet.org/2014/02/22/applebug.html

src/containers/Cinema/Cinema.tsx
```diff
@@ -50,7 +67,7 @@ const Cinema: React.FC<Props> = ({ open, item, title, primaryMetadata, secondary
       return videoProgress * item.duration;
     }
 
-    return 0;
+    return -1;
```
budzworthy: What exactly was the reason for this change? I expected to see a `>= 0` check somewhere downstream to explain this, but can't seem to find it.  Can you add a comment?

src/containers/PlaylistLiveChannels/PlaylistLiveChannels.tsx
```diff
@@ -38,10 +42,13 @@ function PlaylistLiveChannels({ playlist: { feedid, title, playlist } }: { playl
   const history = useHistory();
   const searchParams = useMemo(() => new URLSearchParams(location.search), [location.search]);
   const play = searchParams.get('play') === '1';
+  const liveStartDateTime = searchParams.get('start');
+  const liveEndDateTime = searchParams.get('end');
+  const liveCatchup = searchParams.get('catchup') === '1';
```
budzworthy: I don't know if I understand the term 'catchup' in this context. Is this playing from the beginning? Would `beginning=1` be a more intuitive query param / variable name?

src/hooks/useLiveChannels.ts
```diff
@@ -1,11 +1,15 @@
...
-import { isAfter, isBefore } from 'date-fns';
-import { useCallback, useEffect, useState } from 'react';
+import { isAfter, isBefore, subHours } from 'date-fns';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import type { PlaylistItem } from '#types/playlist';
 import epgService, { EpgChannel, EpgProgram } from '#src/services/epg.service';
 
 const isLiveProgram = (program: EpgProgram) => isBefore(new Date(program.startTime), new Date()) && isAfter(new Date(program.endTime), new Date());
+const isVodProgram = (program: EpgProgram) =>
```
budzworthy: Could you add a comment here explaining the logic for the future?

src/hooks/useLiveChannels.ts
```diff
@@ -16,6 +20,9 @@ const useLiveChannels = (playlist: PlaylistItem[]) => {
   const [channel, setChannel] = useState<EpgChannel | undefined>(channels[0]);
   const [program, setProgram] = useState<EpgProgram | undefined>();
 
+  const isLive = useMemo(() => program && isLiveProgram(program), [program]);
+  const isVod = useMemo(() => program && isVodProgram(program), [program]);
```
budzworthy: Are these memo's going to work properly as time changes (since the current time is not a dependency?)

src/hooks/useLiveChannels.ts
```diff
@@ -1,11 +1,15 @@
 import { useQuery } from 'react-query';
-import { isAfter, isBefore } from 'date-fns';
-import { useCallback, useEffect, useState } from 'react';
+import { isAfter, isBefore, subHours } from 'date-fns';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import type { PlaylistItem } from '#types/playlist';
 import epgService, { EpgChannel, EpgProgram } from '#src/services/epg.service';
 
 const isLiveProgram = (program: EpgProgram) => isBefore(new Date(program.startTime), new Date()) && isAfter(new Date(program.endTime), new Date());
```
budzworthy: Might be more readable to create now, startTime, and endTime const's to use in each of these comparisons.

src/hooks/useLiveChannels.ts
```diff
@@ -16,6 +20,9 @@ const useLiveChannels = (playlist: PlaylistItem[]) => {
   const [channel, setChannel] = useState<EpgChannel | undefined>(channels[0]);
   const [program, setProgram] = useState<EpgProgram | undefined>();
 
+  const isLive = useMemo(() => program && isLiveProgram(program), [program]);
```
budzworthy: Can we put the isVodProgram and isLiveProgram functions inline in these memo calls? It's just adding complexity having separate functions.

src/containers/Cinema/Cinema.tsx
```diff
@@ -66,6 +83,10 @@ const Cinema: React.FC<Props> = ({ open, item, title, primaryMetadata, secondary
     setPlayerInstance(player);
   }, []);
 
+  const handleFirstFrame = useCallback(() => {
```
budzworthy, line 89: I assume you've used this strategy elsewhere in symphony for example?  Can you add a comment as to why seeking in the first frame event is used?

budzworthy (commented):
Mostly clarification questions / requests and a couple small cleanups

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: Sure thing!

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: Ah, this change can be reversed. I've tried some things to make a livestream start at the start of the DVR using the `starttime` property, but this didn't work.

src/containers/PlaylistLiveChannels/PlaylistLiveChannels.tsx
ChristiaanScheermeijer: Yes, I agree!

src/hooks/useLiveChannels.ts
ChristiaanScheermeijer: Good find, I haven't worked that part out yet. I will try some things out

src/containers/Cinema/Cinema.tsx
ChristiaanScheermeijer: We are not using this elsewhere but I found that seeking only works after this event for live streams. Perhaps we can ask the Player team if this is the best approach.

src/hooks/usePlaylistItemCallback.ts
```diff
@@ -5,23 +5,45 @@ import { useConfigStore } from '../stores/ConfigStore';
...
+export const usePlaylistItemCallback = (startDateTime?: string | null, endDateTime?: string | null) => {
   const auth = useAccountStore(({ auth }) => auth);
   const signingConfig = useConfigStore((state) => state.config?.contentSigningService);
 
+  const applyLiveStreamOffset = (item: PlaylistItem) => {
+    if (!startDateTime) return item;
+
+    // The timeParam can either be just a start date like `2022-08-08T20:00:00` (to extend DVR) or a range like
+    // `2022-08-08T20:00:00-2022-08-08T22:00:00` to select a VOD from the live stream.
+    const timeParam = [startDateTime, endDateTime].filter(Boolean).join('-');
```
ChristiaanScheermeijer, line 19: @AntonLantukh @dbudzins busted... 😅

src/hooks/useLiveChannels.ts
ChristiaanScheermeijer: Hmm, those utilities are (and will be) used more often. The `isLive` and `isVod` variables return the state for the current program.

budzworthy: Oh, I didn't see them used anywhere else. If that's the case then forget I said anything.

src/hooks/useLiveChannels.ts
ChristiaanScheermeijer: I've separated the channels and program logic into two hooks. The `useLiveProgram` hook handles the state and also updates periodically.

ChristiaanScheermeijer:
@dbudzins @AntonLantukh updated and added some code comments. Let me know what you think!

PR #132, opened 2022-07-28
Feat / allow deep linking to live channel

ChristiaanScheermeijer:
This PR allows users to deep link to a live channel from any page or shared URLs.

For cards to work on the homepage, the following custom properties must be set:

```
{
  "contentType": "Live",
  "liveChannelsId": "123456" // playlist id with all live channels
}
```

### DIFF
--- a/src/containers/PlaylistLiveChannels/PlaylistLiveChannels.tsx
+++ b/src/containers/PlaylistLiveChannels/PlaylistLiveChannels.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { Helmet } from 'react-helmet';
 import shallow from 'zustand/shallow';
 import { Epg, Layout } from 'planby';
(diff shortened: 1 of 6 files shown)


src/containers/PlaylistLiveChannels/PlaylistLiveChannels.tsx
```diff
@@ -149,8 +156,7 @@ function PlaylistLiveChannels({ playlist: { feedid, title, playlist } }: { playl
             <>
               <StartWatchingButton
                 item={channelMediaItem}
-                playUrl={addQueryParams(`/p/${feedid}`, {
-                  play: 1,
+                playUrl={addQueryParams(liveChannelsURL(feedid || '', channelId, true), {
```
RCVZ, line 159: Maybe extract this to a factory function

RCVZ (approved):
Minor this about the factory function, shouldn't block this pr.

src/containers/PlaylistLiveChannels/PlaylistLiveChannels.tsx
ChristiaanScheermeijer: I had that initially, but since it has multiple optional params: `start`, `end` and `beginning` the util became too long for my taste:

```tsx
                playUrl={liveChannelsURL(feedid || '', channelId, true, isVod ? program?.startTime : undefined, isVod ? program?.endTime : undefined)}
```

And for the "Start from beginning button":

```tsx
                onClick={() => history.push(liveChannelsURL(feedid || '', channelId, true, program?.startTime : undefined, undefined, true))}
```

PR #138, opened 2022-08-03
Fix/epg fixes and improvements

ChristiaanScheermeijer:
This PR fixes some issues found while testing. There are also some UI improvements included which especially improve the EPG UI on mobile devices.

- Fix demo data having wrong dates depending on the timezone and startTime 
  - add the timezone-mock dependency to write tests for this problem
- Fix the last hour not being visible in the EPG
- Fix program outline missing border radius in Safari
- Make EPG UI compacter and fullwidth on mobile devices
- Automatically scroll to top when clicking on a program
- Rename "Start from beginning" button to "Watch from begin"
- Improve SEO and sharing metadata for live channels

### DIFF
--- a/src/hooks/usePlanByEpg.ts
+++ b/src/hooks/usePlanByEpg.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 import { useEpg } from 'planby';
-import { endOfDay, startOfDay } from 'date-fns';
+import { startOfToday, startOfTomorrow } from 'date-fns';
 
 import type { EpgChannel } from '#src/services/epg.service';
 import { is12HourClock } from '#src/utils/datetime';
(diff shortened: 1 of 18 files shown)


src/hooks/usePlanByEpg.ts
```diff
@@ -32,8 +32,11 @@ const usePlanByEpg = (channels: EpgChannel[], sidebarWidth: number, itemHeight:
 
   // this mechanism updates the EPG component range when leaving the page open for a longer period
   // the useEpg hook doesn't accept a formatted date and re-renders when not memoize the start and end dates
-  const date = startOfDay(new Date()).toJSON();
-  const [startDate, endDate] = useMemo(() => [startOfDay(new Date()), endOfDay(new Date())], [date]);
+  // @todo ideally we want to render the schedule X hours before and after the current time, but this doesn't work (yet)
```
budzworthy, line 35: From the meeting today, it sounds like it might work or might work better at least to try the +/- approach from now, but rounded to the nearest whole hour. Karol said that fractional hours don't work well.

budzworthy (approved):

src/hooks/usePlanByEpg.ts
ChristiaanScheermeijer: Ah nice, I will try that immediately

ChristiaanScheermeijer: Here are my findings so far:

1) It does work (the "now" button and line work properly)
2) The timeline does render the wrong times, but this can be fixed in our custom component
3) Increasing the +/- hours from 12 to 24 seem to result in the same problem
4) When the `startDate` lands on the previous day it also results in the same problem. This is when the current time is 1:00 AM and the start time will become 1:00 PM the previous day.

My conclusion, for now, is to keep using the `startOfDay` and `startOfTomorrow` functions until we get this sorted out in the planby component.

budzworthy: Agreed. Thanks for testing it!

PR #154, opened 2022-09-20
Feat / update react-router to 6.4.0

### LINKED ISSUE #39: OTT Web App: Upgrade to React Router 6
See hackathon branch

https://reactrouter.com/docs/en/v6/upgrading/v5

RCVZ:
## Description

This PR adds:

- react-router v6.4.0 
- refactor the router to make it possible to use the [newly](https://reactrouter.com/en/main/start/overview) added functionality of react-router. 

Fixes #39


budzworthy:
A quick scan through the changes and they all look fine, but any idea why some of the tests are failing?

ChristiaanScheermeijer:
> A quick scan through the changes and they all look fine, but any idea why some of the tests are failing?

@dbudzins @RCVZ I've pushed some changes that should fix the e2e tests 🤞

1) Optional params don't work anymore in React Router v6 (`/q/` -> 404, `/q/term` -> SearchPage)
2) A debounce ref was using a stale `navigate` reference which resulted in only the path being updated but not triggering a re-render.

ChristiaanScheermeijer:
Another fix pushed for the `/u/:page/*` route which resulted in a duplicate `page` param

budzworthy (approved):

src/components/Router/Router.tsx
```diff
@@ -1,10 +1,55 @@
...
+import User from '#src/screens/User/User';
+
+type Props = {
+  error?: Error | null;
+};
+
+export default function Router({ error }: Props) {
+  const { t } = useTranslation();
+
+  const router = useMemo(() => {
```
ChristiaanScheermeijer: Maybe add a `@todo` that explains why the routes are currently defined in a memo. Ideally, they would be defined outside the component definition. 

We should also consider moving the Router component to the containers folder or placing it directly in the src directory.

src/screens/Series/Series.tsx
```diff
@@ -33,19 +32,17 @@ import FavoriteButton from '#src/containers/FavoriteButton/FavoriteButton';
...
   const [playTrailer, setPlayTrailer] = useState<boolean>(false);
 
   // Routing
-  const history = useHistory();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const params = useParams();
   const searchParams = useMemo(() => new URLSearchParams(location.search), [location.search]);
-  const id = match?.params.id;
+  const id = params.id as string;
```
ChristiaanScheermeijer: Prevent type casting, because `params.id` can be undefined.

```suggestion
  const id = params.id || '';
```

src/screens/Movie/Movie.tsx
```diff
@@ -29,20 +28,19 @@ import FavoriteButton from '#src/containers/FavoriteButton/FavoriteButton';
...
 
   // Routing
-  const history = useHistory();
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  const params = useParams();
   const searchParams = useMemo(() => new URLSearchParams(location.search), [location.search]);
-  const id = match?.params.id;
+  const id = params.id as string;
```
ChristiaanScheermeijer:
```suggestion
  const id = params.id || '';
```

src/screens/Playlist/Playlist.tsx
```diff
@@ -8,15 +8,9 @@ import LoadingOverlay from '#src/components/LoadingOverlay/LoadingOverlay';
...
-};
-
-function Playlist({
-  match: {
-    params: { id },
-  },
-}: RouteComponentProps<PlaylistRouteParams>) {
+function Playlist() {
+  const params = useParams();
+  const id = params.id as string;
```
ChristiaanScheermeijer:
```suggestion
  const id = params.id || '';
```

src/components/Router/Router.tsx
RCVZ: I have removed the useMemo, it was unnecessary because the router only renders when there is an error which makes the useMemo useless :P

ChristiaanScheermeijer (approved):

ChristiaanScheermeijer:
@dbudzins can we merge this one?

budzworthy:
> @dbudzins can we merge this one?

Yes go for it!

PR #155, opened 2022-09-20
feat(project): support alternate images

### LINKED ISSUE #114: JW alternate images support
Add support for JW Player alternate media images.

## Scope
- Use configured image label for 
  - Homepage Shelves
  - Library 
  - Live Channel 
  - Series
  - Movie
- Fallback to poster if image is not available

## Out-of-scope 
- Shelf aspect ratio support (e.g. vertical images) --> Ticket #152

## Customer workflow 
- Default is the poster image, which is always there and auto generated
- Set custom image labels on config & media level: 
   - `backgroundImage`
   - `channelLogoImage`
- Set custom shelf labels on config & playlist level 
  -  `shelfImage`
  

### Homepage shelves
image

### Library page
image

### Movie page
image

### Series Page
image

### Live Page
image

ChristiaanScheermeijer:
## Description

This PR implements alternate image support for shelf, background and channel logo images. To make it more forgiving when adding/changing alternate images, a fallback mechanism has been made that loads the media poster image when the alternate image fails.

A generic `<Image />` component is introduced which accepts a `src` and `srcFallback` prop. When the `src` fails it tries the `srcFallback` instead.

Feedback is welcome!

Fixes #114


src/components/Image/Image.tsx
```diff
@@ -0,0 +1,37 @@
...
+  className?: string;
+  src?: string;
+  fallbackSrc?: string;
+  onLoad?: () => void;
+  alt?: string;
+};
+
+const Image = ({ className, src, fallbackSrc, onLoad, alt = '' }: Props) => {
+  const [imgSrc, setImgSrc] = useState(src);
+  const imageRef = useRef<HTMLImageElement>(null);
```
ChristiaanScheermeijer: Remove redundant ref

src/screens/Movie/Movie.tsx
```diff
@@ -157,7 +163,7 @@ const Movie = ({ match, location }: RouteComponentProps<MovieRouteParams>): JSX.
         {playlist ? (
           <>
             <div className={styles.related}>
-              <h3>{playlist.title}</h3>
+              <h3>{isPlaylistPlaceholderData || playlistLoading ? 'Loading' : playlist.title}</h3>
```
ChristiaanScheermeijer: This change prevents a layout shift while loading the playlist data.  `Loading` should be translated

src/components/Favorites/Favorites.tsx
```diff
@@ -32,9 +33,19 @@ const cols: Breakpoints = {
...
+  shelfTitles,
+  accessModel,
+  hasSubscription,
+  onCardClick,
+  onCardHover,
+  onClearFavoritesClick,
+  getCardImages,
+}: Props): JSX.Element => {
   const { t } = useTranslation('user');
-  const shelfTitles = useConfigStore((s) => s.config.styling.shelfTitles);
```
ChristiaanScheermeijer, line 37: `<Favorites />` is a component, so it shouldn't be using a store

ChristiaanScheermeijer:
@dbudzins @AntonLantukh are the failed e2e tests in the latest run flaky?

vite.config.ts
```diff
@@ -65,6 +65,7 @@ export default ({ mode }: { mode: 'production' | 'development' | 'test' }) => {
       globals: true,
       environment: 'jsdom',
       setupFiles: ['test/vitest.setup.ts'],
+      css: true,
```
budzworthy, line 61: 2 questions:

1. Should we remove the `mode === 'test'` in the css section of the main config above?
2. Can we use `stable` for `classNameStrategy` so there's less noise in the (future) noise in the snapshots? (https://vitest.dev/config/#css-modules-classnamestrategy)

ChristiaanScheermeijer: 1. Good question, I will check!
2. It defaults to `stable`, it shouldn't create any noise in future PRs

ChristiaanScheermeijer: The CSS override is redundant indeed, I've removed it!

budzworthy:
> @dbudzins @AntonLantukh are the failed e2e tests in the latest run flaky?

They haven't been flaky up to this point, but they depend on cleeng, so if cleeng had a hiccup they could fail... Looks like the latest commit is back to normal.

ChristiaanScheermeijer:
Rebased PR and fixed `$ yarn test-coverage` script by adding the `@vitest/coverage-c8` dependency

src/components/CardGrid/CardGrid.tsx
```diff
@@ -38,6 +39,7 @@ function CardGrid({
   playlist,
   onCardClick,
   onCardHover,
+  getCardImages,
```
budzworthy: It seems like this always traces back to `getShelfItemImages` in `src/stores/ConfigController.ts`.  Does it need to be a prop?

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,36 @@
...
+    if (onLoad) onLoad();
+  };
+
+  const handleError = () => {
+    if (fallbackSrc && fallbackSrc !== src) {
+      setImgSrc(fallbackSrc);
+    }
+  };
+
+  useEffect(() => {
```
RCVZ, line 34: I think this should be a useLayoutEffect because it runs before the browser has a chance to paint the screen.

RCVZ (approved):

src/stores/ConfigController.ts
```diff
@@ -0,0 +1,13 @@
+import { getMediaImages, ImageProperty } from '#src/services/image.service';
+import { useConfigStore } from '#src/stores/ConfigStore';
+import type { Playlist, PlaylistItem } from '#types/playlist';
+
+const getItemImages = (type: ImageProperty, item: PlaylistItem, playlist?: Playlist, width?: number) => {
```
budzworthy: This might be a little dangerous, because it's got a hook hidden inside of it and isn't named `use<Something>`.  Maybe it's better to make config an input variable? Or just rename all of these to `use<Type>Images` and then you can put it into a `useImages` hook.

src/stores/ConfigController.ts
```diff
@@ -0,0 +1,13 @@
...
+  const config = useConfigStore.getState().config;
+
+  return getMediaImages(config, type, item, playlist, width);
+};
+
+export const getShelfItemImages = (item: PlaylistItem, playlist: Playlist, width: number) => getItemImages('shelfImage', item, playlist, width);
```
budzworthy: I'm not sure of putting this all into a config controller class. Seems like it might be more appropriate to put it into a `utils/images.ts` or something

src/services/image.service.ts
```diff
@@ -0,0 +1,34 @@
+import type { Playlist, PlaylistItem } from '#types/playlist';
```
budzworthy: Is this really a service?  It doesn't seem to access any external resources, which the other services do, though I suppose it's coupled to the JWP alternate images API convention...

src/services/image.service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { findPlaylistImageForWidth } from '#src/utils/collection';
+import type { Config } from '#types/Config';
+
+export type ImageProperty = 'shelfImage' | 'backgroundImage' | 'channelLogoImage';
+
+export const generateAlternateImageURL = (mediaid: string, imageLabel: string, width: number) =>
+  `https://img.jwplayer.com/v1/media/${mediaid}/images/${imageLabel}.jpg?width=${width}`;
+
+export const getAlternateImageLabel = (config: Config, propertyName: ImageProperty, item: PlaylistItem, playlist?: Playlist) => {
+  const alternateImageLabel = propertyName === 'shelfImage' ? playlist?.[propertyName] : item[propertyName];
```
budzworthy: Nice to have a small comment to explain the difference here
```suggestion
  // Shelf images are set at the playlist level, all others are at the media item level
  const alternateImageLabel = propertyName === 'shelfImage' ? playlist?.[propertyName] : item[propertyName];
```

src/services/image.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  const configImageLabel = config.custom?.[propertyName];
```
budzworthy: It might be safer to decouple the string used throughout the application from the property name used in the config file in case the latter ever changes or moves, so it doesn't cause quite so much noise.  Also it obviously could cause trouble if the playlist/item property name didn't equal the custom config property name. It could also get complicated if properties move to different places outside of custom config or there's an image type that doesn't have a custom config override.

src/services/image.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  // image label override from playlist or playlist item
+  if (typeof alternateImageLabel === 'string') {
+    return alternateImageLabel;
+  }
+
+  // image label override from config
+  if (typeof configImageLabel === 'string') {
+    return configImageLabel;
+  }
```
budzworthy:
```suggestion
  return alternateImageLabel || configImageLabel;
```

src/services/image.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getMediaImages = (config: Config, propertyName: ImageProperty, item: PlaylistItem, playlist?: Playlist, width = 640) => {
+  const posterImage = findPlaylistImageForWidth(item, width);
+  const alternateImageLabel = getAlternateImageLabel(config, propertyName, item, playlist);
+
+  if (alternateImageLabel) {
+    return [generateAlternateImageURL(item.mediaid, alternateImageLabel, width), posterImage];
+  }
+
+  return [posterImage];
```
budzworthy: I think it's nicer to have named properties returned, otherwise there's no real way to know if image[0] is the main or the fallback except by inspecting the code and deciphering how it's used.  Note, it doesn't hurt to have fallback hardcoded as poster, but I'd be find returning undefined for it if it's already set as the main image.

I would also suggest we use this type for `image` throughout the application so we aren't adding extra props (i.e. having both `image` and `fallbackImage`) to everything.
```suggestion
  return {
    image: alternateImageLabel ? generateAlternateImageURL(item.mediaid, alternateImageLabel, width) : posterImage,
    fallbackImage: posterImage
  };
```

src/components/Image/Image.tsx
ChristiaanScheermeijer: I don't think it is needed here since we're not relying on the src to be set directly. 

https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

src/components/CardGrid/CardGrid.tsx
ChristiaanScheermeijer: There are other ways. It is a prop because the `getShelfItemImages` method relies on a param in the config store. The general convention is to not use stores or controllers inside components directly but to pass them via containers instead.

Do you see alternatives?

src/stores/ConfigController.ts
ChristiaanScheermeijer: I see what you mean, but we're not using `useConfigStore` as a hook but get the raw state by using the [getState](https://github.com/pmndrs/zustand#readingwriting-state-and-reacting-to-changes-outside-of-components) method.

This pattern is also being used in other controllers.

The confusing part is the name which we potentially can solve by either not using named exports e.g.

**ConfigStore.ts**
```ts
export default createStore(...);
```

**ConfigController.ts**
```ts
import configStore from './ConfigStore';
```

**AnyComponent.ts**
```ts
const useConfigStore from '#src/stores/ConfigStore';
```

Or by `importing as` to make it more clear:

```ts
import { useConfigStore as configStore } from './ConfigStore';
```

ChristiaanScheermeijer: We could also wrap it up in a custom `useImages` hook indeed...

src/stores/ConfigController.ts
ChristiaanScheermeijer: The reason being is that it reads from a store and prevents proxying the config for each use

src/services/image.service.ts
ChristiaanScheermeijer: We could consider moving the logic to a custom hook or util and move the `generateAlternateImageUrl` to the API service, what do you think?

src/components/CardGrid/CardGrid.tsx
```diff
@@ -51,13 +53,14 @@ function CardGrid({
...
   const imageSourceWidth = 320 * (window.devicePixelRatio > 1 || isLargeScreen ? 2 : 1);
-  const rows = chunk<PlaylistItem>(playlist, cols[breakpoint]);
+  const rows = chunk<PlaylistItem>(playlist.playlist, cols[breakpoint]);
 
   const cellRenderer = ({ columnIndex, rowIndex, style }: GridCellProps) => {
     if (!rows[rowIndex][columnIndex]) return;
 
     const playlistItem: PlaylistItem = rows[rowIndex][columnIndex];
     const { mediaid, title, duration, seriesId, episodeNumber, seasonNumber } = playlistItem;
+    const [image, fallbackImage] = getCardImages(playlistItem, playlist, imageSourceWidth);
```
budzworthy: It's also a little bit weird how we get the images in the component instead of when the data is fetched (like you did with epg channels.)  It might be nice to calculate these once in api.service.ts and attach them to the items there (the Playlist format returned from the api should ideally be more decoupled from the type used in components, but that may be too much for the moment.)

src/components/CardGrid/CardGrid.tsx
budzworthy: Can they use hooks?  tbh, I'm not a big fan of conventions that create a bunch of boilerplate code, so given the choice between accessing a store in a component and having 8 extra properties, I would always choose the first, but I'm willing to listen to reasons why that breaks things in the long run.

src/stores/ConfigController.ts
budzworthy: Ah, it's not a hook. I would either rename the named export or export default.

src/stores/ConfigController.ts
budzworthy: What do you mean proxying the config?

src/services/image.service.ts
budzworthy: Without actually digging into it, I would say probably it makes sense in the API service, since it's JW delivery API specific.

ChristiaanScheermeijer: Done!

src/services/image.service.ts
ChristiaanScheermeijer: This has been implicitly achieved since we're now mostly using the PlaylistItem properties. The actual properties are defined in an enum and only need to be changed once a different property is used.

src/services/image.service.ts
ChristiaanScheermeijer: The `alternateImageLabel` and `configImageLabel` types are `unknown` so the explicit type check is used to satisfy typescript

src/services/image.service.ts
ChristiaanScheermeijer: This has been updated

src/components/CardGrid/CardGrid.tsx
ChristiaanScheermeijer: Done

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,43 @@
...
+import classNames from 'classnames';
+
+import styles from './Image.module.scss';
+
+import { addQueryParams } from '#src/utils/formatting';
+
+type Props = {
+  className?: string;
+  src?: string;
+  fallbackSrc?: string;
```
budzworthy: Suggest using the `ImageData` type here so you don't manually have to assign both props for each usages

src/utils/image.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const generateAlternateImageURL = (mediaid: string, imageLabel: string, width: number) =>
+  `https://img.jwplayer.com/v1/media/${mediaid}/images/${imageLabel}.jpg?width=${width}`;
+
+export const generateImageData = (config: Config, propertyName: string, item: PlaylistItem, playlist?: Playlist, width = 640) => {
+  const posterImage = findPlaylistImageForWidth(item, width);
+  // when a playlist is given, get the alternate image label from the playlist instead
+  const contentImageLabel = playlist ? playlist[propertyName] : item[propertyName];
+  const configImageLabel = config.custom?.[propertyName];
+  const imageLabel = typeof contentImageLabel === 'string' ? contentImageLabel : typeof configImageLabel === 'string' ? configImageLabel : undefined;
```
budzworthy:
```suggestion
  const contentImageLabel = (playlist ? playlist[propertyName] : item[propertyName]) as string;
  const configImageLabel = config.custom?.[propertyName] as string;
  const imageLabel = contentImageLabel || configImageLabel;
```

src/hooks/usePlanByEpg.ts
```diff
@@ -13,13 +13,21 @@ const isBaseTimeFormat = is12HourClock();
...
+        channelLogoImage: channel.channelLogoImage,
+        backgroundImage: channel.backgroundImage,
+      })),
       channels.flatMap((channel) =>
         channel.programs.map((program) => ({
           channelUuid: channel.id,
           id: program.id,
           title: program.title,
-          image: program.image || '',
+          image: program.shelfImage?.image || '',
```
budzworthy, line 27: Does it make sense to use the fallback image here?

src/screens/Movie/Movie.tsx
```diff
@@ -155,10 +155,10 @@ const Movie = (): JSX.Element => {
         {playlist ? (
           <>
             <div className={styles.related}>
-              <h3>{playlist.title}</h3>
+              <h3>{isPlaylistPlaceholderData || playlistLoading ? t('common:loading') : playlist.title}</h3>
```
budzworthy: The 'Loading' here feels a little weird to me, since it's on a header on a secondary part of the page. I think the suggestion below achieves the same thing without troubling the user with status of something extra. Ideally we would have a grey flashy empty text placeholder block for this and for the playlist item titles, but in the meantime, I think this is an easy fix.

```suggestion
              <h3>{playlist.title || '\u00A0'}</h3>
```

types/playlist.d.ts
```diff
@@ -25,6 +30,9 @@ export type PlaylistItem = {
   feedid: string;
   image: string;
   images: Image[];
+  shelfImage?: ImageData;
```
budzworthy, line 33: I would really like to decouple this so that the delivery api interface and the app's internal data structure aren't mixed together, but probably a nice to have for another day...

budzworthy (approved):
A few really small things in the comments, but otherwise looks nice!

src/hooks/usePlanByEpg.ts
ChristiaanScheermeijer: We don't use the `image` property, but it is required by planby.

src/screens/Movie/Movie.tsx
ChristiaanScheermeijer: Yes, good suggestion

types/playlist.d.ts
ChristiaanScheermeijer: I agree, but indeed too big for now. We should write this down when separating the services from clients.

src/stores/ConfigController.ts
ChristiaanScheermeijer: Done

src/components/Image/Image.tsx
ChristiaanScheermeijer: Done

src/utils/image.ts
ChristiaanScheermeijer: Done

budzworthy (approved):
Looks good to me!

PR #163, opened 2022-09-28
Feat / inline player layout

RCVZ:
## Description

This PR adds:

- chore(project): inline player test config
- feat(project): related video grid component
- feat(project): video layout component
- feat(inlineplayer): video detail inline
- feat(inlineplayer): video list item
- feat(inlineplayer): related video list
- feat(inlineplayer):  public overlay
- feat(project): small tablet breakpoint
- chore: enable dropdown always filter component
- feat(project): implement inlinelayout movie screen
- feat(project): implement inlinelayout serie screen

Fixes #159


src/components/Filter/Filter.tsx
```diff
@@ -3,20 +3,22 @@ import { useTranslation } from 'react-i18next';
...
 
+import useBreakpoint, { Breakpoint } from '#src/hooks/useBreakpoint';
+
 type Props = {
   name: string;
   value: string;
   valuePrefix?: string;
   defaultLabel: string;
   options: string[];
+  dropDown?: boolean;
```
budzworthy: Maybe it's better to call this forceDropdown or something like that, because the filter can sometimes be a dropdown still even when it's false.

src/components/VideoLayout/VideoLayout.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <div className={styles.relatedVideosList}>{relatedVideosList}</div>
+        <div className={styles.videoDetailsInline}>{videoDetailsInline}</div>
+      </div>
+    );
+  }
+
+  return (
+    <div className={styles.videoCinemaLayout}>
+      {cinemaPlayer}
+      {videoDetails}
```
budzworthy: I think with minimal effort, playlistlivechannel can also use VideoLayout (always with inline=false), then you can fold VideoDetails and VideoDetailsInline into this as well.  I appreciate component composition, but in this case I'm worried it's causing more duplication than necessary, because most things only need to be used in 1 place.

src/components/RelatedVideoGrid/RelatedVideoGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+            name="categories"
+            value={currentFilter}
+            valuePrefix={filterValuePrefix}
+            defaultLabel={defaultFilterLabel}
+            options={filters}
+            setValue={setFilter}
+          />
+        )}
+      </div>
+      <CardGrid
```
budzworthy: Having CardGrid and VideoList, it seems that RelatedVideoGrid/List are a bit of overkill, especially since they share duplicated Filter logic. I would try to move these into VideoLayout.

src/components/VideoLayout/VideoLayout.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const VideoLayout: React.FC<Props> = ({
+  inlineLayout,
+  videoDetails,
+  videoDetailsInline,
+  inlinePlayer,
+  cinemaPlayer,
+  relatedVideosList,
+  relatedVideosGrid,
```
ChristiaanScheermeijer: Does it make sense to have separate props for the variants or would it make more sense to use:

```
{
  videoDetails,
  player,
  relatedVideos,
  inline,
}
```

The only downside is that you'll have to do ternary operators where you use the VideoLayout e.g.

```
<VideoLayout
  inline={isInline}
  videoDetails={isInline ? <InlineVideoDetails ...props /> : <VideoDetails ...props />}
  ...
/>
```

@RCVZ @dbudzins

src/containers/Cinema/Cinema.test.tsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+// import { render } from '@testing-library/react';
+
```
ChristiaanScheermeijer: Remove

src/components/VideoLayout/VideoLayout.tsx
RCVZ: I thought of adding the VideoDetail and VideoDetailsInline to the video layout, but it would make the video layout less flexible. But I will change it.

src/containers/InlinePlayer/InlinePlayer.tsx
```diff
@@ -16,22 +24,59 @@ type Props = {
...
-        onComplete={onComplete}
-        liveEndDateTime={liveEndDateTime}
-        liveFromBeginning={liveFromBeginning}
-        liveStartDateTime={liveStartDateTime}
-      />
+      <Fade open={isLocked}>
+        <div className={styles.lockedOverlay}>
+          <Image className={styles.poster} image={item.shelfImage} alt={item.title} width={1280} />
+          <Lock className={styles.lock} />
+          <h2 className={styles.callToAction}>{t('video:sign_up_to_start_watching')}</h2>
```
ChristiaanScheermeijer: I think the `callToAction` name is misused here (no button)

src/components/VideoLayout/VideoLayout.tsx
budzworthy: I would suggest moving the components into VideoLayout and passing necessary props through instead.

src/containers/InlinePlayer/InlinePlayer.tsx
```diff
@@ -16,22 +24,59 @@ type Props = {
...
   liveEndDateTime?: string | null;
   liveFromBeginning?: boolean;
+  startWatchingButton: React.ReactNode;
+  isLogged: boolean;
+  isLocked: boolean;
 };
 
-const InlinePlayer: React.FC<Props> = ({ open, item, onPlay, onPause, onComplete, feedId, liveStartDateTime, liveEndDateTime, liveFromBeginning }: Props) => {
+const InlinePlayer: React.FC<Props> = ({
+  open,
```
ChristiaanScheermeijer: The `open` prop is a bit weird for the inline player. I also think it is redundant actually

src/containers/InlinePlayer/InlinePlayer.tsx
```diff
@@ -16,22 +24,59 @@ type Props = {
   liveStartDateTime?: string | null;
   liveEndDateTime?: string | null;
   liveFromBeginning?: boolean;
+  startWatchingButton: React.ReactNode;
+  isLogged: boolean;
```
ChristiaanScheermeijer: `isLogged` -> `isLoggedIn`

But you can also obtain the `isLoggedIn` state within this container instead of passing it as prop.

src/containers/InlinePlayer/InlinePlayer.tsx
```diff
@@ -16,22 +24,59 @@ type Props = {
...
+  isLocked: boolean;
```
ChristiaanScheermeijer: `isLocked` refers to showing the "paywall" instead of the player

src/containers/InlinePlayer/InlinePlayer.tsx
```diff
@@ -16,22 +24,59 @@ type Props = {
...
-        feedId={feedId}
-        onPlay={onPlay}
-        onPause={onPause}
-        onComplete={onComplete}
-        liveEndDateTime={liveEndDateTime}
-        liveFromBeginning={liveFromBeginning}
-        liveStartDateTime={liveStartDateTime}
-      />
+      <Fade open={isLocked}>
+        <div className={styles.lockedOverlay}>
```
ChristiaanScheermeijer:
```suggestion
        <div className={styles.paywall}>
```

src/containers/InlinePlayer/InlinePlayer.tsx
```diff
@@ -16,22 +24,59 @@ type Props = {
...
+          <Lock className={styles.lock} />
+          <h2 className={styles.callToAction}>{t('video:sign_up_to_start_watching')}</h2>
+          <span className={styles.text}>{t('account:choose_offer.watch_this_on_platform', { siteName })}</span>
+          {startWatchingButton}
+          {!isLogged && <Button onClick={loginButtonClickHandler} label={t('common:sign_in')} />}
+        </div>
+      </Fade>
+      {!isLocked && (
+        <PlayerContainer
+          visible={open}
```
ChristiaanScheermeijer: The `visible` prop can always be true:

```suggestion
          visible={true}
```

src/containers/PlayerContainer/PlayerContainer.tsx
```diff
@@ -38,6 +39,8 @@ const PlayerContainer: React.FC<Props> = ({
   liveFromBeginning,
   liveStartDateTime,
 }: Props) => {
+  const [searchParams] = useSearchParams();
+  const autostart = searchParams.get('play') === '1';
```
ChristiaanScheermeijer, line 44:
```suggestion
  const autostart = useQueryParam('play') === '1';
```

src/components/VideoLayout/VideoLayout.tsx
RCVZ: Yes I do think is increases legibility

src/components/VideoListItem/VideoListItem.tsx
```diff
@@ -0,0 +1,87 @@
...
+    [styles.visible]: imageLoaded,
+  });
+
+  const renderTagLabel = () => {
+    if (loading || !title) return null;
+
+    if (seriesId) {
+      return t('series');
+    } else if (seasonNumber && episodeNumber) {
+      return `S${seasonNumber}:E${episodeNumber}`;
```
budzworthy: Some of this is already moved to `formatting.ts` in the latest on develop.  It would be good to reuse / add to that instead of duplicating this from Card.tsx.

budzworthy (commented):
Can you combine this PR and the inline-video-player PR?

ChristiaanScheermeijer:
@dbudzins I think changing the base branch to develop has the same effect. It probably won't be needed to do a two-step merge since most changes are in this PR anyways.

src/screens/Movie/Movie.tsx
```diff
@@ -115,56 +134,51 @@ const Movie = (): JSX.Element => {
...
-        onComplete={handleComplete}
-        feedId={feedId ?? undefined}
-      />
-      <TrailerModal item={trailerItem} title={`${item.title} - Trailer`} open={playTrailer} onClose={() => setPlayTrailer(false)} />
-      <VideoDetails
         title={item.title}
         description={item.description}
+        inlineLayout={inlineLayout}
         primaryMetadata={primaryMetadata}
-        image={item.backgroundImage}
```
ChristiaanScheermeijer: Is `image` missing in the `VideoDetails` component? I also don't see an image anymore :-(

RCVZ: Fixed

ChristiaanScheermeijer (approved):
Looks good to me, rebased develop and squashed all commits. Will merge when everything is green

PR #168, opened 2022-10-04
Fix / bug fixes and rework

RCVZ:
## Description

This PR fixes:

- inline video player not visible on ios 14
- list item visible in menu on mobile
- inline player line under the buttonbar on mobile
- spacing issues in the related video list
- use background image instead of shelfimage for paywall inline player
- logo stretched under some conditions
- lighthouse accessibility issues


ChristiaanScheermeijer (approved):
Nice work!

PR #171, opened 2022-10-05
Feat / add static page test

RCVZ:
## Description

This PR adds:

- static page end to end test


test-e2e/utils/constants.ts
```diff
@@ -25,6 +25,7 @@ export default {
   loginUrl: `${baseUrl}?u=login`,
   paymentsUrl: `${baseUrl}u/payments`,
   offersUrl: `${baseUrl}u/payments?u=choose-offer`,
+  staticPageUrl: `${baseUrl}m/oGNEecGq`,
```
budzworthy: Can we avoid hard coding media id's into the tests?  It makes it much harder to change or update the test data (i.e. run tests on a different property)

RCVZ: I don't think there is another way, because currently there is no way to navigate to a static page without knowing the mediaid.

budzworthy: Can't we add the static page to a playlist?  Either in a carousel or a library?

PR #187, opened 2022-11-04
Dbudzinski/remove epg files

budzworthy:
## Description

The dummy epg data was being included in the production bundle, so I conditionalized it.

### Steps completed:

- [] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [x] Linting passing
- [x] Unit tests passing
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


vite.config.ts
```diff
@@ -19,6 +20,20 @@ export default ({ mode }: { mode: 'production' | 'development' | 'test' }) => {
     }),
   ];
 
+  // These files are only needed in dev / test, don't include in prod builds
+  if (['1', 'true'].includes(process.env.INCLUDE_EPG_DATA?.toLowerCase() || '')) {
```
AntonLantukh, line 24: @dbudzins Danny, do we need this additional flag? Maybe we can just use mode === 'production' not to include it. Or do we still need for preview builds?

AntonLantukh (approved):

vite.config.ts
budzworthy: We need it for preview/demo builds. I've actually refactored it a bit more to make better use of 'mode' that way, but didn't want to complicate this PR any further.

PR #188, opened 2022-11-08
Dbudzinski/no id dialog

budzworthy:
## Description

Dialog to show in 'demo' mode when config ID is missing

### Steps completed:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/components/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -10,8 +10,8 @@ type Props = {
   open: boolean;
   title: string;
   body: string;
-  onConfirm: () => void;
-  onClose: () => void;
+  onConfirm?: () => void;
```
AntonLantukh: Did you change it intentionally? `onConfirm` handler looks like a required handler for ConfirmationDialog. Probably the same for `onClose`.

src/containers/Layout/Layout.module.scss
```diff
@@ -4,7 +4,6 @@
 .layout {
   display: flex;
   flex-direction: column;
-  height: 100vh;
```
AntonLantukh, line 7: This one was used to show the footer at the bottom of the page. It makes sense when there is not much content on the page.

Screenshot 2022-11-14 at 22 37 49

src/styles/main.scss
```diff
@@ -21,7 +21,7 @@ body {
   margin: 0;
   padding: 0;
   overflow-x: hidden;
-  overflow-y: scroll;
+  overflow-y: auto;
```
AntonLantukh, line 24: Does it work fine when we have > 6-7 playlists, so that they are loaded dynamically when scrolling?

src/components/Dialog/Dialog.tsx
```diff
@@ -8,7 +8,7 @@ import styles from './Dialog.module.scss';
 
 type Props = {
   open: boolean;
-  onClose: () => void;
+  onClose?: () => void;
```
AntonLantukh: Should we probably also leave it as required?

src/utils/common.ts
```diff
@@ -49,7 +49,10 @@ export function calculateContrastColor(color: string) {
   return rgb.r * 0.299 + rgb.g * 0.587 + rgb.b * 0.114 > 186 ? '#000000' : '#FFFFFF';
 }
 
+// Build is either Dev or Production
+// Mode can be dev, jwdev, demo, test, production, etc.
 export const IS_DEV_BUILD = import.meta.env.DEV;
+export const IS_DEMO_MODE = import.meta.env.MODE === 'demo';
```
AntonLantukh, line 55: Do we need to include it also in docs?

src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [showDemoWarning, setShowWarning] = useState(false);
+
+  const clearConfig = () => {
+    clearStoredConfig();
+    window.location.reload();
+  };
+
+  if (configLocation) {
+    return (
+      <div className={styles.note}>
```
AntonLantukh, line 30: I like the way we do it for the dashboard for PR builds. The message is an eye catcher being locate on the top of the page. Can we do the same here? When I first clicking on 'Cancel', it took me a while to find the 'back' link.

src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const cancelDemoClick = () => setShowWarning(false);
+
+  return (
+    <div className={styles.configModal}>
+      <ErrorPage title={t('app_config_not_found')}>
+        <form method="GET" action="/">
+          <TextField required placeholder={t('please_enter_config_id')} name="app-config" />
+          <div className={styles.controls}>
+            <Button label={t('submit_config_id')} type={'submit'} />
+            {fallbackConfig && <Button label={t('cancel_config_id')} onClick={cancelConfigClick} />}
```
AntonLantukh, line 59: Can we use 'Use demo' or 'Show an example' title here? Cancel looks a bit strange because when you first opened the page you didn't make any actions. So there is nothing to cancel.

src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const cancelDemoClick = () => setShowWarning(false);
+
+  return (
+    <div className={styles.configModal}>
+      <ErrorPage title={t('app_config_not_found')}>
+        <form method="GET" action="/">
+          <TextField required placeholder={t('please_enter_config_id')} name="app-config" />
+          <div className={styles.controls}>
+            <Button label={t('submit_config_id')} type={'submit'} />
```
AntonLantukh, line 58: Do we need to add a handler here to use the provided config? For now it doesn't work for me.

src/components/ConfirmationDialog/ConfirmationDialog.tsx
budzworthy: I think this was an artifact from an intermediate refactor. Fixed.

src/containers/Layout/Layout.module.scss
budzworthy: Good catch. I changed this to get the demo dialog to work right. I have now pushed a fix for the footer with a little different styling.

src/styles/main.scss
budzworthy: Yes it seems to work fine. I think the new virtualization framework is more robust.

src/utils/common.ts
budzworthy: I will make a separate big update of docs.

src/components/DemoConfigDialog/DemoConfigDialog.tsx
budzworthy: I agree, but the product use case outweighs that in this case. Most often customers aren't going to be switching between lots of different configs.  And product wants the app to look as much like production as possible. The escape link is just there for the edge case.

src/components/DemoConfigDialog/DemoConfigDialog.tsx
budzworthy: Again, I agree, but product likes the submit / cancel. I think it would also be a little confusing with submit and 'show demo' because in theory submit is also showing you a demo, so it's unclear which button to click in that case. The messaging and dialog should hopefully be enough to clarify what cancel does.

src/components/ConfirmationDialog/ConfirmationDialog.tsx
budzworthy: Are you sure it was a valid config?  Otherwise, what browser are you using? There's no handler because I'm just using the html form to do everything on submit

src/components/Dialog/Dialog.tsx
budzworthy: Fixed

src/components/DemoConfigDialog/DemoConfigDialog.tsx
AntonLantukh: @dbudzins should it work now?

AntonLantukh (approved):

Issue #192, opened 2022-11-17
Create local ini file for per-instance settings


budzworthy:
Instead of compiling the default and allowed config sources, load them from a small ini file at /public/.webapp.ini.  This will allow customers to change the values without having to recompile code, but still having them 'locked' down.

budzworthy:
Done

Issue #182, opened 2022-10-19
Remove config picker when not Dev/local


budzworthy:
The config picker doesn't work well with demo and sales flows on the preview site, so let's only show it in Dev/local

budzworthy:
Done.

Issue #146, opened 2022-09-13
Evaluate the settings for 'default config' and 'allowed configs' at runtime


budzworthy:
Change these variables from compile time .env values to instead be values that can be read at runtime.  Format TBD, either environment variables or a json format.  This will support distributing compiled assets and also make it easier to follow the migration guide and easy deployment settings.

Suggestion: create a small local file app-settings.json

budzworthy:
Done

Issue #134, opened 2022-07-29
Dynamic configuration as part of easy deployment instructions


marcovandeveen:
To get started we're forwarding people to these docs: 
- Easy deployment
- Enable dynamic configuration (from JW Dashboard)

Suggestion: make 'enable dynamic configuration' part of the easy deployment steps.

budzworthy:
I wouldn't suggest to customers that they turn on the dynamic unsafe option, but maybe we can add better details on how to add the ID's of the configs they want to work with.

budzworthy:
This is complete with the ini file architecture

PR #197, opened 2022-11-28
Dbudzinski/config settings

budzworthy:
## Description

This PR touches a lot of files, but essentially it is just introducing .ini files to use for default / allowed config sources instead of getting them from compile time environment variables.

The documentation is extensively updated to describe how this works, but essentially the code loads .webapp.ini from /public. The vite build copies the appropriate ini file into the build based on the `mode`. It also copies ini file templates from a source controlled directory (`/ini/templates`) to an ignored directory (just `/ini`) so that the developer process requires less manually interaction and customer/instance specific data doesn't get stored to git.

I also cleaned up the app loading to make it clearer how the ini file and app config get loaded, show better error feedback, and use more appropriate react code like hooks and queries for navigation and data loading.

This PR solves #192, #182, #146, #134, #127, .

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


src/App.tsx
```diff
@@ -1,92 +1,50 @@
...
-      await initializeAccount();
-    }
-
-    // We only request favorites and continue_watching data if there is a corresponding item in the content section
-    // and a playlist in the features section.
-    // We first initialize the account otherwise if we have favorites saved as externalData and in a local storage the sections may blink
-    if (config.features?.continueWatchingList && config.content.some((el) => el.type === PersonalShelf.ContinueWatching)) {
-      await restoreWatchHistory();
-    }
+  useEffect(() => {
```
AntonLantukh, line 20: Maybe it is better to store i18n logic in a separate context provider?

src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -1,80 +1,178 @@
...
-import { addConfigQueryParam, clearStoredConfig, getConfigLocation } from '#src/utils/configOverride';
+import { getConfigNavigateCallback } from '#src/utils/configOverride';
 import Link from '#components/Link/Link';
 import ConfirmationDialog from '#components/ConfirmationDialog/ConfirmationDialog';
+import LoadingOverlay from '#components/LoadingOverlay/LoadingOverlay';
+import type { Config } from '#types/Config';
+import DevStackTrace from '#components/DevStackTrace/DevStackTrace';
 
-const fallbackConfig = import.meta.env.APP_DEMO_DEFAULT_CONFIG_ID;
+const fallbackConfig = import.meta.env.APP_DEMO_FALLBACK_CONFIG_ID;
```
AntonLantukh, line 19: Can we use a config from `webapp.demo.ini` here?

src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -1,80 +1,178 @@
...
+    else if (configSource === selectedConfigSource) {
+      await configQuery.refetch();
+    }
+    // Get a new config by triggering a query param change
+    else {
+      navigateCallback(configSource);
+    }
+  };
+
+  const [state, setState] = useState<State>(initialState);
```
AntonLantukh: I would suggest to move it before `configNavigate`, just to know about it already when analysing `configNavigate`

src/stores/SettingsController.ts
```diff
@@ -0,0 +1,26 @@
...
+export const initSettings = async () => {
+  const settings = await fetch('/.webapp.ini')
+    .then((result) => result.text())
+    .then((iniString) => ini.parse(iniString) as Settings);
+
+  if (!settings) {
+    throw new Error('Unable to load .webapp.ini');
+  }
+
+  // This will result in an unusable app
```
AntonLantukh, line 14: Maybe we can include a link to the documentation here to explain all the options in details?

src/utils/common.ts
```diff
@@ -58,7 +58,7 @@ export const IS_DEMO_MODE = import.meta.env.MODE === 'demo';
 export const IS_TEST_MODE = import.meta.env.MODE === 'test';
 
 export function logDev(message: unknown, ...optionalParams: unknown[]) {
-  if (IS_DEVELOPMENT_BUILD && !IS_TEST_MODE) {
+  if (IS_DEVELOPMENT_BUILD) {
```
AntonLantukh, line 61: Do we need to include logs in tests?

src/components/Root/Root.tsx
```diff
@@ -1,32 +1,77 @@
...
+    refetchInterval: false,
+  });
+
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const configSource = useMemo(() => getConfigSource(searchParams, settingsQuery.data), [searchParams, settingsQuery.data]);
+
+  // Update the query string to maintain the right params
+  useEffect(() => {
+    if (settingsQuery.data && cleanupQueryParams(searchParams, settingsQuery.data, configSource)) {
```
AntonLantukh, line 32: Do you still think that we need to clean them? Maybe we can just replace the legacy param but leave invalid config ids as is? It seems that in case someone accidentally types a wrong id it won't be clear what exactly went wrong.

AntonLantukh:
Like the way it looks with our logo 👍 

Screenshot 2022-11-30 at 21 37 22

AntonLantukh:
@dbudzins When I click on a Submit button second time 'doctype' error appears. Looks like the content of the page gets loaded.

Screenshot 2022-11-30 at 21 48 25

No `.json` extension is included:

Screenshot 2022-11-30 at 21 53 24

budzworthy:
@AntonLantukh - If you enter 8 digits, it tries to load from the app config api which appends .json.  If you enter any other number of digits, it treats it as an absolute or relative url (in this case relative) so it does in fact end up fetching the page contents because everything routes to index.html.

src/utils/common.ts
budzworthy: It can help to debug when running test mode.  The tests on github run with production build so logs will get excluded. (Remember development / production builds, dev / prod / test / demo / jwdev modes)

AntonLantukh:
> @AntonLantukh - If you enter 8 digits, it tries to load from the app config api which appends .json. If you enter any other number of digits, it treats it as an absolute or relative url (in this case relative) so it does in fact end up fetching the page contents because everything routes to index.html.

@dbudzins Oh yeah, I completely forgot about it when using new interface. In the input we mention only configuration id and not the full url, the same works for the description bellow.

src/App.tsx
budzworthy: I took it out because of this: https://react.i18next.com/latest/i18nextprovider
"You will only need to use the provider if you need to support multiple i18next instances"

src/components/DemoConfigDialog/DemoConfigDialog.tsx
budzworthy: I agree, it kind of makes sense to have all the config sources in the same place, but the reason I didn't put it into the settings is that it's a thing that no customers should probably need to touch. Having it in the settings makes it more visible outside of the demo component.  What do you think?

src/components/DemoConfigDialog/DemoConfigDialog.tsx
budzworthy: Ah, good catch. Kind of crazy that it compiled like this.  I guess it was all getting hoisted, but it's definitely weird. Fixed it!

src/stores/SettingsController.ts
budzworthy: Actually, this should propagate to the error dialog in Root.tsx, which has a "learn more" link that points to the readme about config settings. (There was a bug there in the link that I just fixed, and I added a note about this condition as well)

src/components/Root/Root.tsx
budzworthy: The main reason for this is actually to make sure the app-config param stays on all the urls as the app navigates. In theory we could update every link and navigate to make sure it includes the query param, but this seemed more robust.

budzworthy:
> > @AntonLantukh - If you enter 8 digits, it tries to load from the app config api which appends .json. If you enter any other number of digits, it treats it as an absolute or relative url (in this case relative) so it does in fact end up fetching the page contents because everything routes to index.html.
> 
> @dbudzins Oh yeah, I completely forgot about it when using new interface. In the input we mention only configuration id and not the full url, the same works for the description bellow.

Ya, I thought it was nice to allow url refs here too, but that's probably going to cause more confusion than it's worth. 

> > @AntonLantukh - If you enter 8 digits, it tries to load from the app config api which appends .json. If you enter any other number of digits, it treats it as an absolute or relative url (in this case relative) so it does in fact end up fetching the page contents because everything routes to index.html.
> 
> @dbudzins Oh yeah, I completely forgot about it when using new interface. In the input we mention only configuration id and not the full url, the same works for the description bellow.

@AntonLantukh I updated the behavior to only work with ID's and give better feedback that we expect an 8 character alphanumeric.  I think that addresses everything if you don't mind re-reviewing.

src/components/DemoConfigDialog/DemoConfigDialog.tsx
AntonLantukh: We also expose .env files to GitHub, so I think it could be confusing. I would suggest to store it in `templates/webapp.demo.ini` as a default config. The component is named 'DemoConfigDialog' and it seems logical to me that it uses a demo config as a fallback source.

AntonLantukh (approved):

PR #198, opened 2022-11-30
feat: InPlayer authentication

kmitrovv:
## Description

This PR contains setup of the InPlayer SDK (a library which contains all the needed APIs) that will speed up the development process and implementation of the authentication flow.

Login and logout are functional if InPlayer config is provided. At the moment we load it from an external [url](https://web-ott.s3.eu-west-1.amazonaws.com/apps/configs/demo.json) until the InPlayer config integration is added on the JW dashboard.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


.env.dev
```diff
@@ -1,3 +1,3 @@
 APP_CONFIG_DEFAULT_SOURCE=gnnuzabk
 APP_UNSAFE_ALLOW_DYNAMIC_CONFIG=1
-APP_API_BASE_URL=https://cdn.jwplayer.com
+APP_API_BASE_URL=https://cdn.jwplayer.com
```
budzworthy: I suggest making sure the 'newline at end of files' is enabled in your IDE of choice:
- https://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file
- https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

Let's make sure we sync up our ide settings on all the little formatting things like this to avoid unnecessary noise.

.gitignore
```diff
@@ -13,3 +13,6 @@ firebase-debug.log
 .idea
 .DS_Store
 .vscode/
+
+# env
+.env.dev
```
budzworthy: This should not be ignored.  Please use .env.dev.local if you need to override settings: https://vitejs.dev/guide/env-and-mode.html#env-files

package.json
```diff
@@ -16,7 +16,7 @@
     "test": "TZ=UTC vitest run",
     "test-watch": "TZ=UTC vitest",
     "test-coverage": "TZ=UTC vitest run --coverage",
-    "test-commit": "TZ=UTC vitest run --changed HEAD~1 --coverage",
+    "test-commit": "TZ=UTC LC_ALL=en_US.UTF-8 vitest run --changed HEAD~1 --coverage",
```
budzworthy, line 19: Nice easy fix!  I suggest you add it to all the `test` yarn commands or the other commands will end up having the same problem.

src/App.tsx
```diff
@@ -37,6 +40,12 @@ class App extends Component {
       await initializeAccount();
     }
 
+    if (config?.integrations?.inplayer?.clientId) {
+      // this is temporary until the config is loaded from jw side
+      InPlayer.setConfig(import.meta.env.APP_INPLAYER_SDK);
+      await initializeInPlayerAccount();
```
budzworthy: You might want to sync this with the latest PR, which cleans up and refactors the app init code: https://github.com/jwplayer/ott-web-app/pull/197

.env.dev
kmitrovv: Hmm, `the newline at end of files` is enabled, i guess it does not work for env files. Will fix that

package.json
kmitrovv: Sure thing!

src/services/config.service.ts
```diff
@@ -65,6 +71,7 @@ const configSchema: SchemaOf<Config> = object({
   features: featuresSchema.notRequired(),
   integrations: object({
```
budzworthy, line 72: We are probably going to want to rethink how the 'integrations' config settings are set from the dashboard to prevent wonky configurations like having a full cleeng and full inplayer config set.  For now, we should probably at least throw an error in `validateConfig` if both are present and/or have `clientId` set.

src/App.tsx
kmitrovv: I guess I'll wait for it to be merged with develop

ini/.webapp.dev.ini
```diff
@@ -0,0 +1,4 @@
+; This is a basic Blender demo
```
budzworthy, line 1: This file should be git ignored

src/containers/Layout/Layout.tsx
```diff
@@ -27,8 +28,8 @@ const Layout = () => {
   const navigate = useNavigate();
   const { t } = useTranslation('common');
   const { config, accessModel } = useConfigStore(({ config, accessModel }) => ({ config, accessModel }), shallow);
-  const { menu, assets, siteName, description, integrations, styling, features } = config;
-  const cleengId = integrations?.cleeng?.id;
+  const { menu, assets, siteName, description, styling, features } = config;
+  const { clientId } = useClientIntegration();
```
budzworthy, line 29: Maybe a name like `authServiceId` or `authProviderId` would make this more intuitive?

src/hooks/useClientIntegration.ts
```diff
@@ -0,0 +1,25 @@
...
+const useClientIntegration = () => {
+  const {
+    config: { integrations },
+  } = useConfigStore.getState();
+
+  const isInPlayer = !!integrations?.inplayer?.clientId;
+  const client = isInPlayer ? ClientIntegrations.INPLAYER : ClientIntegrations.CLEENG;
+  const clientId = isInPlayer ? integrations?.inplayer?.clientId : integrations?.cleeng?.id;
+
+  // TODO: More data will follow. Example: access fees will be fetched and returned once user is auth.
```
budzworthy: I'm not sure about this TODO.  This hook seems ok as a way to tell components which service we're using, but other logic should probably still be in the controllers and stores.  This logic could also go in a store, so that it's not recalculated every call as well.

types/account.d.ts
```diff
@@ -1,12 +1,19 @@
...
 
+import type { Config } from '#types/Config';
+
 export type AuthData = {
   jwt: string;
   customerToken: string;
   refreshToken: string;
 };
 
+export type InPlayerAuthData = {
```
budzworthy: It probably makes sense to split these into 3 files (generic account types, cleeng, and inplayer)

src/utils/common.ts
```diff
@@ -82,3 +85,29 @@ export function getOverrideIP() {
 export function testId(value: string | undefined) {
   return IS_DEVELOPMENT_BUILD || IS_TEST_MODE ? value : undefined;
 }
+
+// responsible to convert the InPlayer object to be compatible to the store
+export function processInplayerAccount(account: AccountData): Customer {
```
budzworthy: These methods are probably only ever going to be used in the inplayer account service, so maybe they would be better encapsulated by being non-exported methods there?

src/stores/ConfigStore.ts
```diff
@@ -45,4 +57,13 @@ export const useConfigStore = createStore<ConfigState>('ConfigStore', (_, get) =
 
     return { cleengId, cleengSandbox, monthlyOfferId, yearlyOfferId };
   },
+  getInPlayerData: (): InPlayerData => {
```
budzworthy: I think these methods (this and the cleeng one) can move to be internal methods in the respective services. They are really just syntactical-sugar helpers to reduce deep level property indexing on the config object...

src/utils/configLoad.ts
```diff
@@ -59,10 +54,19 @@ const maybeInjectAnalyticsLibrary = (config: Config) => {
 };
 
 const calculateAccessModel = (config: Config): AccessModel => {
-  const { id, monthlyOffer, yearlyOffer } = config?.integrations?.cleeng || {};
+  if (config?.integrations?.cleeng) {
```
budzworthy, line 60: We should check to make sure that both integrations aren't set and throw an error if they are. Otherwise trying to run with inplayer might get the wrong access model set from left over cleeng data in a config or something.

src/stores/AccountController.ts
```diff
@@ -450,3 +449,19 @@ function useLoginContext<T>(callback: (args: { cleengId: string; cleengSandbox:
 
   return useConfig((config) => callback({ ...config, customerId: user.id, auth }));
 }
+
+function withAccountService<T>(
```
budzworthy, line 453: fyi, I'm not set on this pattern if you find a cleaner, easier path to take. It was just my first attempt at refactoring as quick as I could yesterday.  It's also fine for us to leave it this way and circle back to refactor to something better later.

src/stores/AccountController.ts
```diff
@@ -141,66 +132,70 @@ const refreshJwtToken = async (sandbox: boolean, auth: AuthData) => {
...
+      activePayment: null,
+      customerConsents: null,
+      publisherConsents: null,
+    });
 
-  await restoreFavorites();
-  await restoreWatchHistory();
+    await restoreFavorites();
+    await restoreWatchHistory();
+    await accountService.logout();
```
budzworthy, line 179: Maybe safer to call accountservice.logout before the restore calls. They are intended to restore the local-storage instance data for watchlists for logged out scenario

src/stores/AccountController.ts
```diff
@@ -141,66 +132,70 @@ const refreshJwtToken = async (sandbox: boolean, auth: AuthData) => {
...
-    auth: null,
-    user: null,
-    subscription: null,
-    transactions: null,
-    activePayment: null,
-    customerConsents: null,
-    publisherConsents: null,
-  });
+  await withAccountService(async ({ accountService }) => {
+    persist.removeItem(PERSIST_KEY_ACCOUNT);
```
budzworthy, line 162: Does the inplayer sdk need this or is similar functionality wrapped into the sdk already?  If it doesn't need it, maybe it makes sense to put this into the cleeng service's logout method?

src/stores/AccountController.ts
```diff
@@ -141,66 +132,70 @@ const refreshJwtToken = async (sandbox: boolean, auth: AuthData) => {
...
-    auth: auth,
-    user: response.responseData,
-  });
+    useAccountStore.setState({
+      auth: auth,
+      user: response,
+    });
 
-  await Promise.allSettled([accessModel === 'SVOD' ? reloadActiveSubscription() : Promise.resolve(), getCustomerConsents(), getPublisherConsents()]);
+    await getAccountExtras(accessModel);
```
budzworthy: What was the reason for changing the afterlogin method and using getAccountExtras and getAccount instead? Seems like the end result is the same but now there's some duplicated code in the login method.

src/stores/AccountController.ts
```diff
@@ -30,12 +31,13 @@ export const authNeedsRefresh = (auth: AuthData): boolean => {
 
 export const setJwtRefreshTimeout = () => {
   const auth = useAccountStore.getState().auth;
-  const { cleengSandbox } = useConfigStore.getState().getCleengData();
+
+  if (!auth?.refreshToken) return;
```
budzworthy, line 36: Is this basically to skip this code for inplayer?  If so, can we at least add a comment here to that effect?

budzworthy (commented):
I left some questions and comments, but overall I'm pretty happy with how this is looking. It seems to be the same number of files changes as what we looked at Wednesday, but now pretty much all the changes are isolated to adding service code and small refactors to controllers.

ini/.webapp.dev.ini
kmitrovv: resolved with the merge from your pr

src/containers/Layout/Layout.tsx
kmitrovv: Hmm, not sure if `authServiceId` is semantic enough... Maybe integrationId.. not sure what's best :D

src/hooks/useClientIntegration.ts
kmitrovv: Yes, the TODO will be removed. The access fees will be created as a service.

src/stores/ConfigStore.ts
kmitrovv: Removed the `InPlayer` part since it's not used yet. We can plan to refactor the cleeng part at later stage.

src/stores/AccountController.ts
kmitrovv: It's okay for now, maybe we can think of refactor once we dive more deeply into the project.

src/stores/AccountController.ts
kmitrovv: I think it's okay like this, since for the cleeng part it always returns true, but for the `inplayer` part it invalidates the additional localStorage that is needed for the SDK. Nothing that concerns the current flow.

src/stores/AccountController.ts
kmitrovv: We adapted the entire logic about authentication as the current flow. We use the jwapp.auth as well for the app. The SDK localStorage is the same as the `jwapp.auth` but has no effect on the ott app. It comes out of the box with the SDK so that the endpoints can be tokenised.

src/stores/AccountController.ts
kmitrovv: Wrapped the duplicate code into new function called the same, `afterLogin`.  Btw the method is changed because the logic that was in the previous `afterLogin` method for getting the user at the moment is done in the services level.

darkoatanasovski (approved):

src/utils/configLoad.ts
```diff
@@ -107,7 +107,11 @@ export async function loadAndValidateConfig(configSource: string | undefined) {
 
   maybeInjectAnalyticsLibrary(config);
 
-  if (config?.integrations?.cleeng?.id) {
+  // TODO: refactor this once we have more input how integrations will be handled in dashboard
+  if (config?.integrations?.cleeng?.id && config?.integrations?.inplayer?.clientId) {
+    throw new Error('Invalid client integration.');
```
budzworthy:
```suggestion
    throw new Error('Invalid client integration. You cannot have both Cleeng and Inplayer integrations enabled at the same time.');
```

src/containers/Layout/Layout.tsx
budzworthy: I think the integration that we care about here will always at least be an auth service (it may be an auth+subscription service, but it always needs auth) so I would go more specific than generic, especially because it's used for login based checks here. Not a blocker if we can't agree on a more readable name for now, maybe it will become more obvious in the way we talk about it as the project progresses.

budzworthy (approved):

PR #199, opened 2022-12-01
Inplayer/update account

darkoatanasovski:
## Description

implemented update InPlayer account details.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


ini/.webapp.dev.ini
```diff
@@ -0,0 +1,4 @@
+; This is a basic Blender demo
```
budzworthy: This file should be excluded by .gitignore

darkoatanasovski: ah.. yes :)

src/components/Form/FormSection.tsx
```diff
@@ -171,7 +173,7 @@ export function FormSection<TData extends GenericFormValues>({
             (typeof editButton === 'object' ? (
               (editButton as ReactElement)
             ) : (
-              <Button label={editButton as string} type="button" onClick={onEdit} disabled={isLoading} />
+              <Button label={editButton as string} type="button" onClick={onEdit} disabled={isLoading || isDisabled} />
```
budzworthy: Maybe it's better to call the new property `readOnly` and then just totally hide the edit button.  Otherwise viewers will probably wonder why the button is disabled and create support requests, etc.

darkoatanasovski: hmm.. make sense. I agree

src/services/inplayer.account.service.ts
```diff
@@ -52,6 +52,34 @@ export const getUser = async (): Promise<Customer> => {
 
 export const getFreshJwtToken = async ({ auth }: { auth: AuthData }) => auth;
 
+export const updateCustomer: UpdateCustomer = async (values) => {
```
budzworthy, line 103: Really nice that you can just ignore the cleeng params and the method signature works without any other chanes.

types/account.d.ts
```diff
@@ -146,6 +146,7 @@ export type UpdateCustomerPayload = {
   firstName?: string;
   lastName?: string;
   externalData?: ExternalData;
+  consents?: CustomerConsent[];
```
budzworthy: Does the inplayer sdk work the same way that cleeng does where partial updates are allowed without any problems?  That is, can  I just call with `{firstName: 'New Name'}` in order to just update the firstName and leave all other field the same?

darkoatanasovski: yes, only `fullName` is required field

src/services/inplayer.account.service.ts
```diff
@@ -52,6 +52,34 @@ export const getUser = async (): Promise<Customer> => {
...
+  try {
+    const consents: { [key: string]: string } = {};
+    values.consents?.map((consent: CustomerConsent) => {
```
budzworthy: It doesn't look like the Consents are set via this path anywhere.  I think it would be better to do a different `InPlayer.Account.updateAccount` call in a `updateCustomerConsents` method for the inplayer service.  They are saved separately anyhow.

The other approach would be to keep updateCustomer like this, but then refactor the Cleeng service to make 2 calls (1 to cleeng's update customer endpoint and 1 to cleeng's update consents endpoint.)  However, since the save calls are independent anyhow I lean towards keeping the 2 methods (`updateCustomer` and `updateCustomerConsents`.)  This also is less refactoring and gives a path towards adding customer consents as a top level resource if that becomes necessary instead of saving the data in customer fields.

types/account.d.ts
darkoatanasovski: @dbudzins sorry, it is `PUT` request, and we need to pass all user's data, but still, only `fullName` is required

src/services/inplayer.account.service.ts
```diff
@@ -52,6 +52,34 @@ export const getUser = async (): Promise<Customer> => {
...
+      if (consent.label) {
+        const { customerId, date, newestVersion, needsUpdate, ...rest } = consent;
+        consents[`consents_${consent.name}`] = JSON.stringify(rest);
```
budzworthy: Would it be easier to 'find' the consents if they were in a single field called `customerConsents` that contained the json for all consents? As it currently stands, all of the update calls send all of the consent values already anyhow, so we wouldn't need to worry about partial data and it means less looping and string manipulation because we don't have to read, write, or search for the consent_ names.

src/services/inplayer.account.service.ts
```diff
@@ -52,6 +52,34 @@ export const getUser = async (): Promise<Customer> => {
...
+      }
+    });
+    const response = await InPlayer.Account.updateAccount({
+      fullName: `${values.firstName} ${values.lastName}`,
```
budzworthy: I see that fullname is a built in property in the inplayer api. I'm a little worried about denormalizing the data in to 2 sets of fields that it can get out of sync.  Especially if anyone is using inplayer api's as part of a partner integration or another app besides this web app.

Maybe we can just use the fullname field and add our own unique delimiter like a tab between first and last name or something?  We could store it as a json blob, but that also seems problematic. How do you normally work with Fullname? Have you run into this problem before?

src/services/inplayer.account.service.ts
darkoatanasovski: I think that if we store JSON as a string in the metadata, it will be hard to manipulate the data(e.g. insert or update after parsing the JSON string). But we can try that way :)

src/services/inplayer.account.service.ts
darkoatanasovski: In this case, we set `fullName` because this is a mandatory field, but it is never used. We set first and last names as a part of the metadata. 
Yes, you are right, it will be problematic for the existing users without first and last names.
For this case, we can have a fallback, if the first/last name is not set, we can use `fullName` with a delimiter( e.g. the first white space )

src/services/inplayer.account.service.ts
budzworthy: Certainly if we did individual inserts and updates I would agree, but update consents always takes an array of all consents anyhow.

src/services/inplayer.account.service.ts
budzworthy: Again though, this api will likely be connected to more apps than just the web app, and I don't want to have to make sure all of them know to use the first/last meta fields and ignore full name. I think it's more maintainable to always use fullname and not add the other fields.

types/account.d.ts
budzworthy: Ah, I see. OK, that makes it a bit more difficult then.

budzworthy: Probably to be safest and most flexible, we should pass in the 'current user data' and then keep this payload just for the update data. Then each cleeng and inplayer service implementation can do internally whatever they need to do with the data.

darkoatanasovski: yes, something like this maybe: https://github.com/jwplayer/ott-web-app/blob/e740cdc4ef5099d78d89e7724aa67e6ea8ae0168/src/stores/AccountController.ts#L110

src/services/inplayer.account.service.ts
```diff
@@ -54,50 +54,44 @@ export const getFreshJwtToken = async ({ auth }: { auth: AuthData }) => auth;
...
 export const updateCustomer: UpdateCustomer = async (values) => {
   try {
-    const consents: { [key: string]: string } = {};
-    values.consents?.map((consent: CustomerConsent) => {
-      if (consent.label) {
-        const { customerId, date, newestVersion, needsUpdate, ...rest } = consent;
-        consents[`consents_${consent.name}`] = JSON.stringify(rest);
-      }
-    });
+    const fullName = `${values.firstName} ${values.lastName}`;
```
budzworthy: This is kind of a hacky suggestion, but I think it will work to give us slightly better data results - what if we just use two spaces to delimit between first and last name? We can even sanitize the first name and last name inputs here to replace any double spaces with single spaces beforehand. That way we can reliably split on a double space to get back first and last name even when either part has spaces in it itself (i.e. 'Vincent van Gogh' and 'Mary Grace Johnson' will both parse to 'Vincent' / 'van Gogh' and 'Mary Grace' / 'Johnson' appropriately.)  It also will be virtually unnoticeable if there's anywhere else that just renders full name.

Alternatively we could use a comma like 'van Gogh, Vincent'.

src/pages/User/User.tsx
```diff
@@ -96,7 +97,10 @@ const User = (): JSX.Element => {
       )}
       <div className={styles.mainColumn}>
         <Routes>
-          <Route path="my-account" element={<AccountComponent panelClassName={styles.panel} panelHeaderClassName={styles.panelHeader} />} />
+          <Route
+            path="my-account"
+            element={<AccountComponent panelClassName={styles.panel} panelHeaderClassName={styles.panelHeader} canUpdateEmail={!integration?.inplayer} />}
```
budzworthy: I think it is more scaleable if we inspect the actual controller / service functionality instead of just making everything an if inplayer / cleeng check.  For example, this should be coming from accountController.canUpdateEmail() which will return `accountService.updateEmail !== undefined` (I assume here to split updateUser and updateEmail into 2 methods instead of 1 method with 2 arg types.)

Essentially, we should look at the capabilities of the services we're using, not just the config. It is more semantic, doesn't require inferring how configs and api's work, and will scale nicely as additional functionality or services are added (for example, if we ever add email update functionality to inplayer, we can just add it to the service and everything else will just work without any additional changes)

src/services/inplayer.account.service.ts
budzworthy: I think probably the comma solution is a better one. I noticed that we can "split" the fields in the inplayer dashboard. Do we need to handle anything different for that case?

darkoatanasovski: Yes, in the InPlayer Dashboard you can split the `fullName` field to **first_name** and **surname**, but this is just metadata and `fullName` remains as it is.
If we decide to go with the comma solution, the existing accounts will see only the first name.

I suggest going with first and last name in the metadata if exists, or as a fallback to whitespace solution

budzworthy: OK, then let's use the split metadata fields that inplayer supports (first_name and surname) and make sure to also write to fullname on save.  Then fullname should only be used as a fallback if first_name and surname aren't present.

src/services/inplayer.account.service.ts
darkoatanasovski: because `fullName` is a required field for InPlayer `updateCustomer`, we have to use this method to update the user consents or any other user extras

types/account.d.ts
budzworthy: We shouldn't add or remove from the Payload objects. They define the cleeng api interfaces.

src/services/inplayer.account.service.ts
```diff
@@ -44,32 +44,62 @@ export const logout = async () => {
...
 export const getFreshJwtToken = async ({ auth }: { auth: AuthData }) => auth;
 
+export const updateCustomer: UpdateCustomer = async (values) => {
+  try {
+    const fullName = `${values.firstName} ${values.lastName}`;
+    const response = await InPlayer.Account.updateAccount({
+      fullName,
+      metadata: {
+        ...(values?.consents && { consents: JSON.stringify(values.consents) }),
+        first_name: values.firstName?.replace(/\s\s+/g, ' ')?.trim() || '',
```
budzworthy: Why did you end up replacing multi-spaces with single spaces if we aren't using a double space between first and last name?

src/stores/AccountController.ts
```diff
@@ -293,6 +296,12 @@ export const updateCaptureAnswers = async (capture: Capture) => {
   });
 };
 
+export const setCanUpdateEmail = async () => {
```
budzworthy: None of the async await is needed here is it?

src/stores/AccountController.ts
```diff
@@ -62,6 +62,7 @@ export const initializeAccount = async () => {
   await withAccountService(async ({ accountService, config }) => {
     useAccountStore.setState({ loading: true });
     accountService.setEnvironment(config);
+    await setCanUpdateEmail();
```
```suggestion
    useAccountStore.setState({ canUpdateEmail: accountService.canUpdateEmail() });
```

src/stores/AccountController.ts
darkoatanasovski: yes, otherwise the tests will fails

src/stores/AccountController.ts
budzworthy: This seems simpler to just leave inline

src/stores/AccountController.ts
```diff
@@ -100,23 +101,25 @@ export const initializeAccount = async () => {
 };
 
 export async function updateUser(values: { firstName: string; lastName: string } | { email: string; confirmationPassword: string }) {
-  const { auth, user } = useAccountStore.getState();
+  await withAccountService(async ({ accountService, sandbox }) => {
+    useAccountStore.setState({ loading: true });
 
-  if (!auth || !user) throw new Error('no auth');
+    const { auth, user } = useAccountStore.getState();
```
```suggestion
    const { auth, user, canUpdateEmail } = useAccountStore.getState();
    
    if ('email' in values && !canUpdateEmail) {
      throw new Error('Email update not supported');
    }
```

src/stores/AccountController.ts
budzworthy: Which test? I can't imagine how this would directly impact a test

darkoatanasovski: https://github.com/jwplayer/ott-web-app/actions/runs/3638608893/jobs/6141227698

without async/await, the integrations from config is undefined

budzworthy: That is weird. I wonder if it's a race condition that async somehow works around. Regardless, the inline should avoid integrations anyhow.

src/stores/AccountController.ts
```diff
@@ -293,6 +300,12 @@ export const updateCaptureAnswers = async (capture: Capture) => {
   });
 };
 
+export const setCanUpdateEmail = async () => {
+  await withAccountService(async ({ accountService }) => {
+    useAccountStore.setState({ canUpdateEmail: accountService.canUpdateEmail() });
+  });
+};
+
```
```suggestion
```

src/stores/AccountController.ts
darkoatanasovski: yes, exactly

src/stores/AccountController.ts
darkoatanasovski: this fails on lint

src/services/inplayer.account.service.ts
darkoatanasovski: should we allow more than one space in the name? 
e.g. John     Doe

src/services/inplayer.account.service.ts
```diff
@@ -44,32 +44,62 @@ export const logout = async () => {
...
 
 export const getFreshJwtToken = async ({ auth }: { auth: AuthData }) => auth;
 
+export const updateCustomer: UpdateCustomer = async (values) => {
+  try {
+    const fullName = `${values.firstName} ${values.lastName}`;
+    const response = await InPlayer.Account.updateAccount({
+      fullName,
+      metadata: {
+        ...(values?.consents && { consents: JSON.stringify(values.consents) }),
```
budzworthy: I don't see that consents is actually set anywhere before calling this. Also, for inplayer, don't we need to make sure that this isn't just a partial, but actually has all of the user data, otherwise we will lose anything that's not included because of the `PUT`?

I thought what we were going to do is to also pass in the current user data and consents and combine the update values with the current values.

darkoatanasovski: the consents are set in the `inplayer/register` branch. In this PR, we handle first and last names, but the method is prepared to handle user's consents as well

src/services/inplayer.account.service.ts
budzworthy: Trimming makes sense before we concatenate the first/last names, but I don't now if there's a lot more manipulation we should do on the front end. I assume some of this already happens in the backend.

src/services/inplayer.account.service.ts
budzworthy: OK, what about the issue with the partial data and PUT?

darkoatanasovski: we'll always send whole user data...

src/stores/AccountController.ts
```diff
@@ -104,7 +104,11 @@ export async function updateUser(values: { firstName: string; lastName: string }
   await withAccountService(async ({ accountService, sandbox }) => {
     useAccountStore.setState({ loading: true });
 
-    const { auth, user } = useAccountStore.getState();
+    const { auth, user, canUpdateEmail } = useAccountStore.getState();
+
+    if (Object.prototype.hasOwnProperty.call(values, 'email') && !canUpdateEmail) {
+      throw new Error('Email update not supported');
```
darkoatanasovski, line 119: I don't think that we need this validation here

budzworthy: Never hurts to have sanity checks. Maybe the code is written so that this can never happen now, but what about when someone refactors later?

darkoatanasovski: ok, then, we can go with this check :)

types/account.d.ts
darkoatanasovski: this is also improved in the commits below

src/services/inplayer.account.service.ts
darkoatanasovski: this is also fixed I think

PR #201, opened 2022-12-05
Inplayer/reset password

darkoatanasovski:
## Description

Implemented change and forgot password flows. Also, added new PasswordField component

- [x] Acceptance criteria met
- [x] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/src/components/EditPasswordForm/inplayer/EditPasswordForm.tsx
+++ b/src/components/EditPasswordForm/inplayer/EditPasswordForm.tsx
@@ -1,0 +1,88 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+import styles from '../EditPasswordForm.module.scss';
+
+import type { FormErrors } from '#types/form';
+import type { EditPasswordFormData } from '#types/account';
+import FormFeedback from '#components/FormFeedback/FormFeedback';
+import TextField from '#components/TextField/TextField';
+import Button from '#components/Button/Button';
+import LoadingOverlay from '#components/LoadingOverlay/LoadingOverlay';
+import { testId } from '#src/utils/common';
+import PasswordField from '#src/components/PasswordField/PasswordField';
+import { useAccountStore } from '#src/stores/AccountStore';
+
+type Props = {
+  onSubmit: React.FormEventHandler<HTMLFormElement>;
+  onChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement>;
+  onBlur: React.FocusEventHandler<HTMLInputElement | HTMLTextAreaElement>;
+  error?: string;
+  errors: FormErrors<EditPasswordFormData>;
+  value: EditPasswordFormData;
+  submitting: boolean;
+};
+
+const EditPasswordForm: React.FC<Props> = ({ onSubmit, onChange, onBlur, value, errors, submitting }: Props) => {
+  const { t } = useTranslation('account');
+  const user = useAccountStore.getState().user;
+
+  return (
+    <form onSubmit={onSubmit} data-testid={testId('change-password-form')} noValidate className={styles.forgotPasswordForm}>
+      <h2 className={styles.title}>{t('reset.password_reset')}</h2>
+      {errors.form ? <FormFeedback variant="error">{errors.form}</FormFeedback> : null}
+      {!user ? (
+        <TextField
+          className={styles.textField}
+          value={value.resetPasswordToken}
+          onChange={onChange}
+          onBlur={onBlur}
+          label={t('reset.reset_password_token')}
+          placeholder={t('reset.reset_password_token')}
+          error={!!errors.resetPasswordToken}
+          name="resetPasswordToken"
+          type="text"
+          required
+        />
+      ) : (
+        <PasswordField
+          value={value.oldPassword}
+          onChange={onChange}
+          onBlur={onBlur}
+          label={t('reset.old_password')}
+          placeholder={t('reset.old_password')}
+          error={!!errors.oldPassword}
+          name="oldPassword"
+          showToggleView={false}
+          showHelperText={false}
+          required
+        />
+      )}
+
+      <PasswordField
+        value={value.password}
+        onChange={onChange}
+        onBlur={onBlur}
+        label={t('reset.new_password')}
+        placeholder={t('reset.password')}
+        error={!!errors.password}
+        name="password"
+        required
+      />
+      <PasswordField
+        value={value.passwordConfirmation}
+        onChange={onChange}
+        onBlur={onBlur}
+        label={t('reset.new_password')}
+        placeholder={t('reset.password')}
+        error={!!errors.passwordConfirmation}
+        name="passwordConfirmation"
+        required
+      />
+      <Button type="submit" className={styles.button} fullWidth color="primary" disabled={submitting} label={t('reset.confirm')} />
+      {submitting && <LoadingOverlay transparentBackground inline />}
+    </form>
+  );
+};
+
+export default EditPasswordForm;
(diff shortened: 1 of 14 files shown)


src/components/PasswordField/PasswordField.tsx
```diff
@@ -0,0 +1,57 @@
...
+  showHelperText?: boolean;
+};
+
+const PasswordField: React.FC<Props> = ({ value, showToggleView = true, showHelperText = true, ...props }: Props) => {
+  const { t } = useTranslation('account');
+  const [viewPassword, toggleViewPassword] = useToggle();
+
+  return (
+    <TextField
+      data-testid={'alb'}
```
budzworthy: What does this test id mean?

darkoatanasovski: ahhh... forgot to remove :)

src/components/EditPasswordForm/inplayer/EditPasswordForm.tsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
```
budzworthy, line 1: I would suggest we rename the current EditPasswordForm.tsx to ResetPasswordForm.tsx and create this one in its place without the inplayer subdirectory.

darkoatanasovski: hmm... I am not sure. we have different flows. Cleeng resets the password via email.
InPlayer has 2 ways of changing the password:

1. With an old password
2. With reset token ( forgot password flow )

https://www.youtube.com/watch?v=UNcvHDsQJV4

budzworthy: That's fine. Those are indeed distinct flows, which should have distinct UI's, but let's not couple them to auth provider. Someone could add another provider and if that provider supports a given flow, the correct, preferred UI should be shown.  So `ResetPasswordRequestForm`, `EditPasswordRequestForm`, and `EditPasswordForm` or something like that.

dbudzins: I guess the edit password and reset by token flows are similar enough there's no reason to split them. I think `EditPasswordForm` and `ResetPasswordForm` or `ResetPasswordRequestForm` or something are discernible enough names to differentiate the 2.

src/components/PasswordField/PasswordField.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
budzworthy, line 1: Really nice that you wrapped this functionality into a component! It would be awesome to back-port this to the login and register pages, which will also give us some free e2e test coverage on things like the show/hide toggle and strength feedback.

src/containers/AccountModal/forms/EditPassword.tsx
```diff
@@ -4,49 +4,72 @@ import { object, string } from 'yup';
...
 
   const passwordForm = useForm(
-    { password: '' },
+    { resetPasswordToken: '', oldPassword: '', password: '', passwordConfirmation: '' },
     passwordSubmitHandler,
     object().shape({
+      resetPasswordTokenParam: string(),
+      oldPassword: string(),
       password: string()
         .matches(/^(?=.*[a-z])(?=.*[0-9]).{8,}$/, t('registration.invalid_password'))
```
budzworthy, line 70: Can this regex be build into the new Password field instead of being here?

src/containers/AccountModal/forms/ResetPassword.tsx
```diff
@@ -66,8 +68,9 @@ const ResetPassword: React.FC<Prop> = ({ type }: Prop) => {
     const resetUrl = `${window.location.origin}/?u=edit-password`;
 
     try {
-      await resetPassword(formData.email, resetUrl);
-      navigate(addQueryParam(location, 'u', 'send-confirmation'));
+      await resetPassword({ customerEmail: formData.email, resetUrl });
```
budzworthy, line 71: I'm confused why this is calling reset password and then going to edit-password.  I thought edit-password was only accessible via email links in the password reset email?  What does it mean to edit a password after resetting the password?

src/components/EditPasswordForm/inplayer/EditPasswordForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <form onSubmit={onSubmit} data-testid={testId('change-password-form')} noValidate className={styles.forgotPasswordForm}>
+      <h2 className={styles.title}>{t('reset.password_reset')}</h2>
+      {errors.form ? <FormFeedback variant="error">{errors.form}</FormFeedback> : null}
+      {!user ? (
+        <TextField
+          className={styles.textField}
+          value={value.resetPasswordToken}
+          onChange={onChange}
+          onBlur={onBlur}
+          label={t('reset.reset_password_token')}
```
budzworthy, line 40: Where does this token come from? Is it sent in the password reset email?  Does it need to be visible to the user?

types/account.d.ts
```diff
@@ -129,10 +132,12 @@ export type ResetPasswordPayload = {
 };
 
 export type ChangePasswordPayload = {
```
budzworthy, line 134: I want to avoid changing these Payload objects since they are right now cleeng-specific (see LoginArgs vs LoginPayload as an example.)

src/services/inplayer.account.service.ts
```diff
@@ -52,6 +52,51 @@ export const getUser = async (): Promise<Customer> => {
...
+    if (resetPasswordToken) {
+      await changePasswordWithResetToken(resetPasswordToken, password, passwordConfirmation);
+    } else {
+      await changePasswordWithOldPassword(oldPassword, password, passwordConfirmation);
+    }
+  } catch {
+    throw new Error('Failed to change password.');
+  }
+};
+const changePasswordWithResetToken = async (token: string, password: string, passwordConfirmation: string) => {
```
budzworthy, line 83: I think I'd prefer to expose the 2 non-exported methods here instead of trying to force everything through a single interface. The cleeng change password implementation is the same as `changePasswordWithResetToken`

src/components/EditPasswordForm/inplayer/EditPasswordForm.tsx
darkoatanasovski: yes, this make sense

src/containers/AccountModal/forms/EditPassword.tsx
darkoatanasovski: if we move this regex validation to the `PasswordField`, it won't work for the `oldPassword` field for **InPlayer** because we use different validation, but it can be optional. 
e.g. to pass new parameter validation.
`<PasswordField validation="/^(?=.*[a-z])(?=.*[0-9]).{8,}$/" ... />`

What do you think about this solution?

src/containers/AccountModal/forms/ResetPassword.tsx
darkoatanasovski: Yes, you are right, `edit-password` is only accessible via email links in the password reset email. But this makes sense to me because first, you need to reset the password and get a temporary password or token in the email, then you need to update the password

src/components/EditPasswordForm/inplayer/EditPasswordForm.tsx
darkoatanasovski: Yes, it is sent in the password reset email and looks like this: `325a2693f931750a`, and yes, it should be visible to the users.

src/containers/AccountModal/forms/EditPassword.tsx
budzworthy: I guess lets leave it how it is for now then. It's working as-is and if we can't completely encapsulate it, there's no ROI.

PR #202, opened 2022-12-05
feat: InPlayer register flow

kmitrovv:
## Description

With this PR, InPlayer register flow is almost completed. It includes changes in pretty much all of the methods that are required for this flow, except the favourites and history. This is what will follow in the next PR.

Updates are mostly in the `AccountController.tsx`,  `cleeng.account.service.ts` and `inplayer.account.service.ts`.

No visual changes.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


src/containers/AccountModal/forms/Registration.tsx
```diff
@@ -54,10 +54,13 @@ const Registration = () => {
     } catch (error: unknown) {
       if (error instanceof Error) {
         const errorMessage = error.message.toLowerCase();
-        if (errorMessage.includes('customer already exists.')) {
```
kmitrovv, line 57: this is tweaked so it can match `InPlayer response as well`

src/services/inplayer.account.service.ts
```diff
@@ -44,36 +80,170 @@ export const logout = async () => {
...
   const regDate = new Date(createdAt * 1000).toLocaleString();
 
   return {
     id: id.toString(),
     email,
     fullName,
-    firstName: metadata?.first_name as string,
-    lastName: metadata?.last_name as string,
+    metadata,
+    firstName: metadata?.firstName as string,
```
kmitrovv: these will be removed after the parent branch is merged. already updated there to be only fullName.

src/stores/AccountController.ts
```diff
@@ -450,17 +464,29 @@ function useLoginContext<T>(callback: (args: { cleengId: string; cleengSandbox:
   return useConfig((config) => callback({ ...config, customerId: user.id, auth }));
 }
 
+function useAccountContext<T>(callback: (args: { customerId: string; customer: Customer; auth: AuthData }) => T): T {
```
kmitrovv: This is needed because in some flows we don't need to make an API call, we just need the user data which is already in the store. Example: consents, capture..

types/account.d.ts
```diff
@@ -236,19 +247,20 @@ export type Capture = {
...
 // TODO: Convert these all to generic non-cleeng calls
-type Login = (args: LoginArgs) => Promise<{ auth: AuthData; user: Customer }>;
-type Register = CleengRequest<RegisterPayload, AuthData>;
-type GetPublisherConsents = CleengRequest<GetPublisherConsentsPayload, GetPublisherConsentsResponse>;
-type GetCustomerConsents = CleengAuthRequest<GetCustomerConsentsPayload, GetCustomerConsentsResponse>;
+// type Login = CleengRequest<LoginPayload, AuthData>;
+type Login = (args: AuthArgs) => Promise<{ auth: AuthData; user: Customer }>;
+type Register = (args: AuthArgs) => Promise<{ auth: AuthData; user: Customer }>;
+type GetPublisherConsents = ServiceRequest<Config, GetPublisherConsentsResponse>;
+type GetCustomerConsents = ServiceRequest<GetCustomerConsentsPayload, GetCustomerConsentsResponse>;
```
kmitrovv: We should discuss the approach we wanna go with here. I kind of prefer using the ServiceRequest

budzworthy: ServiceRequest is more concise, so that's fine. Just be careful to decouple the `Payload` and `Response` objects from the args that we use in the services, because they define 1-1 the Cleeng api contract at the moment and we don't want to break that. I think it would also be nice to change the services to just throw errors instead of returning an array of errors in `ApiResponse` which we then just rethrow anyhow.

src/containers/AccountModal/forms/Registration.tsx
```diff
@@ -54,10 +54,13 @@ const Registration = () => {
...
+        if (errorMessage.includes('already exists')) {
```
budzworthy: Probably safer to check more verbosely:
```suggestion
        if (errorMessage.includes('customer already exists.') || errorMessage.includes('Account already exists.')) {
```

src/i18n/locales/en_US/account.json
```diff
@@ -122,7 +122,8 @@
     "sign_up": "Sign up",
     "user_exists": "There is already a user with this email address",
     "view_password": "View password",
-    "email_updates": "Yes, I want to receive {{siteName}} updates by email."
+    "email_updates": "Yes, I want to receive {{siteName}} updates by email.",
+    "failed_to_create": "Unable to register at the moment."
```
budzworthy:
```suggestion
    "failed_to_create": "An error occurred. Please try again later."
```

src/services/cleeng.account.service.ts
```diff
@@ -43,6 +48,31 @@ export const login: Login = async ({ config, email, password }) => {
   };
 };
 
+export const register: Register = async ({ config, email, password }) => {
+  const localesResponse = await getLocales(!!config.integrations.cleeng?.useSandbox);
+
+  if (localesResponse.errors.length > 0) throw new Error(localesResponse.errors[0]);
```
budzworthy: I suggest making a helper method for this pattern since it's already used several times.

src/services/cleeng.account.service.ts
```diff
@@ -56,24 +86,61 @@ export async function getUser({ config, auth }: { config: Config; auth: AuthData
...
+  return response;
 };
 
-export const fetchPublisherConsents: GetPublisherConsents = async (payload, sandbox) => {
-  return get(sandbox, `/publishers/${payload.publisherId}/consents`);
+export const updateCustomerConsents: UpdateCustomerConsents = async (payload) => {
+  const { config, customer, jwt } = payload;
+  const { cleeng } = config.integrations;
+
+  const _payload = {
```
dbudzins, line 119: We don't use the underscore prefix on local variables.

types/account.d.ts
```diff
@@ -70,6 +70,11 @@ export type RegisterPayload = PayloadWithIPOverride & {
   externalData?: string;
 };
 
+export type RegisterArgs = {
+  config: Config;
+  user: RegisterPayload;
+};
+
 export type CaptureFirstNameLastName = {
```
budzworthy, line 91: Let's move all these payload types into the cleeng.d.ts file and avoid manipulating them.

src/services/inplayer.account.service.ts
```diff
@@ -25,14 +39,36 @@ export const login: Login = async ({ config, email, password }) => {
...
   }
 };
 
+export const register: Register = async ({ config, email, password }) => {
+  try {
+    const { data } = await InPlayer.Account.signUpV2({
+      email,
+      password,
+      passwordConfirmation: password,
+      fullName: 'New User',
```
budzworthy: Can this be blank instead?

src/stores/AccountController.ts
```diff
@@ -176,34 +187,27 @@ export const logout = async () => {
...
+    const { auth, user } = await accountService.register({ config, email, password });
+    useAccountStore.setState({
+      auth,
+      user,
+    });
 
-    await updatePersonalShelves();
+    // @todo statement will be removed once the fav and history are done on InPlayer side
+    if (auth.refreshToken) {
+      await updatePersonalShelves();
```
budzworthy, line 218: You could create a stub updateShelves method on the inplayer service instead of this sort of hard to understand if condition.

src/stores/AccountController.ts
```diff
@@ -176,34 +187,27 @@ export const logout = async () => {
...
-        country: localesResponse.responseData.country,
-        currency: localesResponse.responseData.currency,
-        publisherId: cleengId,
-      },
-      cleengSandbox,
-    );
-
-    if (responseRegister.errors.length) throw new Error(responseRegister.errors[0]);
-
-    await getAccount(responseRegister.responseData);
```
budzworthy, line 203: I think there are some pieces missing by removing the getAccount (i.e. the stuff in afterLogin.) Most of it probably doesn't matter for the register flow (there shouldn't be any subscription or shelf data) but the Consents stuff is probably needed.

src/stores/AccountController.ts
budzworthy: I would say just add Customer to the useLoginContext function callback. Otherwise these 2 are almost identical.

src/stores/AccountController.ts
```diff
@@ -229,67 +233,81 @@ export const updatePersonalShelves = async () => {
...
+      const response = await accountService.updateCaptureAnswers({ customer, ...capture }, sandbox, auth.jwt);
 
-    if (response.errors.length > 0) throw new Error(response.errors[0]);
+      if (response.errors.length > 0) throw new Error(response.errors[0]);
 
-    // @todo why is this needed?
-    await getAccount(auth);
+      // if no refresh token present (InPlayer config), update account store
+      // otherwise fetch account
+      if (!auth.refreshToken) {
```
budzworthy: If you add the query from getAccount (service.getUser) into the cleeng implementation of updateCaptureAnswers then you don't need this condition here, you can just always call afterLogin.

src/stores/AccountController.ts
```diff
@@ -229,67 +233,81 @@ export const updatePersonalShelves = async () => {
...
         consents: customerConsents,
-      },
-      cleengSandbox,
-      jwt,
-    );
+      });
 
-    await getCustomerConsents();
+      // if user does not have refresh token (InPlayer user) the response data is user object
+      if (response?.responseData && !refreshToken) {
```
budzworthy: This is a kind of messy execution path.  There are multiple callback helpers nested both here and 4 levels deep in the call to getCustomerConsents below.  You're also passing data between functions using the store, which is valid, but hard to follow.

I propose instead we modify the cleeng updateCustomerConsents to include the getCustomerConsents call like what you did for the other calls in the register flow.  Then the generic updateCustomerConsents can just return the customer consents as well.

src/stores/AccountController.ts
```diff
@@ -229,67 +233,81 @@ export const updatePersonalShelves = async () => {
...
+
+      return response;
+    });
   });
 };
 
-export async function getCustomerConsents() {
-  return await useLoginContext(async ({ cleengSandbox, customerId, auth: { jwt } }) => {
-    const response = await cleengAccountService.fetchCustomerConsents({ customerId }, cleengSandbox, jwt);
+export async function getCustomerConsents(): Promise<ServiceResponse<GetCustomerConsentsResponse>> {
```
budzworthy: If you make the change to updateConsents above, then you can probably add the consents into the return for accountService.getUser.  For inplayer you can parse the consent property and for cleeng you can add a call to getCustomerConsents.  Then you have less nesting of calls and less worry about passing around store data and faking out the method for inplayer.  It's probably more understandable to have the cleeng methods doing several api calls and combining data than it is to understand the faked out method.

src/services/inplayer.account.service.ts
```diff
@@ -44,36 +80,170 @@ export const logout = async () => {
...
+};
+
+export const getPublisherConsents: GetPublisherConsents = async (config) => {
+  try {
+    const { inplayer } = config.integrations;
+    const { data } = await InPlayer.Account.getRegisterFields(inplayer?.clientId || '');
+
+    // @ts-ignore
+    // wrong data type from InPlayer SDK (will be updated in the SDK)
+    const result: Array<Consent> = data?.collection
```
budzworthy: Maybe we should use a convention here to allow both consent and capture questions to be added?  Maybe fields that are consents could be named `consent.<something>` and then all other fields will be considered questions?

src/services/inplayer.account.service.ts
```diff
@@ -44,36 +80,170 @@ export const logout = async () => {
...
+
+export const getCaptureStatus: GetCaptureStatus = async ({ customer }) => {
+  return {
+    errors: [],
+    responseData: {
+      isCaptureEnabled: true,
+      shouldCaptureBeDisplayed: true,
+      settings: [
+        {
+          answer: {
```
budzworthy, line 180: It would be nice to add the other register questions that aren't treated as consents here

src/services/cleeng.account.service.ts
kmitrovv: Okay, makes sense.

src/services/inplayer.account.service.ts
kmitrovv: No actually.. We have a backend validation for at least 3 letters i think.. This gets updated on the next step anyways.. Maybe replace it with something else instead of New User?

src/services/inplayer.account.service.ts
kmitrovv: Hmm.. Do we care about `Do you like this form`? I don't see any other question than that.. Or you are talking about something else?

src/stores/AccountController.ts
kmitrovv: Yes, that's what is meant by the todo. The next thing that I am working is the fav and history, so until it's finished this condition will be here because otherwise it throws an error on inplayer side because of cleengId missing.

src/services/inplayer.account.service.ts
kmitrovv: Actually that is the case. This is the method that just retrieves them in the same format as cleeng's ones. We are storing them in our account metadata as : `consents: {jsonstring}`.
For capture we don't need anything special since only `firstName` and `lastName` are stored in the metadata.

So we have something like:

`customer:{
metadata:{
consents: {...},
firstName: ,
lastName:,
}
}`

src/stores/AccountController.ts
kmitrovv: Yeah, nice catch. Somehow it got forgotten. I will just call afterLogin there as it was, the only difference is now it wont make a call to fetchAccount, since the data is already there.

src/services/inplayer.account.service.ts
budzworthy: Oh, that makes it difficult. Maybe we could use the email address then at least?  So it's not something that seems like a bug?  Longer term we can either look at changing the backend or changing the web app flow to not actually make the register call until the second step of the form where name is entered.

src/stores/AccountController.ts
budzworthy: OK, I just thought there might be a cleaner way to stub it so you didn't have to come back to the todo later.  This is temporary, so I won't worry about it.

src/services/inplayer.account.service.ts
budzworthy: `Do you like this form` was just silly test data.  I do think we want to maintain the functionality for companies to add custom questions to the registration flow.

src/stores/AccountController.ts
kmitrovv: Not sure what you mean.. All the functions for consents are in the same flow as the previous functions. This condition here is needed so that the accountStore gets updated after updating consents if it's inplayer config, because we store the consents in the account metadata.

budzworthy: So inside of updateConsents we have:

- useAccountContext
  - withAccountService
    - accountService.updateCustomerConsents
    - useAccountStore.setState(user)
    - getCustomerConsents
      - useAccountContext which reads the user set 2 lines above
        - withAccountService
          - accountService.getCustomerConsents
          - useAccountStore.setState(customerConsents)
          
So we're calling the wrappers multiple times, which won't be terrible for performance, but does make it hard to follow exactly what is happening, especially with tricks like `setState(user)` updating the store which is then read with the second `useAccountContext` which passed it to `getCustomerConsents` which is a dummy API call that just grabs data off of the user object that we've passed around.  It's even more complex because the setState(user) only happens for the inplayer service based off checking the refresh token.

What is a better flow is to do this:
- accountService.updateCustomerConsents returns `customerConsents`
  - Inplayer service can already do this.  We don't need to include the other `user` data.
  - Cleeng service will just also call `getCustomerConsents` inside of the `updateCustomerConsents` method
- We will also have to update `getUser` to return user and customerConsents
  - Again, inplayer service can already do this
  - Cleeng can again just add a call to `getCustomerConsents` inside of the `getUser` method
- The above change will propagate customerConsents to also be able to be returned from `register` and `login`
- We can add customerConsents to the setState in afterLogin and remove getCustomerConsents from afterLogin
- getCustomerConsents can be removed from the accountController
- getCustomerConsents dummy method can be removed from inplayer and cleeng will only be private (not exported) for cleeng

src/stores/AccountController.ts
kmitrovv: Well, the `useLoginContext` uses `useConfig` where cleeng stuff exists and is used in a lot of places. If we try to modify the `useConfig` it will result in messy solution I think.

src/services/inplayer.account.service.ts
kmitrovv: Well that's okay but for that to happen we would want to create separate endpoints as cleeng. I don't think this is something important for the MVP. The hardcoded captures and saving in metadata should be okay for first and last name at the moment, but other than that, we shouldn't make it messier.. Maybe leave it for some next phase where we can define better and create separate endpoints.

src/stores/AccountController.ts
budzworthy: OK this is hopefully all moderately temporary, so I won't worry about it.

src/services/inplayer.account.service.ts
kmitrovv: Yes, we can do that.

src/stores/AccountController.ts
kmitrovv: Yes, it requires refactoring but it makes sense. Will do it.

types/account.d.ts
kmitrovv: Moving these types in the cleeng.d.ts somehow interferes the request types that are defined there and requires changing through the entire code.

src/stores/AccountController.ts
```diff
@@ -100,23 +110,29 @@ export const initializeAccount = async () => {
...
+      throw new Error('Email update not supported');
+    }
 
-  const response = await cleengAccountService.updateCustomer({ ...values, id: user.id.toString() }, cleengSandbox, auth.jwt);
+    if (!auth || !user) throw new Error('no auth');
 
-  if (!response) {
-    return { errors: Array.of('Unknown error') };
-  }
+    const response = await accountService.updateCustomer({ ...values, id: user.id.toString() }, sandbox, auth.jwt);
```
budzworthy, line 132: Isn't this going to end up erasing the consents in the user metadata in the inplayer service?  Consents are not passed in here and are not populated in the service.

types/cleeng.d.ts
```diff
@@ -2,7 +2,9 @@ interface ApiResponse {
   errors: string[];
 }
 type ServiceResponse<R> = { responseData: R } & ApiResponse;
+type ServiceRequest<P, R> = (payload: P) => Promise<ServiceResponse<R>>;
```
budzworthy: I think these should all be Cleeng... names, right?  There are duplicates between this and account.d.ts and some are unused.

src/stores/AccountController.ts
kmitrovv: There is a check in the inplayer service for that not to happen:

image

types/cleeng.d.ts
kmitrovv: Yes, these are extra and should be removed.

src/stores/AccountController.ts
budzworthy: But I thought updates to the user object were all or nothing? Is metadata treated differently?
https://github.com/jwplayer/ott-web-app/pull/199#discussion_r1039555389

kmitrovv: Yes, there is an exception for the metadata, even though it's a put method it does not remove the values which are not sent.
Only the fullName is required to be sent.

types/cleeng.d.ts
dbudzins: OK, then all these changes to support setting metadata through the service.updateCustomer should be rolled back.

This means the only 'required' parameter is fullname, right? Anything else will be part of the metadata.

kiremitrov123: I guess you meant to write this above on top of the other comment, but the metadata saving should remain in the update account because when we update full name we need to update the metadata for `first_name` and `surname` since this is how the metadata looks if you split the inplayer register fields from InPlayer dashboard.

src/stores/AccountController.ts
budzworthy: OK, then all these changes to support setting metadata through the service.updateCustomer should be rolled back.

This means the only 'required' parameter is fullname, right? Anything else will be part of the metadata.

budzworthy: > I guess you meant to write this above on top of the other comment, but the metadata saving should remain in the update account because when we update full name we need to update the metadata for `first_name` and `surname` since this is how the metadata looks if you split the inplayer register fields from InPlayer dashboard.

Yes sorry, wrong text field.

I understand we need to set first_name and surname, but we can exclude the consents on the updateCustomer calls.  What we want is to include first_name and surname on the updateConsents calls, but that doesn't require that we include consents on the updateCustomer calls.

kmitrovv: consents are not always included in the `updateCustomer` call, they are included only if they are present in the values. They are only present in the values if the `updateCustomer` is called from `updateConsents`.
`...(values?.metadata?.consents && { consents: JSON.stringify(values.metadata?.consents) })`.

`updateConsents` is not making direct api call, but calls `updateCustomer`.

src/components/Account/Account.tsx
```diff
@@ -225,7 +235,7 @@ const Account = ({ panelClassName, panelHeaderClassName }: Props): JSX.Element =
         formSection({
           label: t('account.terms_and_tracking'),
           saveButton: t('account.update_consents'),
-          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
+          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)) as Promise<never>,
```
budzworthy: Same effect, without casting
```suggestion
          onSubmit: async (values) => {
            await updateConsents(formatConsentsFromValues(publisherConsents, values))
          },
```

src/components/Account/Account.tsx
```diff
@@ -73,10 +74,14 @@ const Account = ({ panelClassName, panelHeaderClassName }: Props): JSX.Element =
 
   function translateErrors(errors?: string[]): FormErrors {
     const formErrors: FormErrors = {};
-
     // Some errors are combined in a single CSV string instead of one string per error
     errors
-      ?.flatMap((e) => e.split(','))
+      ?.flatMap((e) => {
+        if (typeof e === 'object') {
```
budzworthy: `errors` is a string array so this should never happen.  Probably handling and converting the error type to string needs to happen in the services for this method.

src/i18n/locales/en_US/user.json
```diff
@@ -14,7 +14,8 @@
       "invalid_password": "Password incorrect!",
       "last_name_too_long": "Please limit Last Name to 50 characters or fewer.",
       "please_enter_valid_email": "Please enter a valid email address.",
-      "unknown_error": "An unknown error occurred"
+      "unknown_error": "An unknown error occurred",
```
budzworthy: Please run `yarn i18next` which should parse the translation strings that are used and make sure they're added to all of the language and definition files.

src/stores/AccountController.ts
budzworthy: My point is not that they are included in every api call, but they are added to the update customer method when they don't need to be. This is going to be really hard for other developers to understand and is likely to lead to bugs.  It also forces us to couple 2 actions that don't need to be coupled (in addition to forcing more rework in cleeng and controller code.)

types/account.d.ts
```diff
@@ -158,10 +168,6 @@ export type UpdateCustomerConsentsPayload = {
   consents: CustomerConsent[];
 };
 
-export type RefreshTokenPayload = {
```
budzworthy, line 161: Why is this deleted?

src/stores/AccountController.ts
kmitrovv: And how will the metadata get updated if we don't add it in updateCustomer? The updateConsents does not make direct call, it goes through updateCustomer, it will still have to process the metadata.

types/account.d.ts
kmitrovv: Ugh, I guess somehow when was trying to move the types.. will restore it

src/stores/AccountController.ts
budzworthy: That's the coupling that shouldn't be happening.  updateCustomer and updateConsents can both call `InPlayer.Account.updateAccount`.  The common logic can even be put into a shared method, but by requiring the updateCustomer method on the account services to also accept consents, you're forcing changes onto the cleeng service that don't need to happen.

kmitrovv: That's another approach and it's okay, but still don't see how this approach is affecting cleeng?

budzworthy: Because the services have to have the same method interfaces, this means that the update customer method on Cleeng also has to accept and handle consent data.

kmitrovv: Well yes, but that's the case with other methods as well, the flows are just not the same and we want to use generic types and methods. 
Cleeng service for updating customer is not handling consents, it's an optional prop, it works same as before.

types/account.d.ts
kmitrovv: Will leave this for later stage.

src/components/Account/Account.tsx
darkoatanasovski: yes, nice catch, but this can happen. For example [here](https://github.com/jwplayer/ott-web-app/blob/2aae0ae3f98a6cad1831880258e95ed6857fbcd9/src/stores/AccountController.ts#L119), if somehow the controller throws an error, the format of the error will be Object.
In this case, the error will be `Uncaught TypeError: e.split is not a function`.

How to reproduce this:
In the [User.tsx](https://github.com/jwplayer/ott-web-app/blob/2aae0ae3f98a6cad1831880258e95ed6857fbcd9/src/pages/User/User.tsx#L101), set `canUpdateEmail` to true and try to update the email.
The controller will throw an error:
```
    if (Object.prototype.hasOwnProperty.call(values, 'email') && !canUpdateEmail) {
      throw new Error('Email update not supported');
    }
```

But, yes, I agree that this is a good starting point for post mvp refactoring

src/stores/AccountController.ts
budzworthy: For maintainability sake, we should err on the side of not expanding methods unless we have to, and in this case, we could have used the metadata partial updates and avoided a lot of changes in the first place. It's done now, so I'm going to approve it, but this will likely get refactored to simplify the methods and flows.

src/components/Account/Account.tsx
budzworthy: Ah, thanks for the repro. I think I see where this needs to be fixed:
https://github.com/jwplayer/ott-web-app/blob/96f6da3bae24a7ebe5f45ff9ec958d0e470c7a90/src/components/Form/FormSection.tsx#L104

Probably something like this:
```
response = { errors: Array.of(error instanceof Error ? error.message : error as string) };
```

darkoatanasovski: OK, fixed :)

src/components/Account/Account.tsx
```diff
@@ -225,7 +232,7 @@ const Account = ({ panelClassName, panelHeaderClassName }: Props): JSX.Element =
         formSection({
           label: t('account.terms_and_tracking'),
           saveButton: t('account.update_consents'),
-          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
+          onSubmit: async (values) => await updateConsents(formatConsentsFromValues(publisherConsents, values)),
```
budzworthy:
```suggestion
          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
```

budzworthy (approved):

PR #206, opened 2022-12-10
chore: implemented favorites and watchlist history

darkoatanasovski:
## Description

Implementation of favorites and watchlist history

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/components/Account/Account.tsx
```diff
@@ -77,7 +77,12 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
     const formErrors: FormErrors = {};
     // Some errors are combined in a single CSV string instead of one string per error
     errors
-      ?.flatMap((e) => e.split(','))
+      ?.flatMap((e) => {
```
budzworthy: I think this is an artifact from merging and branching. It shouldn't be necessary anymore or is that not true?

src/components/Account/Account.tsx
```diff
@@ -232,7 +237,9 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
         formSection({
           label: t('account.terms_and_tracking'),
           saveButton: t('account.update_consents'),
-          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
+          onSubmit: async (values) => {
+            await updateConsents(formatConsentsFromValues(publisherConsents, values));
+          },
```
budzworthy: Another merge / branch artifact:
```suggestion
          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
```

src/components/Form/FormSection.tsx
```diff
@@ -24,7 +24,7 @@ export interface FormSectionProps<TData extends GenericFormValues, TErrors> {
   saveButton?: string;
   cancelButton?: string;
   canSave?: (values: TData) => boolean;
-  onSubmit?: (values: TData) => Promise<{ errors?: string[] }>;
+  onSubmit?: (values: TData) => Promise<{ errors?: string[] }> | Promise<void> | void;
```
budzworthy: We can't add this back in. The voids can too easily cause problems with dropped responses (there was a bug where errors from the email update weren't getting handled)
```suggestion
  onSubmit?: (values: TData) => Promise<{ errors?: string[] }>;
```

src/components/Form/FormSection.tsx
```diff
@@ -93,15 +93,15 @@ export function FormSection<TData extends GenericFormValues>({
       event && event.preventDefault();
 
       if (onSubmit) {
-        let response: { errors?: string[] };
+        let response: { errors?: string[] } | void;
```
budzworthy:
```suggestion
        let response: { errors?: string[] };
```

src/components/Account/Account.tsx
darkoatanasovski: yes, but fails on lint :/

src/components/Form/FormSection.tsx
```diff
@@ -93,15 +93,15 @@ export function FormSection<TData extends GenericFormValues>({
...
 
         try {
           setFormState((s) => {
             return { ...s, isBusy: true };
           });
           response = await onSubmit(values);
         } catch (error: unknown) {
-          response = { errors: Array.of(error instanceof Error ? error.message : (error as string)) };
+          response = { errors: Array.of(error as string) };
```
budzworthy: OK, I think all the changes in this file should be rolled back. Looks like it's all old stuff from old branch.

src/components/Account/Account.tsx
darkoatanasovski: ahh yes, you are right

darkoatanasovski: OK, done :)

src/components/Form/FormSection.tsx
darkoatanasovski: also done

src/components/Form/FormSection.tsx
darkoatanasovski: yes, rolled back

src/components/Account/Account.tsx
darkoatanasovski: done

src/services/inplayer.account.service.ts
```diff
@@ -328,3 +333,75 @@ function parseJson(value: string, fallback = {}) {
 export const canUpdateEmail = false;
 
 export const canChangePasswordWithOldPassword = true;
+
+export const initCustomerExtras = async (): Promise<ExternalData> => {
```
budzworthy: I think it would be advisable to put watch list stuff into its own service. We already have favorites and continue watching controllers, and this would avoid coupling a lot of things like this to the account and user.

src/services/inplayer.account.service.ts
```diff
@@ -328,3 +333,75 @@ function parseJson(value: string, fallback = {}) {
...
+  return {
+    favorites,
+    history,
+  };
+};
+
+export const updatePersonalShelves: UpdatePersonalShelves = async (payload) => {
+  const { favorites, history } = payload.externalData;
+  const user = useAccountStore.getState().user;
+  const currentFavoriteIds = user?.externalData?.favorites?.flatMap((e) => e.mediaid);
```
budzworthy: Do these need to be flatmap or just map?

src/services/inplayer.account.service.ts
```diff
@@ -328,3 +333,75 @@ function parseJson(value: string, fallback = {}) {
...
+  const payloadFavoriteIds = favorites?.flatMap((e) => e.mediaid);
+
+  try {
+    history.forEach(async (history) => {
+      if (user?.externalData?.history?.length) {
```
budzworthy: I don't understand what's happening here. Can you add some commenting (unless we end up refactoring it)

src/services/inplayer.account.service.ts
darkoatanasovski: it can be just `map` I think

src/services/inplayer.account.service.ts
```diff
@@ -328,3 +333,75 @@ function parseJson(value: string, fallback = {}) {
...
+    return processHistoryItem(history);
+  });
+
+  return {
+    favorites,
+    history,
+  };
+};
+
+export const updatePersonalShelves: UpdatePersonalShelves = async (payload) => {
```
budzworthy, line 355: We were talking about a downside to the current cleeng implementation today being that it updates the whole list object instead of doing operations on individual items.  This causes problems when updating from 2 different devices that have different lists cached.  The last write overwrites any intermediate changes from the other device.

Since the Inplayer API supports individual updates, it would be nice to call add/remove/update items directly from the appropriate methods in the controllers.

src/stores/AccountController.ts
```diff
@@ -161,6 +161,11 @@ export const getAccount = async (auth: AuthData) => {
   await withAccountService(async ({ accountService, config, accessModel }) => {
     const response = await accountService.getUser({ config, auth });
 
+    const externalData = await accountService.initCustomerExtras();
```
budzworthy: If we don't end up creating favorites / history controllers, then I would recommend we add this logic into the Inplayer getUser method instead of adding another service method that is only really used by Inplayer anyhow.

budzworthy (changes requested):
I think the favorites and history endpoints will work better in a favorites / history service. This way we can better support individual item operations and take advantage of the flexibility of the new API.

src/services/inplayer.account.service.ts
darkoatanasovski: yes, this makes sense and I already fixed that as you said

src/services/inplayer.account.service.ts
darkoatanasovski: done

src/services/inplayer.account.service.ts
darkoatanasovski: this is a universal and InPlayer-specific method for converting an array of favorites/watch-history items into separate requests to the InPlayer API.

Because we do not support bulk inserts( like cleeng ), in this method, we calculate the delta( the difference between the local arrays and remote arrays) of the arrays from the local storage/account store.

Every time when the array is changed( when the users click on the favorite button, the current flow), we send separate requests for each favorite/watch-history item from the 'delta' to the InPlayer's API.

For this part, for the watch history, we always send a patch request to update/create the watch-history item for the user.

This method is also used to sync the data from the local storage when the user is newly registered or logged in.

src/stores/AccountController.ts
darkoatanasovski: good point.

PR #208, opened 2022-12-13
InPlayer subscriptions

kmitrovv:
## Description

This PR is a starting point of the subscriptions flow.
All the subscription connected changes will be merged here.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


src/components/Account/Account.tsx
```diff
@@ -232,7 +232,7 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
         formSection({
           label: t('account.terms_and_tracking'),
           saveButton: t('account.update_consents'),
-          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
+          onSubmit: async (values) => await updateConsents(formatConsentsFromValues(publisherConsents, values)),
```
budzworthy:
```suggestion
          onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values)),
```

src/stores/AccountController.ts
```diff
@@ -218,6 +218,7 @@ export const register = async (email: string, password: string) => {
       await updatePersonalShelves();
     }
   });
+  useAccountStore.setState({ loading: true });
```
```suggestion
```

src/services/inplayer.notifications.service.ts
```diff
@@ -0,0 +1,59 @@
...
+  SUBSCRIBE_SUCCESS = 'subscribe.success',
+  SUBSCRIBE_FAILED = 'subscribe.failed',
+  PAYMENT_CARD_SUCCESS = 'payment.card.success',
+  PAYMENT_CARD_FAILED = 'payment.card.failed',
+  PAYMENT_CARD_REQUIRES_ACTION = 'payment.card.requires.action',
+  ACCOUNT_LOGOUT = 'account.logout',
+}
+
+const handleLogout = async () => {
+  await logout();
```
budzworthy: Do we expect these methods to grow in the future?  If not, why not just call `logout` and `reloadActiveSubscription` directly?

kmitrovv: If they are called directly it throws an error saying something like `trying to call before initialise`.... however this file will be updated, more of them are like placeholders just so we have the idea how it will work.

src/services/inplayer.notifications.service.ts
```diff
@@ -0,0 +1,59 @@
...
+  [NotificationsTypes.PAYMENT_CARD_REQUIRES_ACTION]: [handleActionStep],
+  [NotificationsTypes.ACCOUNT_LOGOUT]: [handleLogout],
+};
+
+export const subscribeToNotifications = (uuid: string = '') => {
+  if (!InPlayer.Notifications.isSubscribed()) {
+    InPlayer.subscribe(uuid, {
+      onMessage: function (message) {
+        const notification = JSON.parse(message) as Notification;
+        notifications[notification.type]?.forEach((handler) => {
```
budzworthy: Do we really anticipate needing multiple handlers per message type?

budzworthy (commented):
Having notifications as a service is breaking some of our conventions. Services generally shouldn't reference other services (with the exception of the cleeng service, which should really be renamed to cleeng client.)  Also, services definitely shouldn't call controllers (though controllers can call other controllers.)

I think that it would make sense to add a `subscribeToNotifications` method to the inplayer account service and call it from the AccountController as part of the register and login methods there. The `NotificationsTypes` enum can be part of the global types, but the mapping between InPlayer specific strings and the enum should be encapsulated in the account service, so that the input to `InplayerAccountService.subscribeToNotifications` is either a callback with the enum as an arg or a dictionary of enum values to methods.

src/services/inplayer.notifications.service.ts
kmitrovv: I think we will need it for the sca payment flow.

budzworthy: Can't there just be multiple things done in a single method?  Instead of `[doThis, doThat, doTheOtherThing]`, just have 1 method `doAll() {doThis(); doThat(); doTheOtherThing();}`

budzworthy:
The credit card logos should probably go in `public/images` instead of `images/` in the root. The latter is only used for documentation images at the moment and I don't want to mix up doc images with code images.

budzworthy:
Another file comment that it won't let me comment on. I just noticed that yarn.lock switched from yarn to npm at some point. Was this intentional? If so, let's discuss, otherwise please revert.

src/components/CreditCardCVCField/CreditCardCVCField.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type Props = {
+  onChange?: React.ChangeEventHandler<HTMLInputElement>;
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
+  error?: string;
+  value?: string;
+};
+
+const CreditCardCVCField: React.FC<Props> = ({ value, onChange, error, ...props }: Props) => {
+  const formatCVC: React.ChangeEventHandler<HTMLInputElement> = (e) => {
```
budzworthy, line 13: Should this use useCallback?

src/components/CreditCardExpiryField/CreditCardExpiryField.tsx
```diff
@@ -0,0 +1,52 @@
...
+      label={`Expiry date`}
+      {...props}
+      name="cardExpiry"
+      value={value}
+      error={!!error}
+      helperText={error ? error : null}
+      onChange={formatExpirationDate}
+      type="text"
+      pattern="\d*"
+      placeholder="dd/mm"
```
budzworthy: Should probably be mm/yy

src/components/CreditCardCVCField/__snapshots__/PasswordField.test.tsx.snap
```diff
@@ -0,0 +1,57 @@
+// Vitest Snapshot v1
```
budzworthy: Where did this file come from?

src/components/CreditCardExpiryField/__snapshots__/PasswordField.test.tsx.snap
```diff
@@ -0,0 +1,57 @@
+// Vitest Snapshot v1
```
budzworthy: Where did this file come from?

src/components/CreditCardNumberField/CreditCardNumberField.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { t } = useTranslation('account');
+  const [cardIssuer, setCardIssuer] = useState<JSX.Element | null>(null);
+
+  const formatCreditCardNumber: React.ChangeEventHandler<HTMLInputElement> = (e) => {
+    const issuer = Payment.fns.cardType(e.target?.value);
+    const clearValue = e.target?.value.replace(/\D+/g, '');
+
+    let nextValue;
+    setCardIssuer(
+      cardIssuers?.[issuer] ? (
```
budzworthy, line 38: Might be nice to have a fallback here if issuer not found

src/components/CreditCardCVCField/CreditCardCVCField.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const clearValue = e.target.value.replace(/\D+/g, '');
+    if (onChange) {
+      e.target.value = clearValue.slice(0, 4);
```
budzworthy, line 16: Would be nice to have an input here for max length.  If the card number is already set and we know it's a 3 digit cvc, it can avoid some user errors.

src/components/CreditCardCVCField/CreditCardCVCField.tsx
```diff
@@ -0,0 +1,37 @@
...
+      onChange(e);
+    }
+  };
+  return (
+    <TextField
```
budzworthy, line 21: For this and all the other fields, I suggest also connecting the minlength, maxlength, and format html props to the appropriate values.

src/components/CreditCardNumberField/CreditCardNumberField.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <TextField
+      value={value}
+      name="cardNumber"
+      label={`Card number`}
+      onChange={formatCreditCardNumber}
+      onBlur={onBlur}
+      error={!!error}
+      helperText={error ? error : null}
+      pattern="\d*"
```
budzworthy, line 72: Shouldn't pattern match the whitespace delimitation done in lines 47-56?

src/components/CreditCardExpiryField/CreditCardExpiryField.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <TextField
+      label={`Expiry date`}
+      {...props}
+      name="cardExpiry"
+      value={value}
+      error={!!error}
+      helperText={error ? error : null}
+      onChange={formatExpirationDate}
+      type="text"
+      pattern="\d*"
```
budzworthy: Shouldn't pattern be something like `[0-1][0-9]/[0-9]{2}`?

src/components/CreditCardNumberField/__snapshots__/PasswordField.test.tsx.snap
```diff
@@ -0,0 +1,57 @@
+// Vitest Snapshot v1
```
budzworthy: Again, not sure where the password snapshot came from

src/components/PasswordField/PasswordField.module.scss
```diff
@@ -1,90 +0,0 @@
-@use 'src/styles/variables';
```
budzworthy, line 1: Is this file not needed anymore?

src/components/PaymentForm/PaymentForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import CreditCardNumberField from '../CreditCardNumberField/CreditCardNumberField';
+import TextField from '../TextField/TextField';
+
+import styles from './PaymentForm.module.scss';
+
+import type { GenericFormValues } from '#types/form';
+import type { CardPaymentData } from '#types/checkout';
+
+type Props = {
+  paymentDataForm?: CardPaymentData & GenericFormValues;
```
budzworthy: Can this just be CardPaymentData? What is the GenericFormValues giving you?

src/components/PaymentForm/PaymentForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const PaymentForm: React.FC<Props> = ({ paymentDataForm }) => {
+  return (
+    <div className={styles.paymentForm}>
+      <div>
+        <TextField
+          label="Cardholder name"
```
budzworthy, line 66: Missing translation

src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,52 @@
...
+const useNotifications = () => {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const notification = useNotificationStore((state) => state);
+
+  useEffect(() => {
+    switch (notification.type) {
+      case NotificationsTypes.FAILED:
+      case NotificationsTypes.SUBSCRIBE_FAILED:
+        navigate(
```
budzworthy: Is there a risk here that notifications might be delayed?  Example, I try to make a payment and it fails, but I don't get the failure notification right away.  Then I go back to browsing or finish payment via paypal or something. Then a few minutes later a delayed notification come in and appends new query params to the page?

src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,52 @@
...
+          addQueryParams(window.location.href, {
```
budzworthy: Would be cleaner to use the `useSearchParams` hook to only modify the query params

src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,52 @@
...
+        break;
+      default:
+        break;
+    }
+  }, [notification]);
+
+  return notification;
+};
+
+export default useNotifications;
```
budzworthy: After seeing this whole implementation and thinking a bit more about the architecture, I'm a little bit worried about the global notification mechanism and delayed or mis-ordered notifications.

In most cases notifications probably come relatively quickly and it's a moot point, but in theory they could be delayed or come out of the expected order.

A couple ways we could protect against weird behavior in these cases:

1. Use more state checks before acting on notifications (such as making sure we are still on the subscribing page before following the access granted notification)
2. Explicitly subscribing to certain message types where we actually expect them and unsubscribing when they are no longer expected instead of always listening globally.

Option 1 is probably more maintainable, but option 2 might work well as a wrapper component that could subscribe on mount and unsubscribe on unmount. This is probably low risk for MVP, but worth putting TODO's in at least to say "make sure we are still waiting for X to happen" in each of these code paths.

src/hooks/useService.ts
```diff
@@ -0,0 +1,49 @@
...
+    return callback({
+      accountService: inplayerAccountService,
+      subscriptionService: inplayerSubscriptionService,
+      checkoutService: inplayerCheckoutService,
+      config,
+      accessModel,
+      sandbox: !!inplayer.useSandbox,
+      authProviderId: inplayer?.clientId?.toString(),
+    });
+  } else if (cleeng?.id) {
```
budzworthy: Unnecessary `else` in `else if`

src/hooks/useService.ts
```diff
@@ -0,0 +1,49 @@
+import * as inplayerAccountService from '#src/services/inplayer.account.service';
+import * as inplayerSubscriptionService from '#src/services/inplayer.subscription.service';
+import * as inplayerCheckoutService from '#src/services/inplayer.checkout.service';
+import * as cleengCheckoutService from '#src/services/cleeng.checkout.service';
+import * as cleengSubscriptionService from '#src/services/cleeng.subscription.service';
+import * as cleengAccountService from '#src/services/cleeng.account.service';
+import { useConfigStore } from '#src/stores/ConfigStore';
+import type { AccessModel, Config } from '#types/Config';
+
+function useService<T>(
```
budzworthy, line 14: This works well, but feels a bit heavy handed for the long term. It seems like a sort of a global variable pattern or a reinvention of react context. I guess it's fine as a shortcut for now, but I don't know if this is the right long term way to do this, and definitely would want to minimize just starting to put any data we might need into here (config, access model, etc.)

src/hooks/useCheckAccess.ts
```diff
@@ -0,0 +1,46 @@
...
+  offerId?: string;
+};
+const useCheckAccess = () => {
+  const intervalRef = useRef<number>();
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  const intervalCheckAccess = ({ interval = 3000, iterations = 5, offerId }: intervalCheckAccessPayload) => {
+    const { clientOffers } = useClientIntegration();
+    if (!offerId && clientOffers?.[0]) {
```
budzworthy: Should offer ID be required? I think anywhere we would use this, we should know what offer we were trying to purchase.

src/hooks/useCheckAccess.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+    intervalRef.current = window.setInterval(async () => {
+      const hasAccess = await checkEntitlements(offerId);
+
+      if (hasAccess) {
+        await reloadActiveSubscription();
+        navigate(addQueryParam(location, 'u', 'welcome'));
+      } else if (--iterations === 0) {
+        window.clearInterval(intervalRef.current);
+        navigate(removeQueryParam(location, 'u'));
```
budzworthy: What happens when the iterations run out? It just returns to the page behind without any notification? Should we at least show a message like "Something is taking longer than usual. Please refresh the page in a few minutes..." or something?

src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -44,28 +48,61 @@ const Checkout = () => {
     return offerType === 'svod' ? addQueryParam(location, 'u', 'welcome') : removeQueryParam(location, 'u');
   }, [location, offerType]);
 
-  const couponCodeForm = useForm({ couponCode: '' }, async (values, { setSubmitting, setErrors }) => {
+  const paymentDataForm = useForm(
```
budzworthy: This logic seems like it's in the wrong component. Wouldn't all the payment input stuff be better encapsulated in `PaymentForm`? For the cleeng and paypal cases, this stuff isn't relevant. Cleeng and paypal logic would also go to own components.

src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -184,6 +221,9 @@ const Checkout = () => {
     }
 
     if (paymentMethod?.methodName === 'card') {
+      if (paymentMethod?.provider === 'stripe') {
```
budzworthy, line 212: I think the stripe / adyen terminology for provider here is a bit confusing. Yes, payment provider is stripe or adyen behind the scenes but what we really care about here is the payment form to use, so maybe payment form = `adyen iframe` or `local form` would be more intuitive?

src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -44,28 +48,61 @@ const Checkout = () => {
...
       }
     }
 
     setUpdatingOrder(false);
-    setSubmitting(false);
-  });
+    paymentDataForm.setSubmitting(false);
+  };
+
+  useEffect(() => {
```
budzworthy, line 94: What is this code doing? Can you add a comment please?

src/services/inplayer.subscription.service.ts
```diff
@@ -0,0 +1,167 @@
...
+      variant: card_type,
+      lastCardFourDigits: number,
+      cardExpirationDate: `${zeroFillExpMonth}/${exp_year}`,
+    },
+    active: true,
+  } as PaymentDetail;
+};
+
+// TODO: fix PurchaseDetails type in InPlayer SDK
+const processTransaction = (transaction: InPlayerPurchaseDetails): Transaction => {
```
budzworthy: `process` term is confusing in a payment context. I thought at first this was the main process payment method. Maybe just `formatTransaction` is easier?

src/services/inplayer.checkout.service.ts
```diff
@@ -0,0 +1,225 @@
...
+    cvv: parseInt(cardPaymentPayload.cardCVC),
+    accessFee: order.id,
+    paymentMethod: '1',
+    voucherCode: cardPaymentPayload.couponCode,
+    referrer: window.location.href,
+    returnUrl: `${window.location.href}&u=waiting-for-payment`,
+  };
+
+  try {
+    if (order.offerId[0] === 'S') {
```
budzworthy: This is a standard convention in Inplayer?  If so, please add a comment, ideally with a doc ref.

src/services/inplayer.checkout.service.ts
```diff
@@ -0,0 +1,225 @@
...
+    const response = await InPlayer.Payment.getPayPalParams({
+      origin: payload.successUrl,
+      accessFeeId: payload.order.id,
+      paymentMethod: 2,
+    });
+    //@ts-ignore
+    //TODO fix this type in InPlayer SDK
+    if (response.data?.id) {
+      return {
+        errors: ['Already have an active access'],
```
budzworthy, line 90: Translate

src/hooks/useCheckAccess.ts
```diff
@@ -0,0 +1,46 @@
...
+  interval?: number;
+  iterations?: number;
+  offerId?: string;
+};
+const useCheckAccess = () => {
+  const intervalRef = useRef<number>();
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  const intervalCheckAccess = ({ interval = 3000, iterations = 5, offerId }: intervalCheckAccessPayload) => {
```
budzworthy: useCallback

src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -44,28 +48,61 @@ const Checkout = () => {
     return offerType === 'svod' ? addQueryParam(location, 'u', 'welcome') : removeQueryParam(location, 'u');
   }, [location, offerType]);
 
-  const couponCodeForm = useForm({ couponCode: '' }, async (values, { setSubmitting, setErrors }) => {
+  const paymentDataForm = useForm(
+    { couponCode: '', cardholderName: '', cardNumber: '', cardExpiry: '', cardCVC: '', cardExpMonth: '', cardExpYear: '' },
```
budzworthy: It might make sense to keep coupon code form and payment submit form separate

src/hooks/useCheckAccess.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { addQueryParam, removeQueryParam } from '#src/utils/location';
+import { checkEntitlements, reloadActiveSubscription } from '#src/stores/AccountController';
+
+type intervalCheckAccessPayload = {
+  interval?: number;
+  iterations?: number;
+  offerId?: string;
+};
+const useCheckAccess = () => {
```
budzworthy, line 15: This is a potential place to also add and remove the notification subscriptions so that you don't act on notifications that come delayed or out of order.

test/fixtures/monthlyOffer.json
```diff
@@ -1,4 +1,5 @@
 {
+  "id": 1,
```
budzworthy, line 2: It's always a good habit to use random-ish numbers for things like test ID's so it's less likely to get passing values just by accident. Especially make sure all the offers have different ID's so if the wrong property is used somewhere, the tests can catch it.

types/checkout.d.ts
```diff
@@ -1,6 +1,7 @@
 import type { PayloadWithIPOverride } from '#types/account';
 
 export type Offer = {
+  id: number | null;
```
budzworthy, line 4: Why can't we just use the existing offerId property with the number as text?

src/stores/AccountController.ts
```diff
@@ -428,38 +453,39 @@ export async function getMediaItems(watchlistId: string | undefined | null, medi
...
+    // subscribe to listen to web socket notifications
+    await accountService.subscribeToNotifications(user.uuid, (message) => {
+      const notification = JSON.parse(message);
 
-async function getActivePayment({ cleengSandbox, customerId, jwt }: { cleengSandbox: boolean; customerId: string; jwt: string }) {
-  const response = await getPaymentDetails({ customerId }, cleengSandbox, jwt);
+      notifications[notification.type as NotificationsTypes]?.();
 
-  if (response.errors.length > 0) return null;
+      if (notification) {
```
budzworthy: Is there any need for storing notifications in a store?  The store seems to only hold 1 value anyhow and I don't see that it's used anywhere.

src/stores/NotificationStore.ts
```diff
@@ -0,0 +1,12 @@
+import { createStore } from '#src/stores/utils';
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
budzworthy: What's the intended use for this store?

src/stores/AccountController.ts
```diff
@@ -359,57 +369,72 @@ export const changePasswordWithToken = async (customerEmail: string, newPassword
...
-        window.setTimeout(() => {
-          reloadActiveSubscription().finally(resolve);
-        }, delay);
+      const [activeSubscription, transactions, activePayment] = await Promise.all([
+        subscriptionService.getActiveSubscription({ sandbox, customerId, jwt, config }),
+        subscriptionService.getAllTransactions({ sandbox, customerId, jwt }),
+        subscriptionService.getActivePayment({ sandbox, customerId, jwt }),
+      ]);
+      // this invalidates all entitlements caches which makes the useEntitlement hook to verify the entitlements.
+      await queryClient.invalidateQueries('entitlements');
```
budzworthy, line 429: This will be slightly easier to follow if `'entitlements'` is a string constant shared here and where the queries are

src/stores/AccountController.ts
```diff
@@ -428,38 +453,39 @@ export async function getMediaItems(watchlistId: string | undefined | null, medi
...
 
-  return response.responseData.items;
-}
+    // subscribe to listen to web socket notifications
+    await accountService.subscribeToNotifications(user.uuid, (message) => {
+      const notification = JSON.parse(message);
 
-async function getActivePayment({ cleengSandbox, customerId, jwt }: { cleengSandbox: boolean; customerId: string; jwt: string }) {
-  const response = await getPaymentDetails({ customerId }, cleengSandbox, jwt);
+      notifications[notification.type as NotificationsTypes]?.();
```
budzworthy: It's hard to follow what's happening here. I think the lookup is overkill.

src/stores/AccountController.ts
```diff
@@ -428,38 +453,39 @@ export async function getMediaItems(watchlistId: string | undefined | null, medi
...
+    useAccountStore.setState({
+      auth,
+      user,
+      customerConsents,
+    });
 
-  return response.responseData.items;
-}
+    // subscribe to listen to web socket notifications
+    await accountService.subscribeToNotifications(user.uuid, (message) => {
```
budzworthy: I'm having a hard time understanding why 1/2 of the notifications are handled here and 1/2 are handled in the useNotification hook.

src/stores/CheckoutController.ts
```diff
@@ -1,152 +1,165 @@
...
 
-export const createOrder = async (offerId: string, paymentMethodId?: number) => {
-  const { cleengSandbox, cleengId } = useConfigStore.getState().getCleengData();
+export const createOrder = async (offer: Offer, paymentMethodId?: number): Promise<unknown> => {
+  return await useAccount(async ({ customer, auth: { jwt } }) => {
+    return await useService(async ({ checkoutService, accountService, sandbox, authProviderId }) => {
+      if (!authProviderId) throw new Error('auth provider is not configured');
 
-  const { user, auth } = useAccountStore.getState();
+      const localesResponse = await accountService.getLocales(sandbox);
```
budzworthy: Can Locales be hidden inside of the checkout service? Yes in Cleeng it means a service would have to call another service or there would have to be a duplicated method, but that's probably better than having to have faked out methods in INP

src/stores/CheckoutController.ts
```diff
@@ -1,152 +1,165 @@
...
+      return response.responseData;
+    });
+  });
 };
 
-export const adyenPayment = async (paymentMethod: AdyenPaymentMethod) => {
-  const { cleengSandbox, cleengId } = useConfigStore.getState().getCleengData();
-  const { user, auth } = useAccountStore.getState();
-  const { order } = useCheckoutStore.getState();
+export const cardPayment = async (cardPaymentPayload: CardPaymentData): Promise<unknown> => {
```
budzworthy: Instead of calling these adyen and card payment, I think directPostCardPayment and iFrameCardPayment will make it more understandable. Unless it's possible to combine these 2 into a single path that takes a param of card details or reference details.

budzworthy (changes requested):
It's a lot of comments, but the good news is that many are small naming or code readability things.

The biggest concern I had at the moment is with the notifications architecture. It feels very global and a little too much of a firehose. I think we could create a strategy that is more targeted to only listen for notifications when and where you expect them.

src/components/CreditCardCVCField/CreditCardCVCField.tsx
kmitrovv: the effect is the same with useCallback on the field level. Maybe we should consider improving the payment form component, to somehow memoize each part individually, unless the corresponding dependency updates.
But, nothing concerning for now :)

src/components/CreditCardCVCField/__snapshots__/PasswordField.test.tsx.snap
kmitrovv: not needed file. deleted

src/components/CreditCardExpiryField/__snapshots__/PasswordField.test.tsx.snap
kmitrovv: not needed file. deleted

src/components/CreditCardNumberField/CreditCardNumberField.tsx
kmitrovv: if the issuer is not found, it returns null, which is the case of an invalid credit card. Not sure if we should set the fallback to something else?

src/stores/AccountController.ts
kmitrovv: This is an existing code, only the order in the function is updated.

src/stores/CheckoutController.ts
kmitrovv: updated.

src/components/CreditCardCVCField/CreditCardCVCField.tsx
```diff
@@ -28,7 +28,7 @@ const CreditCardCVCField: React.FC<Props> = ({ value, onChange, error, ...props
       value={value}
       onChange={formatCVC}
       pattern="\d*"
-      placeholder="cvc/cvv"
+      placeholder="123"
```
budzworthy: What do you think about just using `###` as the placeholder?

src/i18n/locales/en_US/account.json
```diff
@@ -13,8 +13,9 @@
     "coupon_applied": "Your coupon code has been applied",
     "coupon_discount": "Coupon discount",
     "coupon_not_valid": "Coupon code is not valid",
-    "credit_card": "Credit Card",
-    "credit_card_number": "Credit card number",
+    "credit_card": "Credit card",
+    "credit_card_name": "Credit card name",
+    "credit_card_number": "1234 5678 9012 3456",
```
budzworthy: Does this need translation if it's just numbers?

budzworthy (commented):
Nice set of changes in the last commit!

src/i18n/locales/en_US/account.json
kmitrovv: yeah, it should be hardcoded I guess.

src/components/CreditCardCVCField/CreditCardCVCField.tsx
darkoatanasovski: also, we can validate the `CVC` number:
e.g. only `Amex` card type can have 4 digits `CVC`
```
Payment.fns.validateCardCVC('123'); //=> true
Payment.fns.validateCardCVC('123', 'amex'); //=> true
Payment.fns.validateCardCVC('1234', 'amex'); //=> true
Payment.fns.validateCardCVC('12344'); //=> false
```

kmitrovv: We can do that.

src/components/CreditCardNumberField/CreditCardNumberField.tsx
darkoatanasovski: This is for mobile users. This attribute will cause the numeric keyboard to be displayed on mobiles. Also, we can set `novalidate` attribute to the form

src/stores/AccountController.ts
kmitrovv: updated

src/stores/AccountController.ts
kmitrovv: updated.

src/stores/CheckoutController.ts
kmitrovv: updated.

src/hooks/useNotifications.ts
kmitrovv: updated

src/hooks/useCheckAccess.ts
kmitrovv: We should show a failed payment screen I guess, if there is still no access after the n-th time.

src/components/CreditCardCVCField/CreditCardCVCField.tsx
kmitrovv: yes, we should do that

src/hooks/useNotifications.ts
kmitrovv: It's a possibility, it can be improved in the future on backend side as well.

src/stores/NotificationStore.ts
kmitrovv: deleted

src/stores/AccountController.ts
kmitrovv: updated

src/hooks/useCheckAccess.ts
budzworthy: Ya, probably something that implies it may have succeeded or may not. We want to avoid accidental double payments.

src/services/cleeng.checkout.service.ts
```diff
@@ -16,16 +18,13 @@ import { getOverrideIP } from '#src/utils/common';
 
 export const getOffers: GetOffers = async (payload, sandbox) => {
   const offers: Offer[] = [];
-  payload.offerIds.forEach(async (offerId) => {
-    const response = await getOffer({ offerId: offerId as string }, sandbox);
-
+  for (let i = 0; i < payload.offerIds.length; i++) {
```
budzworthy: What was the reason for this change?

darkoatanasovski: because `Cleeng` has a separate endpoint for every offer( access fee in `InPlayer`), in the other hand, `InPlayer` has a single endpoint for all access fees( offers in `Cleeng`). and that is why we created a new method for fetching `Cleeng` offers in a single method( to simulate a single endpoint request )

darkoatanasovski: ahh if you mean for "`forEach` / `for`" change, this is because `forEach` is async and ignores the `await`. Maybe we can use `forEach`, but at the end, we have to wait for responses (e.g. `Promise.all(...)` )

budzworthy: Oh gotcha. Indeed, let's try to use Promise.all so that they can be in parallel then.

darkoatanasovski: yes, I agree, it will be faster if they run in parallel

src/stores/AccountController.ts
```diff
@@ -69,10 +66,15 @@ export const handleVisibilityChange = () => {
 };
 
 export const initializeAccount = async () => {
-  await withAccountService(async ({ accountService, config }) => {
+  await useService(async ({ accountService, config }) => {
+    if (!accountService) {
```
budzworthy: I don't understand this if case. Can you add a comment?

src/stores/CheckoutController.ts
```diff
@@ -1,152 +1,174 @@
...
-  if (!order) throw new Error('No order created');
-  if (!cleengId) throw new Error('cleengId is not configured');
-  if (!user || !auth) throw new Error('user is not logged in');
+      if (!order) throw new Error('No order created');
+      if (!authProviderId) throw new Error('auth provider is not configured');
+      if (!checkoutService) throw new Error('checkout service is not available');
 
-  const response = await checkoutService.paymentWithoutDetails({ orderId: order.id }, cleengSandbox, auth.jwt);
+      // subscribe to listen to inplayer websocket notifications
+      await subscribeToNotifications(customer?.uuid);
```
budzworthy, line 113: Was moving this here an effort to avoid unexpected notifications?  Imho, I think it's going to be more maintainable to subscribe upon logout still, but then check state data before action on certain messages.

src/services/inplayer.checkout.service.ts
```diff
@@ -0,0 +1,232 @@
...
+    return {
+      errors: [],
+      responseData: {
+        message: '',
+        paymentMethods,
+        status: 1,
+      },
+    };
+  } catch {
+    throw new Error('Failed to get payment methods');
```
budzworthy, line 64: It might be nice on all of these errors to keep the original error as the `cause`

src/stores/AccountController.ts
```diff
@@ -116,7 +118,9 @@ export const initializeAccount = async () => {
 export async function updateUser(
   values: { firstName: string; lastName: string } | { email: string; confirmationPassword: string },
 ): Promise<ServiceResponse<Customer>> {
-  return await withAccountService(async ({ accountService, sandbox }) => {
+  return await useService(async ({ accountService, sandbox = true }) => {
+    if (!accountService) throw new Error('account service is not configured');
```
budzworthy: It seems like most usages have this if (!accountService) check. Maybe it would cleaner to add that in the useService code and just use a try-catch in the few places you still want to do something when the service is undefined or have an optional input for `required` or something and only throw when that's set true (but set it true by default)

test-e2e/codecept.mobile.js
```diff
@@ -12,7 +12,7 @@ exports.config = {
   grep: '(?=.*)^(?!.*@desktop-only)',
   tests: ['./tests/**/*.js', './tests/**/*.ts'],
   output: './output/mobile',
-  timeout: 60,
+  timeout: 3000,
```
budzworthy: I believe this value is in seconds, so 3000 is almost an hour. We probably don't want to have tests that hang for that long. https://codecept.io/configuration/

test-e2e/codecept.desktop.js
```diff
@@ -11,7 +11,7 @@ exports.config = {
   grep: '(?=.*)^(?!.*@mobile-only)',
   tests: ['./tests/**/*.js', './tests/**/*.ts'],
   output: './output/desktop',
-  timeout: 60,
+  timeout: 3000,
```
budzworthy:
```suggestion
  timeout: 60,
```

test-e2e/codecept.mobile.js
budzworthy:
```suggestion
  timeout: 60,
```

src/stores/AccountController.ts
darkoatanasovski: sure, this is only for TVOD or configs without auth integration. But now, I can see that this function is called only for AUTHVOD and SVOD, so I believe that we can remove this if statement. The initial idea was to prevent requesting account services if client integration is not provided.

src/components/PaymentForm/PaymentForm.tsx
kmitrovv: It's not only the card payment data that is passed, but the entire form object. Things like handleBlur and handleChange are used here as well.

src/components/CreditCardExpiryField/CreditCardExpiryField.tsx
kmitrovv: ye, but only for semantics, since the library has inbuilt validator for this field, this doesn't do much really. Will update it anyways.

kmitrovv: This for be useful for mobile devices, to display only numeric keyboards though.

src/containers/AccountModal/forms/Checkout.tsx
kmitrovv: refactored.

src/containers/AccountModal/forms/Checkout.tsx
kmitrovv: We format the month and the year, because it's in different format from the backend. This is moved in the PaymentForm  component.

src/services/inplayer.checkout.service.ts
kmitrovv: Updated. No actually there is already a method for this `isSVODOffer` but there was a problem when using it. All good now.

src/services/inplayer.checkout.service.ts
kmitrovv: This is at service level. Wouldn't be able to use the `useTranslation` hook

src/containers/AccountModal/forms/Checkout.tsx
kmitrovv: already done with the previous comment.

src/hooks/useCheckAccess.ts
kmitrovv: At the moment where the hook is used the offerId is not sent (even though there is a param for that), but checked from the config.. If we want it to be mandatory, we would still need to somehow get it from the config and pass it as an argument.. not sure if it will change anything.

src/services/inplayer.subscription.service.ts
kmitrovv: I guess we started with that terminology and used it everywhere when formatting data. I will update all the process terms with format.

src/hooks/useCheckAccess.ts
kmitrovv: Hm, that way the payment will take longer I guess, because time will be needed after calling the subscribe. 
I don't think there will be moments for delayed notifications since we already updated the subscribe to happen only when payment is submitted.

types/checkout.d.ts
kmitrovv: Separate from offerId, we need the actual id on InPlayer side as well, which is being sent as a param (accessFeeId) for our payment methods.
The offerId we use through the application which is combination from the actual id, but formatted as the one which is already used.

src/stores/CheckoutController.ts
kmitrovv: yes, we only subscribe to listen to notifications when payment is submitted. On logout we unsubscribe.

kmitrovv: With the useCheckAccess hook, the notifications can be excluded though, only this way the entire process is faster. We are only using the notifications for access granted, but that is handled with useCheckAccess as well. 
We should discuss if we want to have the notifications at all. But maybe at some later stage they will be needed for something.. not sure

src/services/inplayer.checkout.service.ts
kmitrovv: That may require some effort since the sdk uses axios, and not all of our endpoints have the same error response data unfortunately.. We will need to create a handler and go through all the endpoints to format the errors. Not sure if it's worth the effort.

budzworthy: I mean something like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#rethrowing_an_error_with_a_cause

kmitrovv: aham, to include the entire error object as the cause, but not throw only the message. That's okay then.

src/services/cleeng.checkout.service.ts
kmitrovv: updated.

src/stores/AccountController.ts
kmitrovv: updated.

budzworthy (approved):

budzworthy:
Approved :tada:

PR #229, opened 2023-01-12
Inplayer/e2e test

darkoatanasovski:
## Description

implemented DataTable for running tests in a loop with different configs

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/services/inplayer.account.service.ts
```diff
@@ -274,8 +369,10 @@ function processAccount(account: AccountData): Customer {
 function processUpdateAccount(customer: UpdateCustomerArgs) {
   const firstName = customer.firstName?.trim() || '';
   const lastName = customer.lastName?.trim() || '';
-  const fullName = `${firstName} ${lastName}`;
-
+  let fullName = `${firstName} ${lastName}`;
+  if (!firstName && !lastName) {
+    fullName = '[EMPTY FULL NAME]';
```
darkoatanasovski: maybe we can use the email address as a default full name

budzworthy: Yes, agreed. I thought we had already decided this, but maybe it got lost in the shuffle.

darkoatanasovski:
note: we need to replace test URL configs with config ids before the merge

src/stores/AccountController.ts
```diff
@@ -66,9 +68,14 @@ export const handleVisibilityChange = () => {
...
 export const initializeAccount = async () => {
   await useService(async ({ accountService, config }) => {
+    if (!accountService) {
+      useAccountStore.setState({ loading: false });
+      return;
+    }
     useAccountStore.setState({
       loading: true,
       canUpdateEmail: accountService.canUpdateEmail,
+      canSupportEmptyFullName: accountService.canSupportEmptyFullName,
```
budzworthy: Do these properties all need to go in the store if they are always still available on the service anyway?

darkoatanasovski: ahh... yes, because these are used in the components, pages, etc., except `canSupportEmptyFullName` :)

src/stores/AccountController.ts
```diff
@@ -470,3 +487,15 @@ async function afterLogin(
     getPublisherConsents(),
   ]);
 }
+
+const validateInputLength = (values: { firstName: string; lastName: string }) => {
+  const errors: string[] = [];
+  if (Number(values?.firstName?.length) > 50) {
+    errors.push('firstName can have max 50 characters.');
```
budzworthy: Should these be added to translations?

src/utils/formatting.ts
```diff
@@ -81,7 +81,7 @@ export const episodeURL = (episode: PlaylistItem, seriesId?: string, play: boole
 export const formatDate = (dateString: number) => {
   if (!dateString) return '';
 
-  return new Date(dateString * 1000).toLocaleDateString();
+  return new Date(dateString * 1000).toLocaleDateString('en-US');
```
budzworthy, line 84: Is this just to fix the tests? I'm hesitant to hardcode this when it should be working to localize based on browser settings.

src/stores/AccountController.ts
```diff
@@ -126,7 +130,20 @@ export async function updateUser(
...
+      return {
+        errors,
+        responseData: {} as Customer,
+      };
+    }
+
+    let payload = values;
+    if (!accountService.canSupportEmptyFullName) {
+      payload = { ...values, email: user.email };
+    }
```
```suggestion
    const payload = values;
    if (!accountService.canSupportEmptyFullName) {
      payload.email = user.email;
    }
```

budzworthy: Also please add a comment here to explain why email is set when 'canSupportEmptyFullname' is false.

src/stores/AccountController.ts
```diff
@@ -470,3 +487,15 @@ async function afterLogin(
     getPublisherConsents(),
   ]);
 }
+
+const validateInputLength = (values: { firstName: string; lastName: string }) => {
```
budzworthy, line 493: Should the 50 character limit be a property on the services like canSupportEmptyFullName?

test-e2e/codecept.desktop.js
```diff
@@ -11,7 +11,7 @@ exports.config = {
   grep: '(?=.*)^(?!.*@mobile-only)',
   tests: ['./tests/**/*.js', './tests/**/*.ts'],
   output: './output/desktop',
-  timeout: 60,
+  timeout: 3000,
```
budzworthy, line 14: I think I've mentioned this before, but according to the config this value is in seconds, so 3000 is a long time. https://codecept.io/configuration/

I see the examples there look like it's ms, but the table of properties says seconds. Are you running into issues?

test-e2e/tests/account_test.ts
```diff
@@ -10,35 +10,27 @@ const firstNameField = 'firstName';
 const lastNameField = 'lastName';
 const consentCheckbox = 'Yes, I want to receive Blender updates by email';
 
-let loginContext: LoginContext;
+const loginContexts: { [key: string]: LoginContext } = {};
```
budzworthy: Does this need to be a dictionary? I would assume the tests run in isolated instances and could just both use a single value here still.

test-e2e/tests/account_test.ts
```diff
@@ -10,35 +10,27 @@ const firstNameField = 'firstName';
...
-
-    I.click('Continue');
-    I.waitForLoaderDone();
-
-    I.clickCloseButton();
-  });
-});
-
-Scenario('I can see my account data', async ({ I }) => {
+Data(configs).Scenario('I can see my account data', async ({ I, current }) => {
```
budzworthy: Since `Data(configs).` has to be prepended to each test, I'm a little nervous that developers are going to forget to do it when adding future account tests. Unfortunately, I don't see a way to use Data at a Feature level, but if you wrap all of the code from Feature through the end into a function (call it runTestSuite or something) then you can just call that twice with your inputs for config, authProvider, etc. It's a little bit more vanilla, but I think it's actually a bit more scaleable and maintainable too. I think if you put 'loginContext' inside that function scope it will also work w/o the dictionary, and all the test parameter data is just global within that function instead of having to do `current.<...>`.  Here's a screenshot of a quick proof of concept I just tested locally:
image

budzworthy: Oh the `Before` works better this way too, so you don't have to call `I.beforeAccount` in each test.

test-e2e/tests/account_test.ts
```diff
@@ -53,41 +45,53 @@ Scenario('I can see my account data', async ({ I }) => {
...
-  I.see('I accept the Terms and Conditions of Cleeng.');
+  I.see(`I accept the Terms and Conditions of ${current.authProvider}.`);
   I.see(consentCheckbox);
 
   I.seeInCurrentUrl(constants.accountsUrl);
 });
 
-Scenario('I can cancel Edit account', async ({ I }) => {
+Data(configs).Scenario('I can cancel Edit account', async ({ I, current }) => {
+  if (!current.canEditEmail) {
```
budzworthy: Does it make sense to move the 5 tests with 'if not edit email then return' check into their own file/suite called edit_email_test or something?

test-e2e/tests/account_test.ts
```diff
@@ -98,13 +102,21 @@ Scenario('I get a wrong password warning', async ({ I }) => {
...
+  }
+  loginContexts[current.config.label] = await I.beforeAccount(current.config, loginContexts[current.config.label], firstName, lastName);
   I.amOnPage(constants.accountsUrl);
   I.click(editAccount);
   await passwordUtils.testPasswordToggling(I, 'confirmationPassword');
 });
 
-Scenario('I can reset my password', async ({ I }) => {
+Data(configs).Scenario('I can reset my password (ResetLink)', async ({ I, current }) => {
+  if (current.resetPasswordType !== 'resetlink') {
```
budzworthy: Is there a test for reset password 'direct'?

src/stores/AccountController.ts
darkoatanasovski: yes :|

test-e2e/codecept.desktop.js
darkoatanasovski: yes, you are right, the timeout value is in seconds. I tried to figure out why tests are failing with more than 2 workers. I will revert back to 60.

budzworthy: Interesting, I run with 8 workers and don't have any problems

test-e2e/tests/account_test.ts
kmitrovv: there is no way to write a test for this since the sandbox mails are received on mailhog.

src/components/PaymentForm/PaymentForm.tsx
```diff
@@ -99,7 +99,7 @@ const PaymentForm: React.FC<Props> = ({ couponCode, setUpdatingOrder }) => {
         </div>
       </div>
       <div>
-        <Button label="continue" variant="contained" color="primary" onClick={paymentData.handleSubmit as () => void} size="large" fullWidth />
+        <Button label="Continue" variant="contained" color="primary" onClick={paymentData.handleSubmit as () => void} size="large" fullWidth />
```
budzworthy: Can you add this to translations?

src/services/inplayer.account.service.ts
```diff
@@ -367,8 +363,10 @@ function formatAccount(account: AccountData): Customer {
 function formatUpdateAccount(customer: UpdateCustomerArgs) {
   const firstName = customer.firstName?.trim() || '';
   const lastName = customer.lastName?.trim() || '';
-  const fullName = `${firstName} ${lastName}`;
-
+  let fullName = `${firstName} ${lastName}`;
+  if (!firstName && !lastName) {
+    fullName = customer.email as string;
+  }
```
budzworthy:
```suggestion
  const fullName = `${firstName} ${lastName}`.trim() || (customer.email as string);
```

src/services/inplayer.account.service.ts
```diff
@@ -402,7 +400,7 @@ function formatPublisherConsents(consent: Partial<GetRegisterField>) {
 }
 
 function getTermsConsent(): Consent {
-  const label = 'I accept the <a href="https://inplayer.com/legal/terms" target="_blank">Terms and Conditions</a> of InPlayer.';
+  const label = 'I accept the <a href="https://inplayer.com/legal/terms" target="_blank">Terms and Conditions</a> of JW Player.';
```
budzworthy: Please add to translations

src/stores/AccountController.ts
```diff
@@ -126,7 +131,21 @@ export async function updateUser(
       throw new Error('no auth');
     }
 
-    const response = await accountService.updateCustomer({ ...values, id: user.id.toString() }, sandbox, auth.jwt);
+    const errors = validateInputLength(values as FirstLastNameInput);
```
budzworthy, line 134: Should there be a check here to make sure we only call this when values is of type FirstLastNameInput? I imagine we might also want basic validation when the type is EmailConfirmPasswordInput?

kmitrovv: hmm, I think this part can be completely refactored and improved, the types and validations as well, but I don't think it should be done in this PR.

kmitrovv (approved):

PR #236, opened 2023-01-24
feat(project): add dev api support

AntonLantukh:
- add dev api for internal testing purposes
- add dev player for testing purposes

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/src/components/Root/Root.tsx
+++ b/src/components/Root/Root.tsx
@@ -14,6 +14,7 @@ import { loadAndValidateConfig } from '#src/utils/configLoad';
 import { initSettings } from '#src/stores/SettingsController';
 import AppRoutes from '#src/containers/AppRoutes/AppRoutes';
 import registerCustomScreens from '#src/screenMapping';
+import { initializeHost } from '#src/stores/ConfigController';
 
 const Root: FC = () => {
   const { t } = useTranslation('error');
@@ -34,6 +35,10 @@ const Root: FC = () => {
       setSearchParams(searchParams, { replace: true });
     }
   }, [configSource, searchParams, setSearchParams, settingsQuery.data]);
+
+  useEffect(() => {
+    initializeHost(searchParams);
+  }, [configSource, searchParams]);
 
   const configQuery = useQuery('config-init-' + configSource, async () => await loadAndValidateConfig(configSource), {
     enabled: settingsQuery.isSuccess,
--- a/src/components/Root/Root.tsx
+++ b/src/components/Root/Root.tsx
@@ -14,7 +14,6 @@ import { loadAndValidateConfig } from '#src/utils/configLoad';
 import { initSettings } from '#src/stores/SettingsController';
 import AppRoutes from '#src/containers/AppRoutes/AppRoutes';
 import registerCustomScreens from '#src/screenMapping';
-import { initializeHost } from '#src/stores/ConfigController';
 
 const Root: FC = () => {
   const { t } = useTranslation('error');
@@ -27,8 +26,7 @@ const Root: FC = () => {
 
   const [searchParams, setSearchParams] = useSearchParams();
 
-  const configSource = useMemo(() => getConfigSource(searchParams, settingsQuery.data), [searchParams, settingsQuery.data]);
-
+  const { configSource, env } = useMemo(() => getConfigSource(searchParams, settingsQuery.data), [searchParams, settingsQuery.data]);
   // Update the query string to maintain the right params
   useEffect(() => {
     if (settingsQuery.data && cleanupQueryParams(searchParams, settingsQuery.data, configSource)) {
@@ -36,11 +34,7 @@ const Root: FC = () => {
     }
   }, [configSource, searchParams, setSearchParams, settingsQuery.data]);
 
-  useEffect(() => {
-    initializeHost(searchParams);
-  }, [configSource, searchParams]);
-
-  const configQuery = useQuery('config-init-' + configSource, async () => await loadAndValidateConfig(configSource), {
+  const configQuery = useQuery('config-init-' + configSource, async () => await loadAndValidateConfig(configSource, env), {
     enabled: settingsQuery.isSuccess,
     retry: configSource ? 1 : 0,
     refetchInterval: false,
--- a/src/config.ts
+++ b/src/config.ts
@@ -15,14 +15,7 @@ export const ADYEN_LIVE_CLIENT_KEY = 'live_BQDOFBYTGZB3XKF62GBYSLPUJ4YW2TPL';
 // how often the live channel schedule is refetched in ms
 export const LIVE_CHANNELS_REFETCH_INTERVAL = 15 * 60_000;
 
-// OTT shared player (for dev and prd api host)
-export const DEFAULT_PLAYER = {
-  dev: 'GbLTFhkw',
-  prd: 'M4qoGvUk',
-};
-
-// Dev JW API
-export const API_HOST = {
-  dev: 'https://content-portal.jwplatform.com',
-  prd: 'https://cdn.jwplayer.com',
+export const ENVIRONMENT = {
+  dev: { player: 'GbLTFhkw', host: 'https://content-portal.jwplatform.com' },
+  prd: { player: 'M4qoGvUk', host: 'https://cdn.jwplayer.com' },
 };
--- a/src/config.ts
+++ b/src/config.ts
@@ -15,5 +15,14 @@ export const ADYEN_LIVE_CLIENT_KEY = 'live_BQDOFBYTGZB3XKF62GBYSLPUJ4YW2TPL';
 // how often the live channel schedule is refetched in ms
 export const LIVE_CHANNELS_REFETCH_INTERVAL = 15 * 60_000;
 
-// OTT shared player
-export const DEFAULT_PLAYER_ID: string = import.meta.env.APP_DEFAULT_PLAYER ?? 'M4qoGvUk';
+// OTT shared player (for dev and prd api host)
+export const DEFAULT_PLAYER = {
+  dev: 'GbLTFhkw',
+  prd: 'M4qoGvUk',
+};
+
+// Dev JW API
+export const API_HOST = {
+  dev: 'https://content-portal.jwplatform.com',
+  prd: 'https://cdn.jwplayer.com',
+};
--- a/src/containers/PlayerContainer/PlayerContainer.tsx
+++ b/src/containers/PlayerContainer/PlayerContainer.tsx
@@ -40,8 +40,7 @@ const PlayerContainer: React.FC<Props> = ({
   liveStartDateTime,
   autostart,
 }: Props) => {
-  const { config } = useConfigStore((s) => s);
-  const continueWatchingList = config?.features?.continueWatchingList;
+  const continueWatchingList = useConfigStore((s) => s.config?.features?.continueWatchingList);
   const watchHistoryEnabled = !!continueWatchingList;
 
   // state
--- a/src/containers/PlayerContainer/PlayerContainer.tsx
+++ b/src/containers/PlayerContainer/PlayerContainer.tsx
@@ -8,7 +8,7 @@ import { useConfigStore } from '#src/stores/ConfigStore';
 import Player from '#components/Player/Player';
 import type { JWPlayer } from '#types/jwplayer';
 import { useWatchHistoryStore } from '#src/stores/WatchHistoryStore';
-import { DEFAULT_PLAYER_ID, VideoProgressMinMax } from '#src/config';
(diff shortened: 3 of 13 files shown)


src/utils/configOverride.ts
```diff
@@ -16,7 +16,10 @@ export function getConfigNavigateCallback(navigate: NavigateFunction) {
     navigate(
       {
         pathname: '/',
-        search: new URLSearchParams([[configQueryKey, configSource]]).toString(),
+        search: new URLSearchParams([
+          [configQueryKey, configSource],
+          ['env', 'prd'],
```
budzworthy: Is this needed? I thought the default would be `prd`?

AntonLantukh: Have not finalised it yet. Was testing one of the concepts using preview build. The idea was to preserve `dev` env when navigating within the app. However we still need to reset it when using config selector or config dialog.

budzworthy: Can you mark as WIP until its ready to review?

AntonLantukh:
@dbudzins should work now.

1. api host value gets saved internally why navigating within application. We don't keep the value in the query, if the page gets reloaded we lose it.
2. In case we change the config to a production one, we use a production api by default (when no additional `env` query provided)
3. I have also added a separate player for `dev` stored in our ott account.
4. player value for `dev` environment is taken only from a local value in the codebase. For production api we first analyse the `.env` file and then use a local value (as it worked previously).

src/stores/ConfigController.ts
```diff
@@ -1,6 +1,25 @@
+import { API_HOST } from '#src/config';
+import { IS_PROD_MODE } from '#src/utils/common';
 import { useConfigStore } from '#src/stores/ConfigStore';
 import { getAdSchedule } from '#src/services/api.service';
 
+const envQueryKey = 'env';
+
+export const initializeHost = (searchParams: URLSearchParams) => {
+  const query = searchParams.get(envQueryKey);
```
budzworthy: I think it would make sense to move the searchParam logic into `getConfigSource` in configOverride.ts (which would return `{configSource: string, host: string}`

src/stores/ConfigStore.ts
```diff
@@ -49,6 +52,13 @@ export const useConfigStore = createStore<ConfigState>('ConfigStore', (_, get) =
   },
   accessModel: 'SVOD',
   adScheduleData: null,
+  apiHost: import.meta.env.APP_API_BASE_URL || API_HOST.prd,
+  getPlayer: (): string => {
+    const host = get().apiHost;
+
+    // For dev api host we use only one testing id, for prod it can be also set in .env file
+    return host === API_HOST.dev ? DEFAULT_PLAYER.dev : import.meta.env.APP_DEFAULT_PLAYER ?? DEFAULT_PLAYER.prd;
```
budzworthy: It's a little non-intuitive that this check depends on comparing the host to the host constant and also doing this as a method instead of just checking and setting it once. I suggest doing this check in `loadAndValidateConfig` in configLoad.ts.

src/config.ts
```diff
@@ -15,5 +15,14 @@ export const ADYEN_LIVE_CLIENT_KEY = 'live_BQDOFBYTGZB3XKF62GBYSLPUJ4YW2TPL';
 // how often the live channel schedule is refetched in ms
 export const LIVE_CHANNELS_REFETCH_INTERVAL = 15 * 60_000;
 
-// OTT shared player
-export const DEFAULT_PLAYER_ID: string = import.meta.env.APP_DEFAULT_PLAYER ?? 'M4qoGvUk';
+// OTT shared player (for dev and prd api host)
```
budzworthy: I would recommend combining the player and host constants into an environments hash like:
`const environments = {dev: {host: 'https://content-portal.jwplatform.com', player: 'GbLTFhkw'}, prd: {host: 'https://cdn.jwplayer.com', player: 'M4qoGvUk'}}` and it can be moved to a non-exported const in `loadAndValidateConfig` assuming the other changes are made as well. Maybe there's also a way to write that code declaritively so that the dev environment gets compiled away in Prod mode instead of sticking around as an inaccessible constant.

src/containers/PlayerContainer/PlayerContainer.tsx
```diff
@@ -40,8 +40,8 @@ const PlayerContainer: React.FC<Props> = ({
   liveStartDateTime,
   autostart,
 }: Props) => {
-  const { features } = useConfigStore((s) => s.config);
-  const continueWatchingList = features?.continueWatchingList;
+  const { config } = useConfigStore((s) => s);
+  const continueWatchingList = config?.features?.continueWatchingList;
```
budzworthy:
```suggestion
  const continueWatchingList = useConfigStore(s => s.config?.features?.continueWatchingList);
```

src/services/api.service.ts
```diff
@@ -57,7 +57,7 @@ export const getPlaylistById = async (id?: string, params: GetPlaylistParams = {
   }
 
   const pathname = drmPolicyId ? `/v2/playlists/${id}/drm/${drmPolicyId}` : `/v2/playlists/${id}`;
-  const url = addQueryParams(`${import.meta.env.APP_API_BASE_URL}${pathname}`, params);
+  const url = addQueryParams(useConfigStore.getState().apiHost + pathname, params);
```
budzworthy:
```suggestion
  const url = addQueryParams(useConfigStore(s => s.apiHost) + pathname, params);
```

src/services/api.service.ts
```diff
@@ -75,7 +75,7 @@ export const getMediaByWatchlist = async (playlistId: string, mediaIds: string[]
   }
 
   const pathname = `/apps/watchlists/${playlistId}`;
-  const url = addQueryParams(`${import.meta.env.APP_API_BASE_URL}${pathname}`, { token, media_ids: mediaIds });
+  const url = addQueryParams(useConfigStore.getState().apiHost + pathname, { token, media_ids: mediaIds });
```
budzworthy:
```suggestion
  const url = addQueryParams(useConfigStore(s => s.apiHost) + pathname, { token, media_ids: mediaIds });
```

src/services/api.service.ts
```diff
@@ -92,7 +92,7 @@ export const getMediaByWatchlist = async (playlistId: string, mediaIds: string[]
  */
 export const getMediaById = async (id: string, token?: string, drmPolicyId?: string): Promise<PlaylistItem | undefined> => {
   const pathname = drmPolicyId ? `/v2/media/${id}/drm/${drmPolicyId}` : `/v2/media/${id}`;
-  const url = addQueryParams(`${import.meta.env.APP_API_BASE_URL}${pathname}`, { token });
+  const url = addQueryParams(useConfigStore.getState().apiHost + pathname, { token });
```
budzworthy:
```suggestion
  const url = addQueryParams(useConfigStore(s => s.apiHost) + pathname, { token });
```

src/services/api.service.ts
```diff
@@ -130,7 +130,7 @@ export const getSeries = async (id: string, params: GetSeriesParams = {}): Promi
   }
 
   const pathname = `/apps/series/${id}`;
-  const url = addQueryParams(`${import.meta.env.APP_API_BASE_URL}${pathname}`, params);
+  const url = addQueryParams(useConfigStore.getState().apiHost + pathname, params);
```
budzworthy:
```suggestion
  const url = addQueryParams(useConfigStore(s => s.apiHost) + pathname, params);
```

src/services/api.service.ts
```diff
@@ -143,7 +143,7 @@ export const getSeries = async (id: string, params: GetSeriesParams = {}): Promi
  */
 export const getSeriesByMediaIds = async (mediaIds: string[]): Promise<{ [key in typeof mediaIds[number]]: Series[] | undefined } | undefined> => {
   const pathname = `/apps/series`;
-  const url = addQueryParams(`${import.meta.env.APP_API_BASE_URL}${pathname}`, {
+  const url = addQueryParams(useConfigStore.getState().apiHost + pathname, {
```
budzworthy:
```suggestion
  const url = addQueryParams(useConfigStore(s => s.apiHost) + pathname, {
```

src/utils/configLoad.ts
```diff
@@ -137,7 +136,7 @@ function formatSourceLocation(source?: string) {
   }
 
   if (source.match(/^[a-z,\d]{8}$/)) {
-    return `${CONFIG_HOST}/apps/configs/${source}.json`;
+    return `${useConfigStore.getState().apiHost}/apps/configs/${source}.json`;
```
budzworthy:
```suggestion
    return `${useConfigStore(s => s.apiHost)}/apps/configs/${source}.json`;
```

src/services/api.service.ts
AntonLantukh: We can't use hooks this way here. Zustand provides non-reactive approach (applied here). Functions receive updated values when being invoked.

budzworthy: Oh, you're right. Ignore all these suggestions then.

src/stores/ConfigStore.ts
```diff
@@ -20,6 +20,7 @@ type CleengData = {
 type ConfigState = {
   config: Config;
   accessModel: AccessModel;
+  environments: { host: string; player: string } | null;
```
budzworthy: Naming suggestion to make singular and don't allow it to be null (initial value can be the production defaults)
```suggestion
  environment: { host: string; player: string };
```

AntonLantukh: @dbudzins env values are optional (not present for all modes) so I chose a local constant to avoid `undefined`.

src/config.ts
```diff
@@ -15,5 +15,7 @@ export const ADYEN_LIVE_CLIENT_KEY = 'live_BQDOFBYTGZB3XKF62GBYSLPUJ4YW2TPL';
 // how often the live channel schedule is refetched in ms
 export const LIVE_CHANNELS_REFETCH_INTERVAL = 15 * 60_000;
 
-// OTT shared player
-export const DEFAULT_PLAYER_ID: string = import.meta.env.APP_DEFAULT_PLAYER ?? 'M4qoGvUk';
+export const ENVIRONMENT = {
```
budzworthy, line 18:
```suggestion
export const ENVIRONMENTS = {
```

src/components/Root/Root.tsx
```diff
@@ -26,16 +26,15 @@ const Root: FC = () => {
 
   const [searchParams, setSearchParams] = useSearchParams();
 
-  const configSource = useMemo(() => getConfigSource(searchParams, settingsQuery.data), [searchParams, settingsQuery.data]);
-
+  const { configSource, env } = useMemo(() => getConfigSource(searchParams, settingsQuery.data), [searchParams, settingsQuery.data]);
```
budzworthy, line 29: Can you think of another term for these besides environments and the abbreviation env?  It will be easy to confuse this with the .env file and values.  Maybe 'host', 'context', 'platform', or even just 'backend'?  Or we could make the variable really verbose but easier to read, like 'deploymentEnvironment'?

AntonLantukh: @dbudzins jwApiHost / jwApiHostVersion?

src/utils/configOverride.ts
```diff
@@ -1,16 +1,31 @@
...
 const configQueryKey = 'app-config';
 const configLegacyQueryKey = 'c';
+const envQueryKey = 'env';
 
 const configFileStorageKey = 'config-file-override';
 
+export const getApiEnv = (searchParams: URLSearchParams): 'dev' | 'prd' => {
+  const query = searchParams.get(envQueryKey);
+
+  if (IS_PROD_MODE) {
```
budzworthy, line 18: I'm not sure this will result in static enough code for the compiler to optimize it through method calls and remove all the unused code in prod mode. Maybe something closer to the usages like a `getHost` method that returns the hardcoded value in prod or the dynamic value otherwise?

src/stores/ConfigStore.ts
```diff
@@ -20,6 +21,7 @@ type CleengData = {
 type ConfigState = {
   config: Config;
   accessModel: AccessModel;
+  environments: { host: string; player: string };
```
budzworthy, line 24: I don't know if I'm comfortable having these values in the Config, since loading the config depends on the host in the first place. Maybe they can be added to the 'settings' instead?

src/components/Root/Root.tsx
budzworthy: apiEnvironment?

budzworthy:
PR is a bit out of date from several refactors. If we pursue this, we can use the new structures to start fresh.

PR #245, opened 2023-02-07
fix: Broken screen when granted access

kmitrovv:
## Description

The payments screen was broken when access was being granted from InPlayer dashboard.
This was happening because the grant access is not creating an active subscription, but only gives access to that certain asset (offer). 

- [x] Acceptance criteria met
- [ ] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -73,7 +73,7 @@ const Payment = ({
       case 'granted':
         return t('user:payment.granted_subscription');
       default:
-        throw 'Unknown period';
+        return t('user:payment.other');
     }
   }
 
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -70,6 +70,8 @@ const Payment = ({
         return t('user:payment.daily_subscription');
       case 'week':
         return t('user:payment.weekly_subscription');
+      case 'granted':
+        return t('user:payment.granted_subscription');
       default:
         throw 'Unknown period';
     }
@@ -87,7 +89,7 @@ const Payment = ({
               <div className={styles.infoBox} key={activeSubscription.subscriptionId}>
                 <p>
                   <strong>{getTitle(activeSubscription.period)}</strong> <br />
-                  {activeSubscription.status === 'active'
+                  {activeSubscription.status === 'active' && activeSubscription.period !== 'granted'
                     ? t('user:payment.next_billing_date_on', { date: formatDate(activeSubscription.expiresAt) })
                     : t('user:payment.subscription_expires_on', { date: formatDate(activeSubscription.expiresAt) })}
                 </p>
@@ -96,7 +98,7 @@ const Payment = ({
                   <small>/{t(`account:periods.${activeSubscription.period}`)}</small>
                 </p>
               </div>
-              {activeSubscription.status === 'active' ? (
+              {activeSubscription.status === 'active' && activeSubscription.period !== 'granted' ? (
                 <Button label={t('user:payment.cancel_subscription')} onClick={onCancelSubscriptionClick} />
               ) : canRenewSubscription ? (
                 <Button label={t('user:payment.renew_subscription')} onClick={onRenewSubscriptionClick} />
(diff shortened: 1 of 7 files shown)


src/components/Payment/Payment.tsx
```diff
@@ -96,7 +98,7 @@ const Payment = ({
                   <small>/{t(`account:periods.${activeSubscription.period}`)}</small>
```
budzworthy, line 98: Can you add a transtlation for `account:periods.granted`? Maybe just 'once'?

src/components/Payment/Payment.tsx
```diff
@@ -70,6 +70,8 @@ const Payment = ({
         return t('user:payment.daily_subscription');
       case 'week':
         return t('user:payment.weekly_subscription');
+      case 'granted':
+        return t('user:payment.granted_subscription');
       default:
         throw 'Unknown period';
```
budzworthy: Maybe it's worth it to return a generic placeholder here instead of throwing an exception? That way the page won't die if there are any other edge cases. Maybe just 'Other' would be OK?

src/services/inplayer.subscription.service.ts
```diff
@@ -108,27 +111,29 @@ const formatCardDetails = (card: Card & { card_type: string; account_id: number
...
-    transactionPriceInclTax: transaction.purchased_amount?.toString(),
+    transactionPriceInclTax: purchasedAmount,
     customerId: transaction.customer_id?.toString(),
     customerEmail: transaction.consumer_email,
     customerLocale: '',
     customerCountry: 'en',
     customerIpCountry: '',
     customerCurrency: '',
-    paymentMethod: transaction.payment_method,
+    paymentMethod: transaction.payment_method || 'grant access',
```
budzworthy: Add translation

src/services/inplayer.subscription.service.ts
```diff
@@ -108,27 +111,29 @@ const formatCardDetails = (card: Card & { card_type: string; account_id: number
 
 // TODO: fix PurchaseDetails type in InPlayer SDK
 const formatTransaction = (transaction: InPlayerPurchaseDetails): Transaction => {
+  const purchasedAmount = transaction?.purchased_amount?.toString() || '0';
+
   return {
-    transactionId: transaction.parent_resource_id,
+    transactionId: transaction.parent_resource_id || 'access granted',
```
budzworthy: Add translation

src/services/inplayer.subscription.service.ts
```diff
@@ -108,27 +111,29 @@ const formatCardDetails = (card: Card & { card_type: string; account_id: number
...
     transactionDate: transaction.created_at,
-    offerId: transaction.purchased_access_fee_id?.toString(),
+    offerId: transaction.purchased_access_fee_id?.toString() || '/',
```
budzworthy: What is the slash for?

src/components/Payment/Payment.tsx
kmitrovv: Wanted to make as less changes as possible so I did not touched it, but I guess it makes sense.

src/services/inplayer.subscription.service.ts
kmitrovv: That is the case when the access is granted. Here we don't have the info about the accessFee, so not sure what to return in that case. Either '/' or '0' maybe..

src/services/inplayer.subscription.service.ts
kmitrovv: hmm but this is at service level.. I don't think we can use hooks here..

src/components/Payment/Payment.tsx
kmitrovv: You mean like this? 
image

src/services/inplayer.subscription.service.ts
budzworthy: Maybe add it to the translations. For English I would recommend 'Other', 'None', 'No Transaction', or something along those lines

src/components/Payment/Payment.tsx
budzworthy: Yep!

src/services/inplayer.subscription.service.ts
budzworthy: You can just import `i18next` and call it via `i18next.t('<namespace>:<key>')`.  See https://github.com/jwplayer/ott-web-app/blob/1733977a06cbfeef97a4cbd54d46235bf3339079/src/services/config.service.ts#L117

budzworthy (approved):

PR #254, opened 2023-03-06
fix: Watch history update on new media

kmitrovv:
## Description

Small fix for a bug on the InPlayer integration when trying to save watch history data while watching new video.

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/services/inplayer.account.service.ts
```diff
@@ -283,7 +284,8 @@ export const updatePersonalShelves: UpdatePersonalShelves = async (payload) => {
             await InPlayer.Account.updateWatchHistory(history.mediaid, history.progress);
           }
         });
-      } else {
+      }
+      if (!currentWatchHistoryIds?.includes(history.mediaid)) {
         await InPlayer.Account.updateWatchHistory(history.mediaid, history.progress);
```
dbudzins, line 289: Wait won't this prevent the progress from being updated?

src/services/inplayer.account.service.ts
```diff
@@ -283,7 +284,8 @@ export const updatePersonalShelves: UpdatePersonalShelves = async (payload) => {
             await InPlayer.Account.updateWatchHistory(history.mediaid, history.progress);
           }
         });
-      } else {
+      }
+      if (!currentWatchHistoryIds?.includes(history.mediaid)) {
```
budzworthy: Wait won't this prevent the progress from being updated?

kmitrovv: No, this checks if the `mediaid` is not present in the store (currentWatchHistoryIds), then make a request with that id.

budzworthy: I'm sorry, I still don't understand this... it looks like these 2 clauses are just doing the same thing now. Both ifs call `await InPlayer.Account.updateWatchHistory(history.mediaid, history.progress);` and they are essentially checking if externalData?.history contains an item with history.mediaid.

kmitrovv: The first one is only called when there is at least one item in the store and checks if the progress of some of the existing ones is updated, then call the endpoint. 
The newly added check, only calls the endpoint for new watch history items or if there is no watch history item at all (since the other one is only valid if there is at least one item).

budzworthy: I think this can be simplified then.  Basically what you want to check is:

`if (!currentWatchHistoryIds?.includes(history.mediaid) || externalData?.history?.some((e) => e.mediaid == history.mediaid && e.progress != history.progress))`

right?

kmitrovv: Didn't wanna spend time on refactoring this, just wanted to quickly fix the bug, but yeah, it will be much simpler this way. I will update it.

budzworthy (approved):
Thanks for fixing the leap day stuff!

PR #255, opened 2023-03-16
JWP signing service

kmitrovv:
## Description

Implemented JW signing service and updated the usage of `useContentProtection` hook.

This PR solves # .
Media items now can be signed with and without DRM protection.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/src/containers/Cinema/__snapshots__/Cinema.test.tsx.snap
+++ b/src/containers/Cinema/__snapshots__/Cinema.test.tsx.snap
@@ -10,10 +10,16 @@ exports[`<Cinema> > renders and matches snapshot 1`] = `
       class="_cinema_555f3c"
     >
       <div
-        class="_container_20cd38"
-        data-testid="player-container"
+        class="_loadingOverlay_eb61cb _inline_eb61cb"
       >
-        <div />
+        <div
+          class="_buffer_d122df"
+        >
+          <div />
+          <div />
+          <div />
+          <div />
+        </div>
       </div>
       <div
         style="transition: opacity 0.25s ease-in-out; opacity: 0;"
--- a/src/containers/PlayerContainer/PlayerContainer.tsx
+++ b/src/containers/PlayerContainer/PlayerContainer.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useLayoutEffect, useMemo, useState } from 'react';
 
 import { useWatchHistoryListener } from '#src/hooks/useWatchHistoryListener';
 import type { PlaylistItem } from '#types/playlist';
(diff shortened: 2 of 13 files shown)


src/containers/Cinema/__snapshots__/Cinema.test.tsx.snap
budzworthy, line 1: Any idea why this snapshot changed?

src/containers/PlayerContainer/PlayerContainer.tsx
```diff
@@ -43,6 +48,7 @@ const PlayerContainer: React.FC<Props> = ({
   const { features } = useConfigStore((s) => s.config);
   const continueWatchingList = features?.continueWatchingList;
   const watchHistoryEnabled = !!continueWatchingList;
+  const { data, isLoading } = useContentProtection('media', item.mediaid, (token, drmPolicyId) => getMediaById(item.mediaid, token, drmPolicyId));
```
budzworthy: Can you rename `data` to `signedItem`?

types/Config.d.ts
```diff
@@ -19,13 +19,23 @@ export type Config = {
...
+  contentProtection?: ContentProtection;
 };
 
 export type ContentSigningConfig = {
   host: string;
   drmPolicyId?: string;
 };
 
+export type ContentProtection = {
+  accessModel: 'free' | 'authfree' | 'authvod' | 'svod';
```
budzworthy: We called the new setting 'freeauth': https://github.com/jwplayer/showcase-services/blob/962f7c48efcd0c9673eb5e8e06dcc4edc801707e/src/management/schemas/app_configs.py#L81

```suggestion
  accessModel: 'free' | 'freeauth' | 'authvod' | 'svod';
```

src/services/config.service.ts
```diff
@@ -72,6 +72,14 @@ const configSchema: SchemaOf<Config> = object({
     host: string().required() as StringSchema<string>,
     drmPolicyId: string().notRequired(),
   }),
+  contentProtection: object()
+    .shape({
+      accessModel: string().oneOf(['free', 'authfree', 'authvod', 'svod']).notRequired(),
```
budzworthy:
```suggestion
      accessModel: string().oneOf(['free', 'freeauth', 'authvod', 'svod']).notRequired(),
```

budzworthy (approved):
Just a few very small things to fix. Overall looks really good!

src/containers/Cinema/__snapshots__/Cinema.test.tsx.snap
kmitrovv: I think it's because of the changes in the `PlayerContainer.tsx`:
```
if (!data || isLoading) {
    return <LoadingOverlay inline />;
}
```

src/containers/PlayerContainer/PlayerContainer.tsx
kmitrovv: yeah we should change the name, what do you think about `mediaItem` , because not always it is a signed item.

budzworthy: I think we should differentiate it from the existing `item` variable. Maybe we name them `itemMetadata` and `playableItem` or something like that?

src/containers/Cinema/__snapshots__/Cinema.test.tsx.snap
budzworthy: Ah, ok. Missed that dependency.

src/containers/PlayerContainer/PlayerContainer.tsx
kmitrovv: `playableItem` sounds good to me.

PR #256, opened 2023-03-21
fix(analytics): fix watch duration by filtering seek events

AntonLantukh:
## Description

Fix watch duration and ad seek event

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


public/jwpltx.js
```diff
@@ -37,10 +10,44 @@ window.jwpltx = window.jwpltx || {};
...
-  // Current time for live streams.
-  let current;
+  // Time watched after last t event was sent
+  let timeWatched = 0;
+  // Last progress of the video stored
+  let lastVp = 0;
+  // Last quantile reached
+  let lastQ;
+  // Seeking state, gets reset after 1 sec of the latest seeked event
+  let isSeeking = null;
```
budzworthy: Can you just initialize this to false?
```suggestion
  let isSeeking = false;
```

public/jwpltx.js
```diff
@@ -37,10 +10,44 @@ window.jwpltx = window.jwpltx || {};
...
+  // Timeout for seeking state
+  let liveInterval = null;
+
+  // How many quantiles have we passed (whether we need to send new t event or not)
+  function getLastVODQuantile(progress, duration, numberOfQuantiles) {
+    return Math.floor(progress / (duration / numberOfQuantiles));
+  }
+
+  // Here we convert seconds watched to the unites accepted by analytics service (res is 0 - 128)
```
budzworthy:
```suggestion
  // Here we convert seconds watched to quantiles, the units accepted by the analytics service (res is 0 - 128)
```

public/jwpltx.js
AntonLantukh: Good catch, thank you

public/jwpltx.js
```diff
@@ -98,27 +150,51 @@ window.jwpltx = window.jwpltx || {};
...
           uri.q = 32;
         }
-        uri.ti = Math.round(uri.vd / uri.q);
+
+        uri.ti = 0;
         uri.pw = 0;
-        sendData('s');
 
+        // Initialize latest quantile to compare further quantiles with
+        lastQ = getLastVODQuantile(vp, vd, uri.q);
```
budzworthy: vd = 0
```suggestion
        lastQ = 1;
```

public/jwpltx.js
```diff
@@ -37,10 +10,44 @@ window.jwpltx = window.jwpltx || {};
...
+  let lastQ;
+  // Seeking state, gets reset after 1 sec of the latest seeked event
+  let isSeeking = null;
+  // Timeout for seeking state
+  let liveInterval = null;
+
+  // How many quantiles have we passed (whether we need to send new t event or not)
+  function getLastVODQuantile(progress, duration, numberOfQuantiles) {
+    return Math.floor(progress / (duration / numberOfQuantiles));
+  }
```
budzworthy: Let's not call this a quantile because the quantiles are 1-128 and this is 1-32 (max depending on the chunks.)
Also, suggest adding 1 so it matches the 1-32 instead of going 0-31.

```suggestion
  // How many quantiles have we passed (whether we need to send new t event or not)
  function getLastChunkNumber(progress, duration) {
    return Math.floor(progress / duration * uri.q) + 1;
  }
```

public/jwpltx.js
```diff
@@ -61,23 +68,68 @@ window.jwpltx = window.jwpltx || {};
...
+      // We need to rewrite progress of the video when seeking to have a valid ti param
+      lastVp = offset;
+      lastQ = getLastVODQuantile(offset, duration, uri.q);
+    }
+  };
+
+  // Process seeked event
+  o.seeked = function () {
+    // There is currently a 1 sec debounce surrounding this event in order to logically group multiple `seeked` events
+    window.setTimeout(() => {
```
budzworthy: Don't we need to clear this timeout somewhere if we seek again?

public/jwpltx.js
```diff
@@ -61,23 +68,68 @@ window.jwpltx = window.jwpltx || {};
...
 
+  // Process seek event
+  o.seek = function (offset, duration) {
+    isSeeking = true;
+    // Clear interval in case of a live stream not to update time watched while seeking
+    if (uri.pw === -1) {
+      clearLiveInterval();
+    } else {
+      // We need to rewrite progress of the video when seeking to have a valid ti param
+      lastVp = offset;
```
budzworthy, line 119: This is going to get you a closer value for total seconds watched, but I think it causes some of the seconds to ultimately be attributed to the wrong quantile, doesn't it?

public/jwpltx.js
```diff
@@ -37,10 +10,44 @@ window.jwpltx = window.jwpltx || {};
...
+  // Timeout for seeking state
+  let liveInterval = null;
+
+  // How many quantiles have we passed (whether we need to send new t event or not)
+  function getLastVODQuantile(progress, duration, numberOfQuantiles) {
+    return Math.floor(progress / (duration / numberOfQuantiles));
+  }
+
+  // Here we convert seconds watched to the unites accepted by analytics service (res is 0 - 128)
+  function getProgressWatched(progress, duration) {
```
budzworthy:
```suggestion
  function getCurrentQuantile(progress, duration) {
```

public/jwpltx.js
```diff
@@ -98,27 +150,51 @@ window.jwpltx = window.jwpltx || {};
...
-        let pw = (Math.floor(vp / uri.ti) * 128) / uri.q;
-        if (pw != uri.pw) {
+        const pw = getProgressWatched(vp, vd);
+        const passedQ = getLastVODQuantile(vp, vd, uri.q);
+
+        // Total time watched since last t event.
+        timeWatched = timeWatched + (vp - lastVp);
+        lastVp = vp;
+
+        if (passedQ > lastQ) {
```
budzworthy: It might be easier to follow if you keep this check in terms of quantiles (1-128) and check that pw (which is the current quantile) is greater than or equal to lastQ + 128 / uri.q

public/jwpltx.js
AntonLantukh: This one is a bit tricky. We use `getLastVODQuantile` because a customer can start watching the video from Continue Watching section and in this case the value can be different.

budzworthy: Oh, missed that. Makes sense.

public/jwpltx.js
AntonLantukh: Yeah, we need to do it, thanks!

public/jwpltx.js
AntonLantukh: Not really. We store current watched seconds in `timeWatched` (seconds watched before seeking). 
Rewriting `lastVp` helps us to starting counting new seconds after seeking using updated video progress as starting point.

public/jwpltx.js
AntonLantukh: These quantiles are ones we use to send `t` events, so for example if the video length is less than 300 but greater than 180, then we send 16 of them during the video, every videoLength / numberOfQuantiles seconds.

lastQ - is a last passed quantile. So for example if we are in the middle of the video, then this one if 8. When it changes to 9, then we send a new `t` event.

AntonLantukh: So maybe we should keep `getLastVODQuantile ` name instead of `getLastChunkNumber` as it is more about quantiles and progress chunks.

budzworthy: Using the 'chunk' terminology is just to differentiate quantiles (1-128) from what I think you called VODQuantiles (1-32.) I don't think that is a common term and it's a bit confusing to have 2 different types of quantiles. If you don't mind refactoring, I think just using quantiles (1-128) for everything will make it the most readable.

AntonLantukh: I tried to do it this way:

```
// This way we can get quantile representation to compare with current progress
function getNextQuantile(progress, duration) {
    return Math.floor(getCurrentProgressQuantile(progress / duration) + MAX_DURATION_IN_QUANTILES / uri.q);
  }

// And then we can compare it with `pw`
if (pw >= nextQuantile) {
 ....
```

However we lose some accuracy in this case and player / ott flows start to behave a bit differently and some `ti` values differ as well.

AntonLantukh: I have changed the names of functions and variables a bit to show where chunks and where quantiles are.

AntonLantukh: One more thing, in docs we use `quantiles` for all metrics:

```
Quantile Metrics

Quantiles are to track how much of a video is watched. It's fired after the video plays past the threshold of a quantile. The number of quantiles in a video depends on video length.

There are 3 related parameters:

Video Duration vd
Video Quantiles q
Quantile Watched pw
```

public/jwpltx.js
budzworthy: Right, but my point is if you have 20s quantiles and you watch 3 seconds, then the user scrolls to 10 seconds into another quantile won't you end up saying that the second quantile had 13 seconds watched when the end is reached?  So it's the right number of seconds but in the wrong spots.

It's probably not a huge issue, but what I would expect is that as soon as a user finishes scrolling we send an event for the quantile and duration they were at when they started to scroll and then start counting a new duration in the new quantile.

AntonLantukh: @dbudzins that is how player flow works now. They also wait for the next quantile to send the `t` event with what was before and after seeking. Ott flow and player flow send events now within approximately the same time.

So if you watched 3 seconds after the quantile started and then seek to 10 of another quantile, we both wait 10 more seconds and send 13 seconds.

public/jwpltx.js
budzworthy: There is certainly confusion around the quantile numbers, counts, values and how they are used interchangeably, that's why I want to be careful about how we describe it too.

I think your math in the first comment needs to factor q differently to make sure you're still looking at the right spacing for the quantiles and then it should work:

```
// This way we can get quantile representation to compare with current progress
function getNextTriggerQuantile(progress, duration) {
    return Math.ceil(progress/duration * uri.q) * MAX_DURATION_IN_QUANTILES / uri.q;
  }

// And then we can compare it with `pw`
if (pw >= nextQuantile) {
 ....

Comment (budzworthy) [reply public/jwpltx.js]: For q = 32, getNextTriggerQuantile will give 4, 8, 12...

File: public/jwpltx.js (hunk 6 above)
Comment (budzworthy) [reply public/jwpltx.js:119]: Oh OK, seems like a quasi-bug to me, but fine if it matches. Might be more of an issue in longer form content, but probably not worth getting upset about.

Comment (AntonLantukh) [reply public/jwpltx.js:119]: Maybe it was implemented originally in accordance with the docs:

**t event** - `A user has watched past a threshold of a quantile which is determined based on video length`

So technically a threshold it only after certain playback points.

File: public/jwpltx.js (hunk 8 above)
Comment (AntonLantukh) [reply public/jwpltx.js]: Thank you! Seems to work properly now. I have changed both functions to use quantiles (0 - 128) now.

File: public/jwpltx.js (hunk 6 above)
Comment (budzworthy) [reply public/jwpltx.js:119]: It's a bit ambiguous whether watching part of a threshold section then seeking counts as "watched past", but as long as it's consistent with what the player seems to be doing and we have a solid grasp of it that's more important than semantics.

File: public/jwpltx.js (hunk 8 above)
Comment (budzworthy) [reply public/jwpltx.js]: Awesome! I will give it another look through!

File: public/jwpltx.js [hunk 9]
```diff
@@ -37,10 +10,63 @@ window.jwpltx = window.jwpltx || {};
...
+  function getCurrentProgressQuantile(progress, duration) {
+    return Math.floor(MAX_DURATION_IN_QUANTILES * (progress / duration));
+  }
+
+  // We convert `q` metric to quantiles (0 - 128) to define next breakpoint for `t` event
+  function getNextTriggerQuantile(progress, duration) {
+    return (Math.ceil((progress / duration) * uri.q) * MAX_DURATION_IN_QUANTILES) / uri.q;
+  }
+
+  // Here we convert seconds watched to quantiles, the units accepted by the analytics service (res is 0 - 128)
```
budzworthy:
```suggestion
  // Sends the last ping when closing the player or window
```

public/jwpltx.js
```diff
@@ -98,27 +165,51 @@ window.jwpltx = window.jwpltx || {};
...
-    if (uri.pw != 128) {
-      uri.pw = 128;
-      sendData('t');
+    // Clear intervals for live streams
+    if (uri.pw === -1) {
+      clearLiveInterval();
+    } else {
+      uri.pw = MAX_DURATION_IN_QUANTILES;
     }
+
```
budzworthy, line 220: Should we also clear the seek timeout here just in case they seeked to the end or almost the end?

public/jwpltx.js
```diff
@@ -37,10 +10,63 @@ window.jwpltx = window.jwpltx || {};
...
+  // Here we convert seconds watched to quantiles, the units accepted by the analytics service (res is 0 - 128)
+  function sendRemainingData() {
+    if (uri.pw === -1) {
+      clearLiveInterval();
+    } else {
+      const pw = getCurrentProgressQuantile(lastVp, uri.vd);
+      uri.pw = pw;
+    }
+
```
budzworthy, line 49: Should we clear the seek timeout here too?

public/jwpltx.js
AntonLantukh: Cleared it

public/jwpltx.js
```diff
@@ -37,10 +10,83 @@ window.jwpltx = window.jwpltx || {};
...
+      if (uri.pw === -1 && !liveInterval) {
+        setLiveInterval();
+      }
+      sendData('vs');
+    }, 1000);
+  }
+
+  // We clear timeout after complete event or when new seeking events are received
+  function clearSeekingTimeout() {
+    clearTimeout(liveInterval);
```
budzworthy:
```suggestion
    clearTimeout(seekingTimeout);
```

public/jwpltx.js
budzworthy: Does it potentially need to be cleared here too?

public/jwpltx.js
AntonLantukh: Good catch, thank you

public/jwpltx.js
AntonLantukh: Missed this place, thought that `sendRemainingData` had to cover it.

public/jwpltx.js
budzworthy: Actually github caught it. It was flagged as an assigned but unused variable. :-)

public/jwpltx.js
AntonLantukh: Fixed it.

budzworthy (approved):

marcovandeveen:
@pooja-jw will do some testing

PR #257, opened 2023-03-23
Dbudzinski/key

budzworthy:
## Description

This PR adds player key as an optional build or ini parameter. Also minor cleanup and doc improvements for ini and build params.

This PR solves [BC-461](https://jwplayer.atlassian.net/browse/BC-461).

### Steps completed:

- [x] Acceptance criteria met
- [ ] ~~Unit tests added~~
- [ ] ~~Translations added~~
- [x] UX tested
- [x] Reviewed own code

[BC-461]: https://jwplayer.atlassian.net/browse/BC-461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/PlayerContainer/PlayerContainer.tsx
```diff
@@ -56,6 +57,8 @@ const PlayerContainer: React.FC<Props> = ({
   // watch history
   const watchHistoryItem = useWatchHistoryStore((state) => (!!item && watchHistoryEnabled ? state.getItem(item) : undefined));
 
+  const { playerId, playerKey } = useSettingsStore((s) => s);
```
AntonLantukh: Maybe it is better to retrieve them in the Player component? Same for TrailerModal

docs/build-from-source.md
```diff
@@ -72,3 +72,53 @@ Production builds optimize code and minimize debug information, while developmen
...
+By default, the OTT Web App uses a global JWP OTT player that has its setting optimized to work properly with the app, but you can change it if you want to use a player directly from your own account.
+
+If you are using pre-compiled builds instead of building the code yourself, you can also set this value with the [playerId ini setting](initialization-file.md#playerid).
+Keep in mind, if the [playerId ini setting](initialization-file.md#playerid) is provided, it will be used even if the `APP_PLAYER_ID` environment variable is set.
+
+> Note: Be careful if using your own player, since some settings in the player can conflict with the way the player is used in the OTT Web App, causing unexpected behavior or UX experiences.
+
+> Note: If you opt to use the default global player, remember to provide your player key via the [APP_PLAYER_KEY](#APP_PLAYER_KEY) env variable or the [playerKey setting](initialization-file.md#playerkey) in the ini file.
+
+### APP_PLAYER_KEY
```
AntonLantukh: Maybe we can use `APP_PLAYER_LICENSE KEY` name? `APP_PLAYER_ID` and `APP_PLAYER_KEY` look really close to each other.

src/containers/PlayerContainer/PlayerContainer.tsx
budzworthy: It is! I actually had it there to start, but I eventually put it here to try to keep the convention of components being dumb and containers accessing stores. If it's in the component for example then the player component tests all have to be refactored to use real context with store data.

docs/build-from-source.md
budzworthy: That's a really good point. I was trying to avoid making the naming of a somewhat sensitive value too obvious, but I can see the risk of easy confusion. I'll rename it.

docs/build-from-source.md
```diff
@@ -72,3 +72,56 @@ Production builds optimize code and minimize debug information, while developmen
...
+By default, the OTT Web App uses a global JWP OTT player that has its setting optimized to work properly with the app, but you can change it if you want to use a player directly from your own account.
+
+If you are using pre-compiled builds instead of building the code yourself, you can also set this value with the [playerId ini setting](initialization-file.md#playerid).
+Keep in mind, if the [playerId ini setting](initialization-file.md#playerid) is provided, it will be used even if the `APP_PLAYER_ID` environment variable is set.
+
+> Note: Be careful if using your own player, since some settings in the player can conflict with the way the player is used in the OTT Web App, causing unexpected behavior or UX experiences.
+
+> Note: If you opt to use the default global player, remember to provide your player key via the [APP_PLAYER_LICENSE_KEY](#APP_PLAYER_LICENSE_KEY) env variable or the [playerLicenseKey setting](initialization-file.md#playerLicenseKey) in the ini file.
+
+### APP_PLAYER_LICENSE_KEY
```
AntonLantukh, line 109: One more question: will providing analytics token work the same way as with player license key? It can be automatically derived from the app config.

budzworthy: Ya, we do have some overlapping data, since the player key also contains the analytics token. The problem with the player key though is that it only applies for web, so having it built into the app makes more sense. Ideally we would just be able to set the analytics token on the player, but that doesn't work.

AntonLantukh (approved):

PR #258, opened 2023-03-28
fix: url signing service

kmitrovv:
## Description

Fixed an issue where if the old contentProtection json format is present, the newly created method for signing will fire instead of the old one.

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/hooks/useContentProtection.ts
```diff
@@ -24,21 +24,20 @@ const useContentProtection = <T>(
...
   const { data: token, isLoading } = useQuery(
     ['token', type, id, params],
     () => {
-      // if provider is JWP
-      if (jwp && configId && !!id && signingEnabled) {
-        return getJWPMediaToken(configId, id);
-      }
       // if provider is not JWP
-      if (!!id && !!host && drmEnabled && signingEnabled) {
+      if (!!id && !!host && !!drmPolicyId) {
```
budzworthy: I don't think drmPolicyID is required for the getMediaToken call to work. If it's blank, getMediaToken will do basic url signing, which we still want to support.

kmitrovv: aha, didn't know that basic signing (without drm) was present there  as well.

src/hooks/useContentProtection.ts
```diff
@@ -24,21 +24,20 @@ const useContentProtection = <T>(
   }));
   const host = signingConfig?.host;
   const drmPolicyId = contentProtection?.drm?.defaultPolicyId ?? signingConfig?.drmPolicyId;
-  const drmEnabled = !!drmPolicyId;
-  const signingEnabled = !!host || !!urlSigning;
+  const signingEnabled = !!urlSigning;
```
budzworthy:
```suggestion
  const signingEnabled = !!urlSigning || !!host || (!!drmPolicyId && !host);
```

kmitrovv: why is host here? This way we will need to include `signingEnabled` in the other providers condition and it will be true if jw integration is present as well.

budzworthy: If host is set for the legacy config, it means signing is enabled (there was not a separate custom param)

kmitrovv: but the signingEnabled in these changes is not present in the other  providers condition, or you've added host because of the `enabled` param of the query?

budzworthy (approved):

PR #269, opened 2023-05-05
fix: update inplayer.js version and types

mirovladimitrovski:
## Description

- Types available from Inplayer SDK `inplayer.js` were not consistent with some of its endpoints' response types, so a new version `3.13.9` (edit: we eventually got to version `3.13.12`) was published with this inconsistency fixed, making previous temporarily done TypeScript workarounds with `// @ts-ignore`, `as unknown as any` or similar techniques no longer necessary.

This PR solves https://www.notion.so/Update-Publish-SDK-types-and-refactor-the-usage-in-the-OTT-App-93077f3eb74b4dc581cc508f84d00259.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/src/services/inplayer.checkout.service.ts
+++ b/src/services/inplayer.checkout.service.ts
@@ -149,10 +149,8 @@ export const updateOrder: UpdateOrder = async ({ order, couponCode }) => {
 };
 
 export const directPostCardPayment = async (cardPaymentPayload: CardPaymentData, order: Order) => {
-  const number = parseInt(String(cardPaymentPayload.cardNumber).replace(/\s/g, ''), 10);
-
   const payload = {
-    number,
+    number: parseInt(String(cardPaymentPayload.cardNumber).replace(/\s/g, ''), 10),
     cardName: cardPaymentPayload.cardholderName,
     expMonth: cardPaymentPayload.cardExpMonth || '',
     expYear: cardPaymentPayload.cardExpYear || '',
--- a/src/services/inplayer.checkout.service.ts
+++ b/src/services/inplayer.checkout.service.ts
@@ -83,12 +83,9 @@ export const paymentWithPayPal: PaymentWithPayPal = async (payload) => {
       origin: `${window.location.origin}?u=waiting-for-payment`,
       accessFeeId: payload.order.id,
       paymentMethod: 2,
-      // TODO: add voucherCode in sdk types
-      // @ts-ignore
       voucherCode: payload.couponCode,
     });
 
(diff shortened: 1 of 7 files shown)


kmitrovv (approved):

AntonLantukh (approved):

src/services/inplayer.checkout.service.ts
```diff
@@ -164,6 +164,7 @@ export const directPostCardPayment = async (cardPaymentPayload: CardPaymentData,
 
   try {
     if (isSVODOffer(order)) {
+      // @ts-ignore
```
mirovladimitrovski: Update the request params type of `InPlayer.Subscription.createSubscription` by changing `number: number;` to `number: string;`. instead of using `ts-ignore`.

kmitrovv: that's the plan when we make sure this is the cause for the tests to fail.

yarn.lock
```diff
@@ -3449,9 +3449,9 @@ core-js@^3.19.2:
   integrity sha512-1t+2a/d2lppW1gkLXx3pKPVGbBdxXAkqztvWb1EJ8oF8O2gIGiytzflNiFEehYwVK/t2ryUsGBoOFFvNx95mbg==
 
 core-js@^3.8.2:
-  version "3.29.0"
-  resolved "https://registry.yarnpkg.com/core-js/-/core-js-3.29.0.tgz#0273e142b67761058bcde5615c503c7406b572d6"
-  integrity sha512-VG23vuEisJNkGl6XQmFJd3rEG/so/CNatqeE+7uZAwTSwFeB/qaO0be8xZYUNWprJ/GIwL8aMt9cj1kvbpTZhg==
+  version "3.30.2"
```
budzworthy: Can you run `yarn upgrade core-js@^3.19.2` so we don't have 2 versions of this?

budzworthy (approved):
One cleanup suggestion for yarn packages, but otherwise looks good. Nice to get rid of all those ignores and todos!

yarn.lock
naumovski-filip: hey, vlade has a day off today and asked me to update here. core-js is a dependency of inplayer.js (the SDK) so that was what actually needed to be bumped here, right after we bumped core-js in the SDK and published a new release of it. I just made a commit upgrading inplayer.js so you can take a look at that

PR #271, opened 2023-05-06
fix(e2e): fix tests

AntonLantukh:
- Fix tests. For some reason Chrome browser on Linux env started to provide opportunity to copy password (as bullets). 
- Update Playwright and CodeceptJS.
- Update Node.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/.github/workflows/codeceptjs.yml
+++ b/.github/workflows/codeceptjs.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [16.x]
+        node-version: [18.x]
         config: [desktop, mobile]
 
     steps:
@@ -21,7 +21,7 @@ jobs:
       - name: Install dependencies
         run: |
           yarn
-          npm install wait-on -g
+          yarn global add wait-on
       - name: Start preview server
         run: yarn start:test &
       - name: Run tests
--- a/.github/workflows/lhci.yml
+++ b/.github/workflows/lhci.yml
@@ -6,10 +6,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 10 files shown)


.github/workflows/codeceptjs.yml
```diff
@@ -4,12 +4,12 @@ on: [pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
```
budzworthy: So, I selfishly would like to keep these tests running properly on Ubuntu latest, otherwise they will also start failing for devs using linux laptops. I also wouldn't be surprised if the same chromium changes come to mac soon anyhow. Maybe we can simplify or drop the copy paste part of those tests? It seems like we are trying to test browser functionality more or less anyhow. If the input type is set, it's the browser controlling copy operations.

budzworthy (commented):
Thanks for updating all of the dependencies! One question about using ubuntu vs mac.

kmitrovv (approved):
LGTM.
Agree with Danny's input. I don't know about the others, but there are devs using linux from team Integrations and they will probably have issues.

.github/workflows/codeceptjs.yml
AntonLantukh: We already used mac for tests and liniting. Worked fine. Changed it to ubuntu now.
I agree that we could meet problems when launching it on ubuntu locally. However I also think that chrome support is better on mac and windows as they are just more widespread.
I think it makes sense to remove the copy / paste part of the tests and leave it up to the browser used.

AntonLantukh:
@dbudzins removed the copy/paste part and updated other actions to use ubuntu.

budzworthy (approved):

PR #275, opened 2023-05-17
feat(web app analytics): InPlayer user id

CarinaDraganJW:
## Description

- Use `id` field from the response from the JWP /  InPlayer user and use it to populate the `oaid` field with the analytics pings

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/src/hooks/useOttAnalytics.ts
+++ b/src/hooks/useOttAnalytics.ts
@@ -9,7 +9,7 @@ const useOttAnalytics = (item?: PlaylistItem, feedId: string = '') => {
   const user = useAccountStore((state) => state.user);
 
   // ott app user id (oaid)
-  const oaid: number | undefined = Number(user?.id);
+  const oaid: number | undefined = user?.id ? Number(user?.id) : undefined;
 
   const [player, setPlayer] = useState<jwplayer.JWPlayer | null>(null);
 
--- a/src/hooks/useOttAnalytics.ts
+++ b/src/hooks/useOttAnalytics.ts
@@ -9,9 +9,7 @@ const useOttAnalytics = (item?: PlaylistItem, feedId: string = '') => {
   const user = useAccountStore((state) => state.user);
 
   // oaid (Ott App User ID) for analytics
-  const userId = Number(user?.id);
-  const isLoggedIn = !!userId;
-  const oaid: number | undefined = isLoggedIn ? userId : undefined;
+  const oaid: number | undefined = Number(user?.id);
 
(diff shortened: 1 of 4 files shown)


src/hooks/useOttAnalytics.ts
```diff
@@ -1,12 +1,20 @@
...
+import jwtDecode from 'jwt-decode';
 
 import type { PlaylistItem } from '#types/playlist';
 import { useConfigStore } from '#src/stores/ConfigStore';
+import { useAccountStore } from '#src/stores/AccountStore';
+import type { DecodedJwtToken } from '#types/account';
 
 const useOttAnalytics = (item?: PlaylistItem, feedId: string = '') => {
   const analyticsToken = useConfigStore((s) => s.config.analyticsToken);
+  const auth = useAccountStore((state) => state.auth);
```
budzworthy: So, I think we can actually get the user ID from the inplayer sdk response to the login method, which is stored in accountStore.user.id (I think).

budzworthy (commented):
I think we can avoid having to decode the token. Any luck with adding the user id to the player analytics?

src/hooks/useOttAnalytics.ts
```diff
@@ -2,9 +2,15 @@ import { useCallback, useEffect, useState } from 'react';
...
 import type { PlaylistItem } from '#types/playlist';
 import { useConfigStore } from '#src/stores/ConfigStore';
+import { useAccountStore } from '#src/stores/AccountStore';
 
 const useOttAnalytics = (item?: PlaylistItem, feedId: string = '') => {
   const analyticsToken = useConfigStore((s) => s.config.analyticsToken);
+  const user = useAccountStore((state) => state.user);
+
+  // oaid (Ott App User ID) for analytics
+  const oaid: number | undefined = Number(user?.id);
```
AntonLantukh: Should we only send it for JW subscription model?

CarinaDraganJW: @dbudzins Do we need to check here the subscription model as well?

budzworthy: You mean for JW instead of cleeng?

AntonLantukh: Looks like Number(undefined) will give us a different result.

CarinaDraganJW: I added a check for it 🙂

AntonLantukh (approved):

PR #276, opened 2023-05-17
Feat / cleeng session stability improvements

ChristiaanScheermeijer:
## Description

This PR significantly changes how the Cleeng session is stored and refreshed. The following changes are made:

- Add Cleeng Auth Service
- Add Broadcaster and PromiseQueue utils
- Remove auth data from Zustand state
- Remove the `jwt` argument in Cleeng requests
- Remove token polling refresh mechanism in AccountController

The following problems are solved:

- Users are being logged out unexpectedly
  - Could happen when a request fails due to a network hiccup
  - Could happen when the browser closes the connection (closing tab/browser or for performance reasons)
  - Could happen when two or more tabs are open and invalidate each other by refreshing the token
- Refreshing the token every ~10 minutes even though no resources are requested

Good to know that the Cleeng JWT token is now refreshed on demand. When a request is made, the token is tested and refreshed before making the actual request.

When multiple resources are requested, only one refresh call will be made (even over multiple tabs in the same browser which share the same login session). The rest will be put into a PromiseQueue and will resolve when the refresh succeeds or fails. Via a BroadcastChannel (when supported) other browser tabs will be notified when the token is being refreshed or when the refresh is completed.

Let me know what you think!


ChristiaanScheermeijer:
@dbudzins @AntonLantukh Since InPlayer already uses Axios, I'm now also considering the fact that we can use Axios for the Cleeng requests and [axios-jwt](https://github.com/jetbridge/axios-jwt) for the refresh mechanism. 

Although the BroadcastChannel is not supported, this might be considered overkill anyway... 

What do you think?

src/services/cleeng.account.service.ts
```diff
@@ -31,8 +31,27 @@ import type {
...
+export const initialize = async (config: Config, logoutCallback: () => Promise<void>) => {
+  await cleengAuthService.initialize(!!config.integrations.cleeng?.useSandbox, logoutCallback);
+};
+
+export const getAuthData = async () => {
+  if (cleengAuthService.tokens) {
+    return {
+      jwt: cleengAuthService.tokens.accessToken,
+      refreshToken: cleengAuthService.tokens.refreshToken,
+    } as AuthData;
```
RCVZ, line 49: We can remove `customerToken` from the AuthData since it not being used anymore

src/services/cleeng.auth.service.ts
```diff
@@ -0,0 +1,318 @@
...
+  private sandbox = false;
+  tokens: Tokens | null = null;
+
+  /**
+   * The `logoutCallback` is a delegate that can be set to handle what should happen when the token is expired or not
+   * valid anymore. Since this service shouldn't care about the UI or Zustand state, a controller can use this delegate
+   * to update the UI and clear the logged in state. Alternatively, this could dispatch an event, but a delegate is
+   * easier to align with the InPlayer integration.
+   */
+  private logoutCallback: () => Promise<void> = () => Promise.resolve();
```
RCVZ: I would make it optional since it now looks like this is always set which can complicate debugging.

src/services/cleeng.auth.service.ts
```diff
@@ -0,0 +1,318 @@
...
+  };
+
+  /**
+   * This function is called when a broadcast message is received from another browser tab (same origin)
+   */
+  private handleBroadcastMessage = async (data: MessageData) => {
+    this.isRefreshing = data.action === 'refreshing';
+
+    if (data.tokens) {
+      await this.setTokens(data.tokens);
```
RCVZ, line 130: Has no effect `localStorage` is synchronous

src/services/cleeng.auth.service.ts
```diff
@@ -0,0 +1,318 @@
...
+   * Try to restore tokens from the storage and overwrite the current when they are newer.
+   */
+  async restoreTokensFromStorage() {
+    const tokensString = await getItem(AUTH_PERSIST_KEY);
+    let tokens;
+
+    if (typeof tokensString !== 'string') return;
+
+    try {
+      tokens = JSON.parse(tokensString || '');
```
RCVZ:
```suggestion
      tokens = JSON.parse(tokensString);
```

src/services/cleeng.account.service.ts
budzworthy, line 1: Seems like there are a lot of methods that are no longer used in this service because of the refactor. Can these be cleaned up?

budzworthy: At least `getFreshJwtToken`

src/services/cleeng.auth.service.ts
```diff
@@ -0,0 +1,318 @@
...
+   */
+  private logoutCallback: () => Promise<void> = () => Promise.resolve();
+
+  /**
+   * This function does the actual refresh of the access token by calling the `refreshToken` endpoint in the Cleeng
+   * service.
+   */
+  private getNewTokens: (tokens: Tokens) => Promise<Tokens | null> = async ({ refreshToken }) => {
+    try {
+      const { responseData: newTokens } = await accountService.refreshToken({ refreshToken }, this.sandbox);
```
budzworthy: Is `refreshToken` used anywhere else? If not, can it be moved into auth service? It would be nice not to have circular dependency between auth service and account service.

src/services/cleeng.service.ts
```diff
@@ -1,12 +1,16 @@
+import cleengAuthService from '#src/services/cleeng.auth.service';
+
 export const getBaseUrl = (sandbox: boolean) => (sandbox ? 'https://mediastore-sandbox.cleeng.com' : 'https://mediastore.cleeng.com');
 
-export const performRequest = async (sandbox: boolean, path: string = '/', method = 'GET', body?: string, jwt?: string) => {
+export const performRequest = async (sandbox: boolean, path: string = '/', method = 'GET', body?: string, authenticate = false) => {
```
budzworthy: Maybe nice to use a named option instead of the plain boolean for readability when called.
```suggestion
export const performRequest = async (sandbox: boolean, path: string = '/', method = 'GET', body?: string, options?: {authenticate?: boolean} = {authenticate: true}) => {
```

budzworthy (approved):
A couple small comments, but otherwise really cool stuff and well documented on the tricky parts. Nice!

budzworthy:
> @dbudzins @AntonLantukh Since InPlayer already uses Axios, I'm now also considering the fact that we can use Axios for the Cleeng requests and [axios-jwt](https://github.com/jetbridge/axios-jwt) for the refresh mechanism.
> 
> Although the BroadcastChannel is not supported, this might be considered overkill anyway...
> 
> What do you think?

It's an interesting thought.... converting everything over to axios would be time consuming and it increases our bundle size, but if there are big pros that outweigh the cons I'd be open to it.

src/services/cleeng.account.service.ts
ChristiaanScheermeijer: Sure!

src/services/cleeng.auth.service.ts
ChristiaanScheermeijer: Done!

src/services/cleeng.auth.service.ts
ChristiaanScheermeijer: That's correct!

src/services/cleeng.auth.service.ts
ChristiaanScheermeijer: Updated!

src/services/cleeng.auth.service.ts
ChristiaanScheermeijer: Nope, and you're entirely right. The auth service was initially setup as a generic service (that doesn't know about Cleeng), but now that it is coupled I moved the `refreshToken` function the cleeng.auth.service

src/services/cleeng.service.ts
ChristiaanScheermeijer: Yes, good call! I almost also added `sandbox` to the options... 😅

ChristiaanScheermeijer:
Thanks for the reviews @dbudzins and @RCVZ, I've updated the PR.

> It's an interesting thought.... converting everything over to axios would be time consuming and it increases our bundle size, but if there are big pros that outweigh the cons I'd be open to it.

Yes, that's true. Let's not do that in this PR 😃 The only pro is that we don't have to maintain the Cleeng API and auth service. But since that's already in now, there is no need...

src/services/cleeng.service.ts
budzworthy: Actually that would not be a terrible thing to do next time we touch this code

src/services/cleeng.account.service.ts
RCVZ: And now you don't have to cast the object to AuthData anymore

RCVZ (approved):

src/services/cleeng.account.service.ts
ChristiaanScheermeijer: These have been removed from the `Cleeng.account.service.ts` 👍

PR #279, opened 2023-05-22
feat(user): delete account data

naumovski-filip:
## Description

Notion card: https://www.notion.so/jwplayer/Viewer-delete-account-from-web-app-0cc99d39fe1d4a6e92d720afe9cee84b

### Steps completed:

- [X] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [X] UX tested
- [X] Reviewed own code


src/components/Account/Account.tsx
```diff
@@ -53,13 +53,14 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
...
-  const { customer, customerConsents, publisherConsents, canChangePasswordWithOldPassword, canExportAccountData } = useAccountStore(
-    ({ user, customerConsents, publisherConsents, canChangePasswordWithOldPassword, canExportAccountData }) => ({
+  const { customer, customerConsents, publisherConsents, canChangePasswordWithOldPassword, canExportAccountData, canDeleteAccount } = useAccountStore(
+    ({ user, customerConsents, publisherConsents, canChangePasswordWithOldPassword, canExportAccountData, canDeleteAccount }) => ({
       customer: user,
       customerConsents,
       publisherConsents,
       canChangePasswordWithOldPassword,
       canExportAccountData,
+      canDeleteAccount: canDeleteAccount,
```
AntonLantukh:
```suggestion
      canDeleteAccount,
```

src/components/Account/Account.tsx
```diff
@@ -288,6 +289,28 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
                 }),
               ]
             : []),
+          ...(canDeleteAccount
```
AntonLantukh: I would prefer doing it this way:  `canDeleteAccount && formSection({...`
And then we could filter the array using `filter(Boolean)` method.

src/components/DeleteAccountModal/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const removedU = removeQueryParam(location, 'u');
+    navigate(removedU.split('&confirmation')[0]);
+  };
+
+  return enteredPassword ? (
+    <div className={styles.formContainer}>
+      <h2 className={styles.heading}>{t('account.delete_account_title')}</h2>
+      <div className={styles.innerContainer}>
+        <p className={styles.paragraph}>{t('account.delete_account_modal_text_1')}</p>
+        <p className={styles.paragraph}>{t('account.delete_account_modal_text_2')}</p>
```
AntonLantukh: Can we use `\n` in the translation?

src/i18n/locales/en_US/user.json
```diff
@@ -29,7 +30,19 @@
...
     "export_data_success": "Please check your email.",
-    "export_data_error": "An error occurred while exporting your account data."
+    "export_data_error": "An error occurred while exporting your account data.",
+    "delete_account_title": "Delete account",
+    "delete_account_body": "Permanently delete your account and all of your content.",
+    "delete_account_modal_title": "Enter password to continue",
+    "delete_account_modal_placeholder": "Enter your password",
+    "delete_account_modal_text_1": "Please note that if you choose to proceed with this process, all of your personal data will be erased from our system. This includes any personal information you provided during the registration process, as well as any location and device information that we gathered during the authentication process. Additionally, all other personal information stored in various parts of our database will be deleted as well. However, please be aware that due to some legislations we are required to keep certain personal data as part of transaction and subscription records.",
```
AntonLantukh: I would suggest to group the keys this way:

```
 "confirmation_dialog": {
    "close": "Cancel",
    "confirm": "Yes"
  },
```

src/i18n/locales/en_US/user.json
```diff
@@ -29,7 +30,19 @@
...
+    "delete_account_modal_text_2": "Once the erasure is complete, all of your active accesses will be revoked, and any active login sessions, including this one, will be terminated. Your account information will no longer be accessible to our staff in any capacity.",
+    "delete_account_modal_text_warning_1": "All of your active subscriptions purchased through a web browser will be automatically canceled, in addition to the data erasure.",
+    "delete_account_modal_text_warning_2": "However, all active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription.",
+    "delete_account_modal_text_3": "Please be aware that this process cannot be undone, so proceed with caution.",
+    "delete_account_modal_text_4": "If you have any further questions or concerns, please do not hesitate to contact our support team.",
```
AntonLantukh: Can we replace numbers using subtitles?

src/services/inplayer.account.service.ts
```diff
@@ -313,18 +314,41 @@ export const updatePersonalShelves: UpdatePersonalShelves = async (payload) => {
...
+    if (code !== 200) {
+      throw new Error(message);
+    }
+    return {
+      errors: [],
+      responseData: {
+        message,
+        code,
+      },
+    };
```
AntonLantukh: Looks like part of the code we can reuse for several endpoints using a function. You can have a look on `getDataOrThrow` as an example.

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -96,9 +101,12 @@ const AccountModal = () => {
     }
   };
 
+  const shouldHideBanner = ['delete-account'].includes(view ?? '');
```
AntonLantukh:
```suggestion
  const shouldHideBanner = view === 'delete-account'
```

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -96,9 +101,12 @@ const AccountModal = () => {
...
+  const dialogSize = isConfirmation ? 'large' : 'small';
+
   return (
-    <Dialog open={!!viewParam} onClose={closeHandler}>
-      <div className={styles.banner}>{banner ? <img src={banner} alt="" /> : null}</div>
+    <Dialog size={dialogSize} open={!!viewParam} onClose={closeHandler}>
+      {!shouldHideBanner && banner && <div className={styles.banner}>{banner ? <img src={banner} alt="" /> : null}</div>}
```
AntonLantukh: Looks like we have already checked for a banner here

```suggestion
      {!shouldHideBanner && banner && <div className={styles.banner}><img src={banner} alt="" /></div>}
```

src/components/Dialog/Dialog.tsx
```diff
@@ -9,13 +9,14 @@ import ModalCloseButton from '#components/ModalCloseButton/ModalCloseButton';
...
+  size?: 'small' | 'large';
   children: React.ReactNode;
 };
 
-const Dialog: React.FC<Props> = ({ open, onClose, children }: Props) => {
+const Dialog: React.FC<Props> = ({ open, onClose, size = 'small', children }: Props) => {
   return (
     <Modal open={open} onClose={onClose} AnimationComponent={Slide}>
-      <div className={styles.dialog}>
+      <div className={`${styles.dialog}${size === 'large' ? ` ${styles.largeDialog}` : ''}`}>
```
AntonLantukh:
```suggestion
      <div className={classNames(styles.dialog, size === 'large' && styles.largeDialog}>
```

src/components/Dialog/Dialog.module.scss
```diff
@@ -18,3 +18,7 @@
     display: none;
   }
 }
+
+.largeDialog {
+  max-width: 768px;
```
AntonLantukh, line 26: Will it also work fine for mobile devices?

src/styles/_variables.scss
```diff
@@ -17,6 +17,11 @@ $blue: #2b3b57 !default;
 $green: #5aae4a !default;
 $orange: #e9a95b !default;
 
+$dark-danger-main: #ec0041 !default;
```
AntonLantukh: Maybe it is better to use a color names, what do you think (red-dark here)?
Btw, $red and $dark-danger-main look really close to each other, do we need both of them?

src/i18n/locales/en_US/user.json
naumovski-filip: Not sure which subtitles to use here, open to any suggestions

src/components/DeleteAccountModal/DeleteAccountModal.tsx
naumovski-filip: I had issues with trying to use `\n` and break tags in the translation so decided to split them up in separate html elements

src/containers/AccountModal/AccountModal.tsx
naumovski-filip: Implemented this as an array as there will be another case where hiding the banner is necessary (edit card details) which is also in review

src/styles/_variables.scss
naumovski-filip: I was going off the figma names when naming the colors and deciding on their hex values, I think it's best to keep them the same name as in figma so it's more obvious if they need to be reused in the future.

AntonLantukh: The problem is that it doesn't fit into the current paradigm where we have color names and not states colors correspond to. I would ask to change the figma names then to have the same approach.

src/i18n/locales/en_US/user.json
AntonLantukh: @dbudzins should we translate all new keys to Spanish now?

src/components/Account/Account.tsx
AntonLantukh: I meant to add `filter(Boolean)` to the whole array of sections:

```
[
formSection({...,
canDeleteAccount && formSection({...
].filter(Boolean)
```

src/i18n/locales/en_US/user.json
```diff
@@ -31,18 +31,22 @@
...
+        "text_warning_1": "All of your active subscriptions purchased through a web browser will be automatically canceled, in addition to the data erasure.",
+        "text_warning_2": "However, all active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription.",
+        "text_3": "Please be aware that this process cannot be undone, so proceed with caution.",
+        "text_4": "If you have any further questions or concerns, please do not hesitate to contact our support team.",
+        "warning_1": "All of your active subscriptions purchased through a web browser will be automatically canceled, in addition to the data erasure.",
```
AntonLantukh: Do we need both `text_warning_1` and `warning_1` as well as `text_warning_2` and `warning_2`?

AntonLantukh: Looks like we don't need this paragraph at all as the second one already includes this information

src/i18n/locales/en_US/user.json
```diff
@@ -31,18 +31,22 @@
...
+        "warning_2": "However, all active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription."
```
AntonLantukh: I would suggest to remove the first "However":

```
"warning_2": "All active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of...."
```

public/locales/en/user.json
```diff
@@ -29,7 +26,27 @@
...
+    "export_data_body": "Export all of your personal account data to <strong>{{email}}</strong> as an attachment.",
+    "export_data_success": "Please check your email.",
+    "export_data_error": "An error occurred while exporting your account data.",
+    "delete_account": {
+      "title": "Delete account",
+      "body": "Permanently delete your account and all of your content.",
+      "modal": {
+        "title": "Enter password to continue",
+        "placeholder": "Enter your password",
+        "text_1": "Please note that if you choose to proceed with this process, all of your personal data will be erased from our system. This includes any personal information you provided during the registration process, as well as any location and device information that we gathered during the authentication process. Additionally, all other personal information stored in various parts of our database will be deleted as well. However, please be aware that due to some legislations we are required to keep certain personal data as part of transaction and subscription records.",
```
AntonLantukh:
```suggestion
        "text_data_erasure": "Please note that if you choose to proceed with this process, all of your personal data will be erased from our system. This includes any personal information you provided during the registration process, as well as any location and device information that we gathered during the authentication process. Additionally, all other personal information stored in various parts of our database will be deleted as well. However, please be aware that due to some legislations we are required to keep certain personal data as part of transaction and subscription records.",
```

public/locales/en/user.json
```diff
@@ -29,7 +26,27 @@
...
+        "text_2": "Once the erasure is complete, all of your active accesses will be revoked, and any active login sessions, including this one, will be terminated. Your account information will no longer be accessible to our staff in any capacity.",
```
AntonLantukh:
```suggestion
        "text_revoked_access": "Once the erasure is complete, all of your active accesses will be revoked, and any active login sessions, including this one, will be terminated. Your account information will no longer be accessible to our staff in any capacity.",
```

public/locales/en/user.json
```diff
@@ -29,7 +26,27 @@
...
+        "text_warning_1": "All of your active subscriptions purchased through a web browser will be automatically canceled, in addition to the data erasure.",
+        "text_warning_2": "However, all active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription.",
+        "text_3": "Please be aware that this process cannot be undone, so proceed with caution.",
+        "text_4": "If you have any further questions or concerns, please do not hesitate to contact our support team.",
```
AntonLantukh:
```suggestion
        "text_contacts": "If you have any further questions or concerns, please do not hesitate to contact our support team.",
```

public/locales/en/user.json
```diff
@@ -29,7 +26,27 @@
...
+        "text_2": "Once the erasure is complete, all of your active accesses will be revoked, and any active login sessions, including this one, will be terminated. Your account information will no longer be accessible to our staff in any capacity.",
+        "text_warning_1": "All of your active subscriptions purchased through a web browser will be automatically canceled, in addition to the data erasure.",
+        "text_warning_2": "However, all active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription.",
+        "text_3": "Please be aware that this process cannot be undone, so proceed with caution.",
```
AntonLantukh:
```suggestion
        "text_cant_be_undone": "Please be aware that this process cannot be undone, so proceed with caution.",
```

src/i18n/locales/en_US/user.json
budzworthy: Yes, I'm asking around what the best way we can go about this is... We can start with a machine translation and then hopefully just fix up. I'm trying to get clarity on if we want to do this with each commit though...

src/components/Button/Button.tsx
```diff
@@ -8,7 +8,7 @@ import Spinner from '#components/Spinner/Spinner';
 
 type Color = 'default' | 'primary';
 
-type Variant = 'contained' | 'outlined' | 'text';
+type Variant = 'contained' | 'outlined' | 'text' | 'danger' | 'delete';
```
ChristiaanScheermeijer, line 11: Shouldn't `danger` and `delete` be a color instead of a button variant?

src/components/Dialog/Dialog.tsx
```diff
@@ -9,13 +10,14 @@ import ModalCloseButton from '#components/ModalCloseButton/ModalCloseButton';
...
+  size?: 'small' | 'large';
   children: React.ReactNode;
 };
 
-const Dialog: React.FC<Props> = ({ open, onClose, children }: Props) => {
+const Dialog: React.FC<Props> = ({ open, onClose, size = 'small', children }: Props) => {
   return (
     <Modal open={open} onClose={onClose} AnimationComponent={Slide}>
-      <div className={styles.dialog}>
+      <div className={classNames(styles.dialog, size === 'large' && styles.largeDialog)}>
```
ChristiaanScheermeijer: To make this expendable (and a bit cleaner), you could use the size as a key as well. 

```suggestion
      <div className={classNames(styles.dialog, styles[size])}>
```

In the CSS module you would have:

```scss
.small {
  max-width: 450px; // extracted from .dialog selector
}

.large {
  max-width: 768px;
}
```

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -60,7 +58,12 @@ const AccountModal = () => {
     if (!!viewParam && !loading && !user && !isPublicView) {
       toLogin();
     }
-  }, [viewParam, loading, user, isPublicView, toLogin]);
+  }, [viewParam, navigate, location, loading, isPublicView, user, toLogin]);
+
+  const closeHandler = () => {
+    const removedU = removeQueryParam(location, 'u');
+    navigate(isConfirmation ? removedU.split('&confirmation')[0] : removedU);
+  };
```
ChristiaanScheermeijer: This function is now reactive and will be new for every render. This will cause problems in other modals because it is used in `useEffects` and `useCallback`, which are expected to be run only once. Therefor the function was wrapped with the `useEventCallback` hook.

Instead of checking the existence of the confirmation query param, I suggest always removing it without checking. Perhaps, you can even create a `removeQueryParams` utility that accepts multiple keys like so:

```ts
navigate(removeQueryParams(location, ['u', 'confirmation']));
```

Although, it might be weird that the AccountModal is aware of a query param that is set by a child component...

src/components/Button/Button.tsx
naumovski-filip: I put them as variants as they additionally have different styles for the outlines (both of them) and `outlined` is already a variant, so having them as colors seemed like it could potentially overlap styles with each other and wasn't that intuitive to me.

src/containers/AccountModal/AccountModal.tsx
naumovski-filip: I'll go ahead with implementing a removeQueryParams with multiple parameters as arguments. Regarding AccountModal being aware of child query params, I do agree that the approach is not the best. I went with it because of the AccountModal/Dialog implementation and global state management solutions all seemed overkill. If you have a better idea I would be more than happy to reimplement it in this case

src/components/Account/Account.tsx
naumovski-filip: I can't seem to make this work, typescript is complaining about false still being a potential value. Could you try and suggest it as a commit exactly?

src/containers/AccountModal/AccountModal.tsx
ChristiaanScheermeijer: It's better now since you won't need the `isConfirmation` variable. The `useEventCallback` is still missing, though.

I see two alternate solutions to refactor the `confirmation` query param.

1) By using local state in the `DeleteAccountModal` indicating it is the confirm step. We probably won't need to have a direct route to the confirmation state anyway
2) The second option is to create a separate `u` key for the confirmation step. You can still use the same component, but you must pass the password variable to the next modal. One solution for this is to use the same component instance and check the `u` parameter in the modal or pass the state as a prop like so;

```ts
switch (view) {
      //...
      case 'delete-account':
      case 'delete-account-confirmation':
        return <DeleteAccountModal confirmation={view === 'delete-account-confirmation'} />;
}
```

naumovski-filip: Right, updated with `useEventCallback` in the latest commit. Regarding the alternate solutions, how would we make the `AccountModal`/`Dialog` components aware of the child's state then though? We need the info about which step the child component is at (in `AccountModal`) to handle the size of the `Dialog` component (at least with the current `Dialog` implementation).

In the second example, we'd still have to update the query params from `DeleteAccountModal`, and we'd still have to use the updated query parameter to be able to change the `Dialog` size.

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -60,7 +58,11 @@ const AccountModal = () => {
     if (!!viewParam && !loading && !user && !isPublicView) {
       toLogin();
     }
-  }, [viewParam, loading, user, isPublicView, toLogin]);
+  }, [viewParam, navigate, location, loading, isPublicView, user, toLogin]);
```
ChristiaanScheermeijer: I think these deps are incorrect after fixing the merge conflicts. The `navigate` and `location` deps aren't needed anymore

src/containers/AccountModal/AccountModal.tsx
ChristiaanScheermeijer: Yes, but the AccountModal is aware of the different views that it can render, so theoretically, you could use the same check as done for the `shouldHideBanner`;

```ts
  const shouldHideBanner = ['delete-account'].includes(view ?? '');
  const dialogSize = ['delete-account-confirmation'].includes(view ?? '') ? 'large' : 'small';  
```

The AccountModal only knows that there is a `delete-account-confirmation` view, but doesn't need to know more. 

From the `DeleteAccountModal` component (which should be a container because it uses a controller), you can navigate to the `?u=delete-account-confirmation` step after filling in the password.

public/locales/en/user.json
```diff
@@ -29,7 +26,24 @@
...
+        "text_revoked_access": "Once the erasure is complete, all of your active accesses will be revoked, and any active login sessions, including this one, will be terminated. Your account information will no longer be accessible to our staff in any capacity.",
+        "text_cant_be_undone": "Please be aware that this process cannot be undone, so proceed with caution.",
+        "text_contacts": "If you have any further questions or concerns, please do not hesitate to contact our support team.",
+        "warning_1": "All active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription."
```
AntonLantukh:
```suggestion
        "warning": "All active subscriptions that were purchased through a web browser will be automatically canceled upon erasure of personal data. However, if you purchased a subscription through iOS, Android, Roku, or Amazon apps, please be aware that it will not be canceled automatically. In this case, you will need to follow the usual subscription cancelation steps for each respective platform to manually cancel the subscription."
```

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -117,9 +121,12 @@ const AccountModal = () => {
...
   };
 
+  const shouldHideBanner = ['delete-account', 'delete-account-confirmation'].includes(view ?? '');
+  const dialogSize = ['delete-account-confirmation'].includes(view ?? '') ? 'large' : 'small';
+
   return (
-    <Dialog open={!!viewParam} onClose={closeHandler}>
-      <div className={styles.banner}>{banner ? <img src={banner} alt="" /> : null}</div>
+    <Dialog size={dialogSize} open={!!viewParam} onClose={closeHandler}>
+      {!shouldHideBanner && banner && <div className={styles.banner}>{<img src={banner} alt="" />}</div>}
```
AntonLantukh: We could probably change it to `shouldHShowBanner` not to negate the expression always.

AntonLantukh (commented):
Looks good to me! Thanks for addressing all the comments! Just waiting for the final decision on translations.

src/components/DeleteAccountModal/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const [enteredPassword, setEnteredPassword] = useState<string>('');
+
+  const deleteAccount = useMutation(deleteAccountData, {
+    onSuccess: async () => {
+      await logout();
+      navigate('/');
+    },
+    onError: () => {
+      setEnteredPassword('');
```
ChristiaanScheermeijer, line 29: When entering a wrong password this modal seems to be closed and reset. Is this clear enough for the user that the entered password was wrong?

src/components/DeleteAccountModal/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+    initialValues,
+    () => {
+      setEnteredPassword(values.password);
+      navigate(addQueryParam(location, 'u', 'delete-account-confirmation'));
+    },
+    validationSchema,
+  );
+
+  const handleCancel = () => {
+    navigate(removeMultipleQueryParams(location, ['confirmation', 'u']));
```
ChristiaanScheermeijer: This is not needed anymore, right?

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -60,7 +57,11 @@ const AccountModal = () => {
     if (!!viewParam && !loading && !user && !isPublicView) {
       toLogin();
     }
-  }, [viewParam, loading, user, isPublicView, toLogin]);
+  }, [viewParam, loading, isPublicView, user, toLogin]);
+
+  const closeHandler = useEventCallback(() => {
+    navigate(removeMultipleQueryParams(location, ['u', 'confirmation']));
```
ChristiaanScheermeijer: This is also not needed anymore?

budzworthy:
> Looks good to me! Thanks for addressing all the comments! Just waiting for the final decision on translations.

Since it's someone outside of engineering who is essentially volunteering, I'd suggest we don't put too much pressure onto them by making them a critical path for web app commits. Let's do the thing of opening 2 pr's (1 for logic with English and 1 for (spanish) translations) for now.

AntonLantukh (approved):

src/components/DeleteAccountModal/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    validationSchema,
+  );
+
+  const handleCancel = () => {
+    navigate(removeQueryParam(location, 'u'));
+  };
+
+  if (deleteAccount.isError) {
+    return <Alert open isSuccess={false} onClose={handleCancel} message={t('account.delete_account.error')} />;
+  }
```
ChristiaanScheermeijer: Minor and opinionated, but perhaps sending the user back to the first step is more user-friendly instead of closing the modal? 

```suggestion
  const handleCancel = () => {
    navigate(removeQueryParam(location, 'u', { replace: true }));
  };
  
  const handleAlertClose = () => {
    // reset password (perhaps a state reset of the form is also needed)
    setEnteredPassword('');
    // go back one step
    navigate(addQueryParam(location, 'u', 'delete-account'), { replace: true });
  };

  if (deleteAccount.isError) {
    return <Alert open isSuccess={false} onClose={handleCancel} message={t('account.delete_account.error')} />;
  }
```

src/components/DeleteAccountModal/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const validationSchema: SchemaOf<DeleteAccountFormData> = object().shape({
+    password: string().required(t('login.field_required')),
+  });
+  const initialValues: DeleteAccountFormData = { password: '' };
+  const { handleSubmit, handleChange, values, errors } = useForm(
+    initialValues,
+    () => {
+      setEnteredPassword(values.password);
+      navigate(addQueryParam(location, 'u', 'delete-account-confirmation'));
```
ChristiaanScheermeijer: Not sure if you want to be able to use the back button. I think it ends up in a mismatch between the `u` parameter and the presented modal because the confirmation modal is shown based on the `enteredPassword` state.

```suggestion
      navigate(addQueryParam(location, 'u', 'delete-account-confirmation', { replace: true }));
```

ChristiaanScheermeijer:
Thanks for the changes and fixes @naumovski-filip! Nice work 😄 

I have one concern left regarding the history state and the modal. I think the URL needs to be replaced instead of pushed to prevent the user from going back to the state by using the browser's back/forward buttons. 

@dbudzins what do you think?

naumovski-filip:
> Thanks for the changes and fixes @naumovski-filip! Nice work 😄
> 
> I have one concern left regarding the history state and the modal. I think the URL needs to be replaced instead of pushed to prevent the user from going back to the state by using the browser's back/forward buttons.
> 
> @dbudzins what do you think?

@ChristiaanScheermeijer Take a look at how I approached it in https://github.com/jwplayer/ott-web-app/pull/279/commits/eacd3b7575cc0d6cbf891c7ed7943c9b4b66ab76 and tell me what you think. I think I tested out every case, back button, redirect after error, writing the url without entering the password etc.

budzworthy:
> > Thanks for the changes and fixes @naumovski-filip! Nice work smile
> > I have one concern left regarding the history state and the modal. I think the URL needs to be replaced instead of pushed to prevent the user from going back to the state by using the browser's back/forward buttons.
> > @dbudzins what do you think?
> 
> @ChristiaanScheermeijer Take a look at how I approached it in [eacd3b7](https://github.com/jwplayer/ott-web-app/commit/eacd3b7575cc0d6cbf891c7ed7943c9b4b66ab76) and tell me what you think. I think I tested out every case, back button, redirect after error, writing the url without entering the password etc.

I agree with Christiaan on this one. It's something that's bugged me with other modals in the app that need to be fixed and somewhere there's a backlog ticket for it. When a modal is shown and then closed, I shouldn't see it pop up again when I use the back button (it's not enough to just clear the modal state.) I think using `replace` achieves this.

darkoatanasovski (approved):

AntonLantukh:
@naumovski-filip could you please create a PR with Spanish translations?

PR #280, opened 2023-05-23
feat(project): Update React to version 18.2.0

MelissaDTH:
## Description

This PR updates React version from 17.0.2 to 18.2.0, the following changes are made:

- Updated React version to 18.2.0 in package.json and related dependencies ([React Changelog](https://github.com/facebook/react/blob/main/CHANGELOG.md))
   - Adds necessary typings throughout the project to align with the updated React version
- Explicitly declared the children prop in the components that use React.FC type, as the children prop has been removed from React.FC in React 18

Note:
This PR is still in draft since the update to React 18 is not completely compatible with the current TileSlider component. There's an issue with that component when you use the right or left controls, the transition is not smooth. Further work is required to resolve this compatibility issue


test/testUtils.tsx
```diff
@@ -34,10 +34,13 @@ export const wrapper = ({ children }: WrapperProps) => (
...
-  vi.stubGlobal('location', {
-    pathname: path,
-    assign: vi.fn(),
-  });
+  const location = new URL(`https://www.jwplayer.com/${path}`) as unknown as Location;
+
+  if ('location' in globalThis) {
+    // @ts-ignore
+    delete globalThis.location;
+    globalThis.location = location;
```
budzworthy: Should this line be outside the 'if'?

AntonLantukh: Removed this condition

budzworthy (approved):

PR #281, opened 2023-05-23
Feat / update payment details

RCVZ:
## Description

This PR adds:

- update Adyen payment logic to checkout controller
- update Paypal payment logic to checkout controller
- show Paypal as active payment detail(instead of always showing creditcard)
- payment-method modal
- payment-method form

This makes it possible to update your payment details with cleeng as your provider.
Paypal -> Paypal
CC -> CC
Paypal -> CC
CC -> Paypal

### DIFF
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -27,6 +27,7 @@ type Props = {
   panelHeaderClassName?: string;
   onShowAllTransactionsClick?: () => void;
   showAllTransactions: boolean;
+  canUpdatePaymentMethod: boolean;
   canRenewSubscription?: boolean;
 };
 
@@ -42,6 +43,7 @@ const Payment = ({
   onShowAllTransactionsClick,
   showAllTransactions,
   canRenewSubscription = false,
+  canUpdatePaymentMethod,
 }: Props): JSX.Element => {
   const { t } = useTranslation(['user', 'account']);
   const hiddenTransactionsCount = transactions ? transactions?.length - VISIBLE_TRANSACTIONS : 0;
@@ -140,7 +142,9 @@ const Payment = ({
             <p>{!isLoading && t('user:payment.no_payment_methods')}</p>
           </div>
         )}
-        <Button label={t('user:payment.update_payment_details')} type="button" onClick={() => navigate(addQueryParam(location, 'u', 'payment-method'))} />
+        {canUpdatePaymentMethod && (
+          <Button label={t('user:payment.update_payment_details')} type="button" onClick={() => navigate(addQueryParam(location, 'u', 'payment-method'))} />
+        )}
       </div>
       <div className={panelClassName}>
         <div className={panelHeaderClassName}>
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -12,6 +12,7 @@ import { formatDate, formatPrice } from '#src/utils/formatting';
 import { addQueryParam } from '#src/utils/location';
 import type { PaymentDetail, Subscription, Transaction } from '#types/subscription';
 import type { AccessModel } from '#types/Config';
+import PayPal from '#src/icons/PayPal';
 
 const VISIBLE_TRANSACTIONS = 4;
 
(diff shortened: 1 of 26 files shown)


src/components/PaymentMethod/PaymentMethod.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const navigate = useNavigate();
+  const activePayment = useAccountStore((state) => state.activePayment);
+  const currentPaymentId = activePayment?.id;
+  const [paymentError, setPaymentError] = useState<string | undefined>(undefined);
+  const [paymentMethodId, setPaymentMethodId] = useState<number | undefined>(parsedPaymentMethodId);
+
+  const paymentMethods = useCheckoutStore((state) => state.paymentMethods);
+
+  const [processing, setProcessing] = useState<boolean>(false);
```
ChristiaanScheermeijer: This component shouldn't access stores and do API calls (or it should be a container)

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -92,6 +93,9 @@ const AccountModal = () => {
         return <CancelSubscription />;
       case 'renew-subscription':
         return <RenewSubscription />;
+      case 'payment-method':
+      case 'payment-method-success':
+        return <PaymentMethod onCloseButtonClick={closeHandler} />;
```
ChristiaanScheermeijer: I think the component/container should be called `UpdatePaymentMethod` or `ChangePaymentMethod`. It now implies that is a single payment method and it can also be used in the checkout flow.

src/stores/CheckoutController.ts
```diff
@@ -134,19 +136,28 @@ export const directPostCardPayment = async (cardPaymentPayload: CardPaymentData)
...
+      const getOrderIdIfInitialPayment = () => {
+        if (!isInitialPayment) return undefined;
+
+        if (!order) {
+          throw new Error('No order created');
+        }
+
+        return order.id;
+      };
+
```
ChristiaanScheermeijer: This can be simplified:

```suggestion
      const orderId = order?.id;
      
      if (isInitialPayment && !orderId) throw new Error('There is no order to pay for');

```

src/components/Payment/Payment.tsx
```diff
@@ -117,22 +118,29 @@ const Payment = ({
...
+                <TextField label={t('user:payment.security_code')} value={'******'} editing={false} />
+              </div>
+            </div>
+          )
         ) : (
           <div>
             <p>{!isLoading && t('user:payment.no_payment_methods')}</p>
           </div>
         )}
+        <Button label={t('user:payment.update_payment_details')} type="button" onClick={() => navigate(addQueryParam(location, 'u', 'payment-method'))} />
```
ChristiaanScheermeijer: Should this button be hidden when not supported? Perhaps add a feature flag to the account service like currently done here:

https://github.com/jwplayer/ott-web-app/blob/ac37dfb714b8fcafcb30561490504ffde666b868/src/services/cleeng.account.service.ts#L224-L230

```ts
export const canUpdatePaymentDetails = true;
```

And then set that feature flag in state here:

https://github.com/jwplayer/ott-web-app/blob/ac37dfb714b8fcafcb30561490504ffde666b868/src/stores/AccountController.ts#L76-L81

src/stores/CheckoutController.ts
```diff
@@ -245,3 +256,84 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
     });
   });
 };
+
+export const updatePaypalPaymentMethod = async (
```
ChristiaanScheermeijer: Minor, but the rest of the code uses Pay**P**al

```suggestion
export const updatePayPalPaymentMethod = async (
```

src/stores/CheckoutController.ts
```diff
@@ -245,3 +256,84 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
...
+  paymentMethodId: number,
+  currentPaymentId?: number,
+) => {
+  return await useAccount(async ({ auth: { jwt } }) => {
+    return await useService(async ({ checkoutService, sandbox = true, authProviderId }) => {
+      if (!authProviderId) throw new Error('auth provider is not configured');
+      if (!checkoutService) throw new Error('checkout service is not available');
+      if (!('updatePaymentMethodWithPaypal' in checkoutService)) throw new Error('updatePaymentMethodWithPaypal is not available in checkout service');
+
+      const response = await checkoutService.updatePaymentMethodWithPaypal(
```
ChristiaanScheermeijer: Same here

```suggestion
      const response = await checkoutService.updatePaymentMethodWithPayPal(
```

ChristiaanScheermeijer (commented):
The PR looks good @RCVZ, I have some minor feedback 😄

src/components/PaymentMethod/PaymentMethod.tsx
RCVZ: It seems that JW has let go of this convention. But I will change it.

src/containers/AccountModal/AccountModal.tsx
RCVZ: yes your are right, will change it.

src/containers/UpdatePaymentMethod/UpdatePaymentMethod.tsx
```diff
@@ -2,12 +2,12 @@ import React, { useEffect, useState } from 'react';
...
 import { useCheckoutStore } from '../../stores/CheckoutStore';
-import LoadingOverlay from '../LoadingOverlay/LoadingOverlay';
-import PayPal from '../PayPal/PayPal';
+import LoadingOverlay from '../../components/LoadingOverlay/LoadingOverlay';
+import PayPal from '../../components/PayPal/PayPal';
 import { useAccountStore } from '../../stores/AccountStore';
 import PaymentMethodForm from '../../components/PaymentMethodForm/PaymentMethodForm';
 import useQueryParam from '../../hooks/useQueryParam';
-import { getPaymentMethods, updatePaypalPaymentMethod } from '../../stores/CheckoutController';
+import { getPaymentMethods, updatePayPalPaymentMethod } from '../../stores/CheckoutController';
```
ChristiaanScheermeijer: Can you also fix the imports? 😄

ChristiaanScheermeijer (approved):
Nice @RCVZ, let's merge!

src/stores/CheckoutController.ts
RCVZ: What was I thinking......

src/containers/UpdatePaymentMethod/UpdatePaymentMethod.tsx
ChristiaanScheermeijer: We can do this later

src/components/PaymentMethod/PaymentMethod.tsx
budzworthy: Indeed, we haven't enforced that convention

PR #283, opened 2023-05-24
Feat / add functionality to switch offers

MelissaDTH:
## Description

This feature introduces the ability to upgrade or downgrade a subscription. The PR is currently in draft because of a Cleeng issue and missing tests, but feedback is already appreciated! It includes the following changes:

- Added services that enables the offer switching (e.g. _cleeng.checkout.services.ts, CheckoutController_)
- Updated the Payment page to allow for changing the subscription with a 'Change subscription' button
- Updates the ChooseOffer modal to support switching between available offers

#### Additional Notes

- I tested all code changes with a Symphony-config and it works as expected there. However, with the JW-config (_ozylzc5m_) set, switching offers does not correctly work. From the /availability response by Cleeng, I keep getting the status 'SWITCH_IN_PROGRESS'. I consulted with Robin and it is suspected that there is a bug in the offers we get for JW from Cleeng.
  - I've created a ticket for this issue for Cleeng (#832593).

- I've added a function `determineSwitchDirection()`. It defaults to 'upgrade' when the period is not 'month' or 'year'. Please me know of this is not desired
<br />

#### The updated ChooseOffer modal, in which only the upgradable/downgrade option is displayed:
Screenshot 2023-05-23 at 22 29 15


src/components/Payment/Payment.tsx
```diff
@@ -98,8 +111,17 @@ const Payment = ({
...
               </div>
+              <Button
+                className={styles.upgradeSubscription}
+                label={t('user:payment.change_subscription')}
+                onClick={onUpgradeSubscriptionClick}
+                fullWidth={isMobile}
+                color="primary"
+                disabled={!offersAvailable}
+                data-testid="change-subscription-button"
+              />
```
ChristiaanScheermeijer: Maybe instead of disabling the button, hide it when there are no switch offers available?

src/stores/CheckoutController.ts
```diff
@@ -172,3 +179,43 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
     });
   });
 };
+
+export const getSubscriptionSwitches = async (): Promise<unknown> => {
+  const { cleengSandbox } = useConfigStore.getState().getCleengData();
+  const { user, subscription, auth } = useAccountStore.getState();
+
+  if (!user || !subscription || !auth) return;
+  const response = await checkoutService.getSubscriptionSwitches(
```
ChristiaanScheermeijer: The `InPlayer` service doesn't contain the `getSubscriptionSwitches` method. I think we don't have to throw an error here, but early return so that the `offerSwitches` state will remain empty.

```suggestion
  if (!('getSubscriptionSwitches' in checkoutService) return;
  
  const response = await checkoutService.getSubscriptionSwitches(
```

In the Payment component, we can hide the "upgrade" button so that it still works with the InPlayer integration.

src/stores/CheckoutController.ts
```diff
@@ -172,3 +179,43 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
     });
   });
 };
+
+export const getSubscriptionSwitches = async (): Promise<unknown> => {
+  const { cleengSandbox } = useConfigStore.getState().getCleengData();
+  const { user, subscription, auth } = useAccountStore.getState();
```
ChristiaanScheermeijer: Instead of using the stores/services directly, we should use the `use*` helpers like the above methods:

```suggestion
  return await useAccount(async ({ auth: { jwt } }) => {
    return await useService(async ({ checkoutService, sandbox = true, authProviderId }) => {
```

ChristiaanScheermeijer: These methods already check if the user is logged in and use the correct `checkoutService` for the configured integration.

src/stores/CheckoutController.ts
```diff
@@ -172,3 +179,43 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
...
+
+  // Sort offers for proper ordering in "Choose Offer" modal when applicable
+  const offerSwitches = offers.sort((a, b) => a?.responseData.offerPrice - b?.responseData.offerPrice).map((item) => item.responseData);
+  useCheckoutStore.setState({ offerSwitches });
+};
+
+export const switchSubscription = async (toOfferId: string, switchDirection: 'upgrade' | 'downgrade') => {
+  const { cleengSandbox } = useConfigStore.getState().getCleengData();
+  const { user, subscription, auth } = useAccountStore.getState();
+  if (!user || !subscription || !auth) return;
```
ChristiaanScheermeijer: The same here:

```suggestion
  return await useAccount(async ({ auth: { jwt } }) => {
    return await useService(async ({ checkoutService, sandbox = true, authProviderId }) => {
```

src/stores/CheckoutController.ts
```diff
@@ -172,3 +179,43 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
...
+
+  const payload = { toOfferId, customerId: user.id, offerId: subscription.offerId, switchDirection: switchDirection, jwt: auth.jwt };
```
ChristiaanScheermeijer: This method does need to throw when the method is not defined in the `checkoutService`:

```suggestion
  if (!('switchSubscription' in checkoutService)) throw new Error('SwitchSubscription not supported');
  
  const payload = { toOfferId, customerId: user.id, offerId: subscription.offerId, switchDirection: switchDirection, jwt: auth.jwt };
```

src/stores/CheckoutController.ts
```diff
@@ -172,3 +179,43 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
...
+
+  await checkoutService.switchSubscription(payload, cleengSandbox, auth.jwt);
+
+  reloadActiveSubscription();
+  queryClient.invalidateQueries(['subscriptions']);
```
ChristiaanScheermeijer: We don't have an `useQuery` with a `['subscriptions']` cache key in the OTT Web App, so this invalidate isn't needed.

types/checkout.d.ts
```diff
@@ -217,4 +249,6 @@ export type GetPaymentMethods = EmptyAuthServiceRequest<PaymentMethodResponse>;
 export type PaymentWithoutDetails = AuthServiceRequest<PaymentWithoutDetailsPayload, Payment>;
 export type PaymentWithAdyen = AuthServiceRequest<PaymentWithAdyenPayload, Payment>;
 export type PaymentWithPayPal = AuthServiceRequest<PaymentWithPayPalPayload, PaymentWithPayPalResponse>;
+export type GetSubscriptionSwitches = CleengAuthRequest<GetSubscriptionSwitchesPayload, GetSubscriptionSwitchesReponse>;
+export type SwitchSubscription = CleengAuthRequest<SwitchSubscriptionPayload, SwitchSubscriptionResponse>;
```
ChristiaanScheermeijer:
```suggestion
export type GetSubscriptionSwitches = AuthServiceRequest<GetSubscriptionSwitchesPayload, GetSubscriptionSwitchesReponse>;
export type SwitchSubscription = AuthServiceRequest<SwitchSubscriptionPayload, SwitchSubscriptionResponse>;
```

src/stores/CheckoutController.ts
MelissaDTH: Thank you! I initially used the `useAccount` and `useService ` methods, but because the InPlayer service didn't contain the getSubscriptionSwitches method, I proceeded this way. But with the check that you suggested (`if (!('switchSubscription' in checkoutService))`), it now works like a charm.

src/stores/CheckoutController.ts
```diff
@@ -172,3 +175,55 @@ export const paypalPayment = async (successUrl: string, cancelUrl: string, error
...
+      if (!authProviderId) throw new Error('auth provider is not configured');
+      if (!checkoutService) throw new Error('checkout service is not available');
+      if (!('switchSubscription' in checkoutService)) throw new Error('SwitchSubscription not supported');
+      const { subscription } = useAccountStore.getState();
+
+      if (!subscription) return;
+
+      const SwitchSubscriptionPayload = { toOfferId, customerId: customerId, offerId: subscription.offerId, switchDirection: switchDirection, jwt };
+
+      await checkoutService.switchSubscription(SwitchSubscriptionPayload, sandbox, jwt);
```
ChristiaanScheermeijer: `jwt` is not needed in the payload, right?

```suggestion
      const switchSubscriptionPayload = { toOfferId, customerId: customerId, offerId: subscription.offerId, switchDirection };

      await checkoutService.switchSubscription(switchSubscriptionPayload, sandbox, jwt);
```

MelissaDTH: Ah yes, I see that you are right. It's not necessary

ChristiaanScheermeijer (approved):
Nice!

RCVZ (approved):

PR #284, opened 2023-05-24
Feat / checkout and payment improvements

MelissaDTH:
## Description

This PR addresses two bugs:

- Remove pricing on payment page for granted subscriptions
   - With `activeSubscription.period` it's determined whether or not to show the pricing. This was necessary because the data is mocked for a "granted" subscription (see `formatGrantedSubscription`), which caused a wrong currency symbol to appear for countries that don't use 'EUR'.
- Hide tax and payment provider fees when not applicable
   - If there is no tax and/or no payment provider fee, they're now hidden during the checkout process

### DIFF
--- a/src/components/CheckoutForm/CheckoutForm.module.scss
+++ b/src/components/CheckoutForm/CheckoutForm.module.scss
@@ -115,7 +115,7 @@
   }
 
   tfoot {
-    tr:first-child {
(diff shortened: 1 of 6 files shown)


ChristiaanScheermeijer (approved):
Looks good!

RCVZ (approved):

PR #285, opened 2023-05-25
Feat / open and view invoices

MelissaDTH:
## Description

This PR adds functionality to view invoices on the Payment page. The PR It includes the following changes:

- The `responseData` from the receipt is encoded in a base64 format, so it can be decoded with `window.atob` in the `handleShowReceiptClick()` function
  - That same function sets the opener property of `newWindow` to null to prevent the newly opened window from having access to the original window that opened it.
  
**Note**
- Known issue: In the testing environment, if the pricing of the invoice is 0, the invoice will open an empty tab. This is not the case in production (correct me if I'm wrong @ChristiaanScheermeijer)
- Tests will be added later
  
  
Screenshot 2023-05-25 at 08 16 33

### DIFF
--- a/src/components/Payment/Payment.tsx
+++ b/src/components/Payment/Payment.tsx
@@ -1,9 +1,10 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLocation, useNavigate } from 'react-router-dom';
 
 import useBreakpoint, { Breakpoint } from '../../hooks/useBreakpoint';
-import { getSubscriptionSwitches } from '../../stores/CheckoutController';
+import IconButton from '../IconButton/IconButton';
+import ExternalLink from '../../icons/ExternalLink';
 
 import styles from './Payment.module.scss';
 
@@ -27,6 +28,7 @@ type Props = {
   customer: Customer;
   isLoading: boolean;
   offerSwitchesAvailable: boolean;
+  onShowReceiptClick: (transactionId: string) => void;
   panelClassName?: string;
   panelHeaderClassName?: string;
   onShowAllTransactionsClick?: () => void;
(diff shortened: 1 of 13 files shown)


src/stores/AccountController.ts
```diff
@@ -501,3 +503,12 @@ const validateInputLength = (values: { firstName: string; lastName: string }) =>
...
   return errors;
 };
+
+export const getReceipt = async (transactionId: string) => {
+  const { cleengSandbox } = useConfigStore.getState().getCleengData();
+  const { auth } = useAccountStore.getState();
+
+  if (!auth) throw new Error('user is not logged in');
+
+  return (await subscriptionService.fetchReceipt({ transactionId }, cleengSandbox, auth.jwt)).responseData;
```
ChristiaanScheermeijer: - Can you use the `useAccount` and `useService` helpers here?
- We should check if the subscriptionService contains the `fetchReceipt` method and throw if not defined (InPlayer integration)

src/components/Payment/Payment.tsx
```diff
@@ -149,11 +154,16 @@ const Payment = ({
...
-                </p>
+                <div className={styles.transactionDetails}>
+                  <p>
+                    {transaction.transactionId}
+                    <br />
+                    {formatDate(transaction.transactionDate)}
+                  </p>
+                  <IconButton aria-label={t('user:payment.show_receipt')} onClick={() => !isLoading && onShowReceiptClick(transaction.transactionId)}>
+                    <ExternalLink />
+                  </IconButton>
```
ChristiaanScheermeijer: Do you think we can see if we can show the invoice? For example, using a service feature flag?

I also remember that we should hide the button when the invoice price is zero/granted because these invoices don't work.

ChristiaanScheermeijer (changes requested):
Nice PR! Could you rebase/cherry-pick it on the JWP branch instead of the latest version of develop? This PR now also contains changes from the develop branch.

PR #288, opened 2023-05-25
Fix / TVOD infinite loop and render optimizations

ChristiaanScheermeijer:
## Description

This PR fixes an infinite render loop when opening the Choose Offer modal for a JWP TVOD platform. This was caused by the `clientOffers` being constructed each render while being a dependency in the `useOffers` hook.

Besides that, for a TVOD platform, the `clientOffer` resulted in `['null']` when not set. This causes the `useOffers` hook to request non-existing offers for JWP and Cleeng.

The other commit fixes some unnecessary renders in the Choose Offer form caused by depending on the `location` or `navigate` (which also depends on `location`) reactive variables.

### DIFF
--- a/src/containers/AccountModal/AccountModal.tsx
+++ b/src/containers/AccountModal/AccountModal.tsx
@@ -24,6 +24,7 @@ import { addQueryParam, removeQueryParam } from '#src/utils/location';
 import FinalizePayment from '#components/FinalizePayment/FinalizePayment';
 import WaitingForPayment from '#components/WaitingForPayment/WaitingForPayment';
 import UpdatePaymentMethod from '#src/containers/UpdatePaymentMethod/UpdatePaymentMethod';
+import useEventCallback from '#src/hooks/useEventCallback';
 
 const PUBLIC_VIEWS = ['login', 'create-account', 'forgot-password', 'reset-password', 'send-confirmation', 'edit-password'];
 
@@ -41,6 +42,14 @@ const AccountModal = () => {
   } = config;
   const isPublicView = viewParam && PUBLIC_VIEWS.includes(viewParam);
 
+  const toLogin = useEventCallback(() => {
+    navigate(addQueryParam(location, 'u', 'login'));
+  });
+
+  const closeHandler = useEventCallback(() => {
+    navigate(removeQueryParam(location, 'u'));
+  });
+
   useEffect(() => {
     // make sure the last view is rendered even when the modal gets closed
     if (viewParam) setView(viewParam);
@@ -48,13 +57,9 @@ const AccountModal = () => {
 
   useEffect(() => {
     if (!!viewParam && !loading && !user && !isPublicView) {
-      navigate(addQueryParam(location, 'u', 'login'));
+      toLogin();
     }
-  }, [viewParam, navigate, location, loading, user, isPublicView]);
-
-  const closeHandler = () => {
-    navigate(removeQueryParam(location, 'u'));
-  };
+  }, [viewParam, loading, user, isPublicView, toLogin]);
 
   const renderForm = () => {
     if (!user && loading && !isPublicView) {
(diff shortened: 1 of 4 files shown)


src/hooks/useEntitlement.ts
```diff
@@ -55,6 +57,7 @@ const useEntitlement: UseEntitlement = (playlistItem) => {
       queryFn: () => checkoutService?.getEntitlements({ offerId }, sandbox),
       enabled: !!playlistItem && !!user && !!offerId && !isPreEntitled,
       refetchOnMount: 'always' as const,
+      notifyOnChangeProps,
```
budzworthy, line 60: Is there a reason we can't just use string literals here?
```suggestion
      ['data', 'isLoading'],
```

src/hooks/useClientIntegration.ts
```diff
@@ -11,10 +13,24 @@ const useClientIntegration = () => {
...
   const clientId = isJWP ? integrations.jwp?.clientId : integrations.cleeng?.id;
-  const clientOffers = isJWP ? [`${integrations.jwp?.assetId}`] : [`${integrations.cleeng?.monthlyOffer}`, `${integrations.cleeng?.yearlyOffer}`];
+
+  // prevent infinite updates when `clientOffers` is used in as dependency
+  const clientOffers = useMemo(() => {
+    if (isJWP && !!integrations.jwp?.assetId) {
+      return [`${integrations.jwp.assetId}`];
+    }
+
+    if (isCleeng && integrations.cleeng?.monthlyOffer && integrations.cleeng?.yearlyOffer) {
```
budzworthy, line 27: Should we try to support monthly OR yearly as well?

src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -36,45 +49,46 @@ const ChooseOffer = () => {
     offerId: defaultOfferId,
   };
 
-  const determineSwitchDirection = () => {
-    const currentPeriod = subscription?.period;
+  const closeModal = useEventCallback((replace = false) => {
+    navigate(removeQueryParam(location, 'u'), { replace });
+  });
```
budzworthy, line 54: Do you need dependencies on navigate and location here and below?

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -41,20 +42,24 @@ const AccountModal = () => {
   } = config;
   const isPublicView = viewParam && PUBLIC_VIEWS.includes(viewParam);
 
+  const toLogin = useEventCallback(() => {
```
budzworthy, line 45: Same question if we need dependencies on navigate and location?

src/hooks/useEntitlement.ts
ChristiaanScheermeijer: Without the `as const` modifiers, the type results in `string[]` and is not accepted by react-query.

If I use `['data', 'isLoading'] as const` the type results in `readonly ['data', 'isLoading']` and is also not accepted by react-query.

This does work but is not much cleaner IMHO

```ts
const notifyOnChangeProps: QueryObserverOptions['notifyOnChangeProps'] = ['data', 'isLoading'];
```

src/containers/AccountModal/forms/ChooseOffer.tsx
ChristiaanScheermeijer: It should be in a regular `useCallback`, but not in the [useEventCallback](https://github.com/reactjs/rfcs/pull/220).

We've found that when a modal has an effect, memo or callback on the location/navigate properties, all will trigger an update when closing the modal because of the location update.

Because this is hard to discover, we've chosen to wrap these dependant functions with `useEventCallback` to ensure functions depending will not update unnecessarily (and possibly cause issues).

For example, this `useEffect`:

```ts
useEffect(() => {
    // close auth modal when there are no offers defined in the config
    if (!isLoading && !offers.length) {
      closeModal(true);
    }
  }, [isLoading, offers, closeModal]);
```

This effect closes the modal (only once) based on the reactive variables `isLoading` and `offers`. The `closeModal` is still considered a reactive variable, but never updates because wrapped in `useEventCallback`.

The problem without the `useEventCallback` is that we can't ensure that this effect is only running when `isLoading` or `offers` are changed. Because it will also run when `location` changes (after a navigate).

RCVZ (approved):

PR #293, opened 2023-05-26
Feat / BCL live events

RCVZ:
## Description

This PR adds BCL live events to the ott webapp.

- adds Media event screen
- adds useLiveEvent hook
- adds Media event metadata to videodetail
- remove the live tag from media items that are not currently "live"

**Summary of how the media events work**
Suppose a playlist contains a scheduled or live media item the playlist will be refetched every 30 seconds. For playlists that only have live items the custom property refect can be added to the playlist in the dashboard. A media item that contains the VCH.EvenState and VCH.ScheduledStart will be handled as a live event. The media event screen will refetch the media item in intervals of 30 seconds if it’s either a live event or a scheduled event.

A media item will be considered LIVE and playable if either VCH.EventState = LIVE_PUBLISHED  or VCH.ScheduledStart > now. A media item will be considered VOD and playable if  VCH.EventState = INSTANT_VOD || VOD_PUBLIC or VCH.ScheduledEnd < now. A media item will be considered SCHEDULED and not playable if VCH.EventState = PRE_LIVE || LIVE_UNPUBLISHED

### DIFF
--- a/src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
+++ b/src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
@@ -30,7 +30,7 @@ import InlinePlayer from '#src/containers/InlinePlayer/InlinePlayer';
 import StatusIcon from '#components/StatusIcon/StatusIcon';
 
 const MediaEvent: ScreenComponent<PlaylistItem> = ({ data, isLoading }) => {
-  const { t } = useTranslation('video');
+  const { t, i18n } = useTranslation('video');
 
   const [playTrailer, setPlayTrailer] = useState<boolean>(false);
   const breakpoint = useBreakpoint();
@@ -89,8 +89,8 @@ const MediaEvent: ScreenComponent<PlaylistItem> = ({ data, isLoading }) => {
 
   const primaryMetadata = (
     <>
-      <StatusIcon mediaStatus={liveEvent.mediaStatus} />
-      {formatLiveEventMetaString(data)}
+      <StatusIcon mediaStatus={data.mediaStatus} />
+      {formatLiveEventMetaString(data, i18n.language)}
     </>
   );
 
--- a/src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
+++ b/src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
@@ -1,0 +1,192 @@
+import React, { useCallback, useEffect, useState } from 'react';
+import { useNavigate, useParams } from 'react-router';
+import { Helmet } from 'react-helmet';
+import { useTranslation } from 'react-i18next';
+import shallow from 'zustand/shallow';
+
+import { useLiveEvent } from '#src/hooks/useLiveEvent';
+import { MediaStatus } from '#src/utils/liveEvent';
+import type { ScreenComponent } from '#types/screens';
+import VideoLayout from '#components/VideoLayout/VideoLayout';
+import { isLocked } from '#src/utils/entitlements';
+import { formatLiveEventMetaString, mediaURL } from '#src/utils/formatting';
+import type { PlaylistItem } from '#types/playlist';
+import useMedia from '#src/hooks/useMedia';
+import { generateMovieJSONLD } from '#src/utils/structuredData';
+import { useConfigStore } from '#src/stores/ConfigStore';
+import { useAccountStore } from '#src/stores/AccountStore';
+import usePlaylist from '#src/hooks/usePlaylist';
+import useEntitlement from '#src/hooks/useEntitlement';
+import StartWatchingButton from '#src/containers/StartWatchingButton/StartWatchingButton';
+import Cinema from '#src/containers/Cinema/Cinema';
+import useBreakpoint, { Breakpoint } from '#src/hooks/useBreakpoint';
+import TrailerModal from '#src/containers/TrailerModal/TrailerModal';
+import ShareButton from '#components/ShareButton/ShareButton';
+import FavoriteButton from '#src/containers/FavoriteButton/FavoriteButton';
+import PlayTrailer from '#src/icons/PlayTrailer';
+import Button from '#components/Button/Button';
+import useQueryParam from '#src/hooks/useQueryParam';
+import InlinePlayer from '#src/containers/InlinePlayer/InlinePlayer';
+import StatusIcon from '#components/StatusIcon/StatusIcon';
+
+const MediaEvent: ScreenComponent<PlaylistItem> = ({ data, isLoading }) => {
+  const { t } = useTranslation('video');
+
+  const [playTrailer, setPlayTrailer] = useState<boolean>(false);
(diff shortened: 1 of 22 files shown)


src/services/api.service.ts
```diff
@@ -29,6 +33,9 @@ export const transformMediaItem = (item: PlaylistItem, playlist?: Playlist) => {
     backgroundImage: generateImageData(config, ImageProperty.BACKGROUND, item),
     channelLogoImage: generateImageData(config, ImageProperty.CHANNEL_LOGO, item),
     mediaOffers: item.productIds ? filterMediaOffers(offerKeys, item.productIds) : undefined,
+    scheduledStart: item['VCH.ScheduledStart'] ? parseISO(item['VCH.ScheduledStart'] as string) : undefined,
+    scheduledEnd: item['VCH.ScheduledEnd'] ? parseISO(item['VCH.ScheduledEnd'] as string) : undefined,
+    mediaStatus: getMediaStatusFromEventState(item),
   };
```
ChristiaanScheermeijer: The `getMediaStatusFromEventState` utility makes use of the transformed `scheduledStart` / `scheduledEnd` properties which are not present yet in `item`.

Should we skip precalculating the `mediaStatus` here and use the `useLiveEvent` hook and `getMediaStatusFromEventState` util instead so we always get the most current media status?

Alternatively, we can split the transformed item into two steps:

```ts
  const tempItem = {
    ...item,
    backgroundImage: generateImageData(config, ImageProperty.BACKGROUND, item),
    channelLogoImage: generateImageData(config, ImageProperty.CHANNEL_LOGO, item),
    mediaOffers: item.productIds ? filterMediaOffers(offerKeys, item.productIds) : undefined,
    scheduledStart: item['VCH.ScheduledStart'] ? parseISO(item['VCH.ScheduledStart'] as string) : undefined,
    scheduledEnd: item['VCH.ScheduledEnd'] ? parseISO(item['VCH.ScheduledEnd'] as string) : undefined,
  };

  // mediaStatus uses the `scheduledStart` and `scheduledEnd` properties
  return {
    ...tempItem,
    mediaStatus: getMediaStatusFromEventState(tempItem),
  };
```

src/utils/formatting.ts
```diff
@@ -110,3 +112,27 @@ export const formatSeriesMetaString = (seasonNumber?: string, episodeNumber?: st
...
+export const formatVideoSchedule = (scheduledStart?: Date, scheduledEnd?: Date) => {
+  if (!scheduledStart) {
+    return '';
+  }
+
+  if (!scheduledEnd) {
+    return format(scheduledStart, 'PPP • p');
+  }
+
+  return `${format(scheduledStart, 'PPP • p')} - ${format(scheduledEnd, 'p')}`;
```
ChristiaanScheermeijer: I think we should use the `Intl.DateTimeFormat` here instead of `date-fns` i18n functions. Because we don't know which languages are supported and this prevents publishers from adding the date fns `locale` file.

```ts
const formatLocalizedDate = (date: Date, locale: string) => new Intl.DateTimeFormat(locale, { day: 'numeric', month: 'numeric', year: 'numeric' }).format(date);

const formatLocalizedTime = (date: Date, locale: string) => new Intl.DateTimeFormat(locale, { hours: 'numeric', minutes: 'numeric', seconds: 'numeric' }).format(date);

const formatLocalizedDateTime = (date: Date, locale: string, separator = ' ') => `${formatLocalizedDate(date, locale)}${separator}${formatLocalizedTime(date, locale)}`;
```

```suggestion
  if (!scheduledEnd) {
    return formatLocalizedDateTime(scheduledStart, locale, '•');
  }

  return `${formatLocalizedDateTime(scheduledStart, locale, '•')} - ${formatLocalizedTime(scheduledEnd, 'locale')}`;
```

src/services/api.service.ts
RCVZ: I need the mediaStatus in the shelf's so I went for the second solution

src/utils/formatting.ts
```diff
@@ -125,14 +123,21 @@ export const formatLiveEventMetaString = (media: PlaylistItem) => {
...
-  return `${format(scheduledStart, 'PPP • p')} - ${format(scheduledEnd, 'p')}`;
+  return `${formatLocalizedDateTime(scheduledStart, locale, '•')} - ${formatLocalizedTime(scheduledEnd, 'locale')}`;
 };
+
+const formatLocalizedDate = (date: Date, locale: string) => new Intl.DateTimeFormat(locale, { day: 'numeric', month: 'long', year: 'numeric' }).format(date);
+
+const formatLocalizedTime = (date: Date, locale: string) => new Intl.DateTimeFormat(locale, { hour: 'numeric', minute: 'numeric' }).format(date);
+
+const formatLocalizedDateTime = (date: Date, locale: string, separator = ' ') =>
+  `${formatLocalizedDate(date, locale)} ${separator} ${formatLocalizedTime(date, locale)}`;
```
ChristiaanScheermeijer:
```ts
const formatLocalizedDateTime = (date: Date, locale: string, separator = '') =>
  `${formatLocalizedDate(date, locale)} ${separator} ${formatLocalizedTime(date, locale)}`;
```
or

```ts
const formatLocalizedDateTime = (date: Date, locale: string, separator = ' ') =>
  `${formatLocalizedDate(date, locale)}${separator}${formatLocalizedTime(date, locale)}`;
```

Now, without a separator, it render three whitespaces between the date and time:

`30-05-2023   09:09`

I think the last example is the most flexible because you can either render a single whitespace or a character padded with whitespaces. If you don't like the separator being a whitespace or a whitespace padded character, we could also do something like this:

```ts
const formatLocalizedDateTime = (date: Date, locale: string, separator = '') =>
  [formatLocalizedDate(date, locale), separator, formatLocalizedTime(date, locale)].filter(Boolean).join();

formatLocalizedDateTime(new Date(), 'nl', '');  // 30-05-2023 09:09
formatLocalizedDateTime(new Date(), 'nl', '•'); // 30-05-2023 • 09:09
```

src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
```diff
@@ -0,0 +1,192 @@
...
+    (document.scrollingElement || document.body).scroll({ top: 0 });
+  }, [id]);
+
+  // UI
+  const pageTitle = `${data.title} - ${siteName}`;
+  const canonicalUrl = data ? `${window.location.origin}${mediaURL(data)}` : window.location.href;
+
+  const primaryMetadata = (
+    <>
+      <StatusIcon mediaStatus={data.mediaStatus} />
```
ChristiaanScheermeijer, line 92: This now uses the `data.mediaStatus` and only updates when the media item gets refreshed and transformed. I don't think this will ever switch when the state changes when already on this page, right?

RCVZ: It will update. The usePlaylist will update the media item.

ChristiaanScheermeijer (approved):

PR #296, opened 2023-05-30
Feat/social login buttons

naumovski-filip:
## Description

Notion card: https://www.notion.so/jwplayer/Implement-buttons-for-social-login-register-ce5ba133c6f94881b11dfd177afce07e?pvs=4

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/components/Root/Root.tsx
```diff
@@ -45,6 +48,32 @@ const Root: FC = () => {
...
+      const auth = getAuthParams();
+      if (!auth.token || !auth.refreshToken || !auth.expires) {
+        return;
+      }
+      InPlayer.Account.setToken(auth.token, auth.refreshToken, auth.expires);
+      await getAccount();
+      navigate(location.pathname);
+    };
+    getUserInfo();
+  }, [getAuthParams, location.pathname, navigate, userLoading]);
```
ChristiaanScheermeijer: I would move this "bootstrap" logic into the `inplayer.account.service.ts` initialize method:

https://github.com/jwplayer/ott-web-app/blob/develop/src/services/inplayer.account.service.ts#L39

You're now creating a hard dependency on the InPlayer integration in the root file. I also think the `useEffect` will run unnecessarily because it depends on the `location.pathname`, which constantly changes when navigating the app.

src/components/SocialButton/SocialButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  useEffect(() => {
+    const getIcon = async () => {
+      const iconSvg = await (import(`../../assets/icons/${variant}.svg`) as Promise<{ default: string }>);
+      setIcon(iconSvg.default);
+    };
+    getIcon();
+  }, [variant]);
+
+  return (
+    <a href={href} className={styles.socialButtonContainer}>
```
ChristiaanScheermeijer:
```suggestion
  // t('login_via_facebook');
  // t('login_via_google');
  // t('login_via_twitter');
  return (
    <a href={href} className={styles.socialButtonContainer} aria-label={t(`login_via_${variant}`)}>
```

src/components/Root/Root.tsx
naumovski-filip: https://github.com/jwplayer/ott-web-app/pull/296/commits/e3db47ae3c132424d64615355189d4ce1d8e7462

Did you mean like this?

ChristiaanScheermeijer: Yes, very nice! The only feedback I have is that the translation comments are missing. When using variables in `t()` calls, i18next parser wont be able to extract all translation keys when running `$ yarn i18next`. 

You can test this by running the i18next script and observe the changes in the translation files. I suggest to always use the i18next script instead of adding translation keys manually to ensure the structure is always correct and all languages are in sync.

public/locales/en/account.json
```diff
@@ -84,7 +84,10 @@
     "sign_up": "Sign up",
     "view_password": "View password",
     "wrong_combination": "Incorrect email/password combination",
-    "wrong_email": "Please check your email and try again."
+    "wrong_email": "Please check your email and try again.",
```
AntonLantukh: I suggest to create a separate PR for these keys in Spanish as agreed.

src/components/Root/Root.tsx
naumovski-filip: Got it, thank you. I wasn't aware of the i18next script and was doing all the translation keys manually, updated in https://github.com/jwplayer/ott-web-app/pull/296/commits/bbc18f267e65a540ff2f3d4cfc3eee8f5be27f90.

darkoatanasovski (approved):

PR #301, opened 2023-05-31
Feat / switch subscription enhancements

ChristiaanScheermeijer:
## Description

This PR adds (or fixes) two parts of the switch subscription mechanism.

- Show confirmation modal when the subscription switch succeeds (or error when it has failed)
- Show the pending offer information on the payments page (screenshot)

We can consider making this look nicer in the future. 

**Payments page (pending switch)**
image

**Success modal**
image

**Error modal**
image


RCVZ (approved):

MelissaDTH:
Nice, works like a charm with an annual subscription! I've also tested this with a monthly subscription, and in that flow we also state that the change will be visible on the next billing date. But this 'upgrade' is set to change immediately, so after refreshing the page (which is a bug, we shouldn't need to refresh the page), I see that I have an annual subscription. 

So, we should not show the success modal when we upgrade I believe.

ChristiaanScheermeijer:
@MelissaDTH, thanks for the review! As discussed via Slack, I've created a separate state for a "pending" switch that also shows a different message in the modal.

We've concluded that a switch is not visible in the subscription immediately, and even a 2-second delay is not enough. Because this isn't a "fixed" delay, we agreed to set the delay higher on purpose.

MelissaDTH (approved):
LGTM! Performing an upgrade or downgrade now both have the correct flow and it's clear for the user what's happening.

ChristiaanScheermeijer:
The e2e tests are failing due to a change in the series API, this isn't related to this PR

ChristiaanScheermeijer:
For later reference, the modal has been updated to show two variants. This is because the publisher can decide which switch strategy to use in the Cleeng dashboard. The options are; immediate switch (instantly updates the subscription) or deferred switch (switches after the next billing date).

**Deferred switch**
image (11)

**Immediate switch**
image (10)

PR #304, opened 2023-06-01
feat(series): add season titles as filters

AntonLantukh:
## Description

- add season titles to filters
- fix content type casing

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


AntonLantukh:
@dbudzins We use `season_title` now instead of a `season_number` for cases when it is present. I suggest not to check the presence of `Season` in the string and just let customers decide if the logic needs to be changed. Legacy series flow remains the same.

budzworthy:
@AntonLantukh I'm not sure if this bug is from the PR, but I noticed when I select All Seasons, Season 1 stays selected too. Probably just an indexing thing...
image

AntonLantukh:
@dbudzins thank you! Should be fixed

src/pages/LegacySeries/utils.ts
```diff
@@ -6,11 +8,17 @@ import type { EpisodeMetadata } from '#types/series';
...
+export const getFiltersFromSeries = (playlist: Playlist | undefined, t: TFunction): string[] => {
   // Old series doesn't have sorting supported and just aggregates all episodes in one playlist
   // So we need to sort the playlist manually based on the selected filter (season).
   return (playlist?.playlist || [])
-    .reduce((filters: string[], item) => (item.seasonNumber && filters.includes(item.seasonNumber) ? filters : filters.concat(item.seasonNumber || '')), [])
+    .reduce(
+      (filters: string[], item) =>
+        item.seasonNumber && filters.includes(item.seasonNumber)
+          ? filters
+          : filters.concat(item.seasonNumber ? `${t('season_prefix')} ${item.seasonNumber}` : ''),
```
budzworthy: `season_prefix` should probably change to `season_number_filter_template` with a value of "Season {{season_number}}" or something so we have the ability to support more than "Season #" especially considering that this format won't work for some languages.

src/utils/series.ts
```diff
@@ -4,8 +4,10 @@ import { getEpisodes, getSeasonWithEpisodes } from '#src/services/api.service';
 /**
  * Get an array of options for a season filter
  */
-export const getFiltersFromSeries = (series: Series | undefined): string[] => {
-  return series && 'seasons' in series ? series.seasons.map((season) => String(season.season_number)) : [];
+export const getFiltersFromSeries = (series: Series | undefined): { label: string; value: string }[] => {
+  return series && 'seasons' in series
+    ? series.seasons.map((season) => ({ label: season.season_title || String(season.season_number), value: String(season.season_number) }))
```
budzworthy, line 9: Shouldn't we include the seasons prefix here since it was removed from the filterValuePrefix prop?

src/pages/LegacySeries/utils.ts
```diff
@@ -6,11 +8,17 @@ import type { EpisodeMetadata } from '#types/series';
 /**
  * Get an array of options for a season filter
  */
-export const getFiltersFromSeries = (playlist: Playlist | undefined): string[] => {
+export const getFiltersFromSeries = (playlist: Playlist | undefined, t: TFunction): string[] => {
```
budzworthy: You don't need to pass `t` in. You can either get it from an import like this: https://github.com/jwplayer/ott-web-app/blob/develop/src/services/config.service.ts#L114
or you can convert this to a hook if there's any concern that the filters won't regenerate when the language get's switched.

src/pages/LegacySeries/utils.ts
AntonLantukh: Nice catch, thank you!

src/utils/series.ts
AntonLantukh: I suggest we don't have any prefix and just show what a customer chose for a season title. By default it is 'Season X'.

budzworthy: Good point... will season_title ever end up empty unless they try really hard to do so? If that's the case, then I'm not worried about this I guess.

budzworthy (approved):

src/utils/series.ts
AntonLantukh: It can be empty now in UI, but we can make it `required` field. Looks like we should.

budzworthy: Let's leave it as optional and emptyable for now. Some customers may prefer the language support of unnamed seasons instead of customizing titles themselves. In this case, we should use `season_number_filter_template` here.

Honestly, I can't think of an example where a series I've watched has meaningful names for seasons. Maybe for e-learning type courses it would make sense, but normal content not so much.

AntonLantukh: I saw that sometimes customers use years for season titles. I think e-learning is also a valid example.

budzworthy: OK, I'm fine with however it's setup now. We can always iterate on it.

PR #307, opened 2023-06-06
Feat/simultaneous login notification

borkopetrovicc:
## Description

https://www.notion.so/Handle-simultaneous-logins-on-the-WebApp-03ef3f51408a450e8adf5be9dd0bd03b

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/components/Root/Root.tsx
```diff
@@ -45,6 +47,15 @@ const Root: FC = () => {
     registerCustomScreens();
   }, []);
 
+  const user = useAccountStore((state) => state.user);
+  const [isSubscribed, setIsSubscribed] = useState(false);
```
darkoatanasovski: this logic is better to be in `initializeAccount` maybe? What do you think?

borkopetrovicc: 22fba788335d7bc86349dd1d04137288c28c724a

borkopetrovicc: https://github.com/jwplayer/ott-web-app/pull/307/commits/22fba788335d7bc86349dd1d04137288c28c724a

borkopetrovicc: https://github.com/jwplayer/ott-web-app/pull/307/commits/22fba788335d7bc86349dd1d04137288c28c724a

borkopetrovicc: Check the latest changes.

src/services/inplayer.account.service.ts
```diff
@@ -107,8 +107,8 @@ export const register: Register = async ({ config, email, password }) => {
 
 export const logout = async () => {
   try {
-    await InPlayer.Account.signOut();
-    return InPlayer.Notifications.unsubscribe();
+    InPlayer.Notifications.unsubscribe();
```
kmitrovv, line 120: We need await here. 
If somehow logout happens before the notifications.unsubscribe, the modal for simultaneous logins will show, which we don't want to in this case.

src/containers/AccountModal/AccountModal.module.scss
```diff
@@ -8,3 +8,9 @@
     max-width: 50%;
   }
 }
+
+.title {
```
kmitrovv: I would add more descriptive className since the module is general for all the forms.

src/services/inplayer.account.service.ts
naumovski-filip: I had the same thought, but

```typescript 
InPlayer.Notifications.unsubscribe()
```

returns void so awaiting it shouldn't make a difference in this case.

borkopetrovicc: Screenshot 2023-06-06 at 15 53 11

src/containers/AccountModal/AccountModal.module.scss
borkopetrovicc: 19748e7bb093fe28b134b3ac04a40d040fcf06d1

src/services/inplayer.account.service.ts
kmitrovv: In that case we can maybe do smth like this:

```
await new Promise((resolve) => {
   InPlayer.Notifications.unsubscribe()
    resolve();
  });
InPlayer.Account.signOut();
```

ChristiaanScheermeijer: There is no need to wrap the unsubscribe call with a promise when it doesn't return a promise.

When the function is blocking, it should have unsubscribed before proceeding to the `signOut` call.

This is essentially the same as your suggestion:

```ts
InPlayer.Notifications.unsubscribe();
InPlayer.Accounts.signOut();
```

In case the unsubscribe method is not blocking, a different method needs to be implemented to acknowledge the subscriptions are disposed before calling `signOut`, but I assume this is not the case.

src/stores/AccountController.ts
```diff
@@ -107,13 +109,23 @@ export async function updateUser(values: FirstLastNameInput | EmailConfirmPasswo
...
+      useAccountStore.setState({ loading: false });
+    } catch (error: unknown) {
+      useAccountStore.setState({
+        user: null,
+        subscription: null,
+        transactions: null,
+        activePayment: null,
+        customerConsents: null,
+        publisherConsents: null,
+      });
```
ChristiaanScheermeijer: Should the loading state also be set to `false` on error?

src/services/inplayer.account.service.ts
kmitrovv: Yeah, you are right. The promise will resolve instantly after calling the unsubscribe and we have pretty much the same effect.
Then I guess it's okay to stay like this, but I am afraid if somehow unsubscribe does not happen and logout is made, the simultaneous login modal will popup..

src/stores/NotificationsController.ts
```diff
@@ -14,30 +15,36 @@ export enum NotificationsTypes {
...
+        if (message) {
+          const notification = JSON.parse(message);
+          switch (notification.type) {
+            case NotificationsTypes.FAILED:
+            case NotificationsTypes.SUBSCRIBE_FAILED:
+              window.location.href = addQueryParams(window.location.href, { u: 'payment-error', message: notification.resource?.message });
+              break;
+            case NotificationsTypes.ACCOUNT_LOGOUT:
+              await getAccount();
+              window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
```
darkoatanasovski, line 29: we need to show SimultaneousLoginModal warning only if the user reached the limit of simultaneous logins, not on every logout.

src/stores/NotificationsController.ts
```diff
@@ -14,30 +15,38 @@ export enum NotificationsTypes {
...
-          case NotificationsTypes.ACCESS_REVOKED:
-            await reloadActiveSubscription();
-            break;
-          case NotificationsTypes.CARD_REQUIRES_ACTION:
-          case NotificationsTypes.SUBSCRIBE_REQUIRES_ACTION:
-            window.location.href = notification.resource?.redirect_to_url;
-            break;
-          default:
-            break;
+  return await useAccount(async () => {
```
darkoatanasovski: can we remove `useAccount` here?

borkopetrovicc: 8b288aeab7252f3fb3dd50e5ddae5fb492d153e3

borkopetrovicc:
Is this pr ready for merging @darkoatanasovski @kiremitrov123 @naumovski-filip @ChristiaanScheermeijer @dbudzins

darkoatanasovski (approved):

src/containers/AccountModal/forms/SimultaneousLoginsNotification.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const navigate = useNavigate();
+  const location = useLocation();
+  function loginClickHandler() {
+    navigate(addQueryParam(location, 'u', 'login'));
+  }
+
+  return (
+    <React.Fragment>
+      <h1 className={styles.simultaneousLoginsTitle}>You have been logged out because the simultaneous logins limit has been reached.</h1>
+      <Button type="submit" onClick={loginClickHandler} label="Ok" variant="contained" color="primary" size="large" fullWidth />
```
ChristiaanScheermeijer: Can you add these texts to the translations?

src/stores/NotificationsController.ts
```diff
@@ -23,6 +23,12 @@ export const subscribeToNotifications = async (uuid: string = '') => {
           case NotificationsTypes.SUBSCRIBE_FAILED:
             window.location.href = addQueryParams(window.location.href, { u: 'payment-error', message: notification.resource?.message });
             break;
+          case NotificationsTypes.ACCOUNT_LOGOUT:
+            if (notification.resource?.reason === 'sessions_limit') {
+              window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
```
ChristiaanScheermeijer, line 29: I realize now... by using `window.location.href = 'new url'` the page actually refreshes. This doesn't result in a history push. Are you sure this is working as intended?

ChristiaanScheermeijer (commented):
Overall the PR looks good to me. I left a few comments.

src/stores/NotificationsController.ts
borkopetrovicc: Yes it works as intended.

src/stores/AccountController.ts
borkopetrovicc: 1cd4df139ddaf21349f5e8db55ad2703422e6531

src/stores/AccountController.ts
```diff
@@ -107,13 +109,23 @@ export async function updateUser(values: FirstLastNameInput | EmailConfirmPasswo
...
-      await afterLogin(response.user, response.customerConsents, accessModel);
+      useAccountStore.setState({ loading: false });
+    } catch (error: unknown) {
+      useAccountStore.setState({
+        user: null,
+        subscription: null,
+        transactions: null,
+        activePayment: null,
+        customerConsents: null,
+        publisherConsents: null,
```
darkoatanasovski, line 127:
```suggestion
        publisherConsents: null,
        loading: false,
```

PR #310, opened 2023-06-07
Feat/Custom register fields

mirovladimitrovski:
## Description

Depends on: https://github.com/jwplayer/ott-web-app/pull/317

Implemented `Custom register fields`:

- In `Register` screen:
image

- In `Account` section:
image

This PR solves [this task](https://www.notion.so/Custom-register-fields-on-the-web-app-cross-selling-phase-232792864d044a40b83bb7b4ea49141a)

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


types/account.d.ts
```diff
@@ -218,15 +220,42 @@ export type UpdateCustomerArgs = {
   fullName?: string;
 };
 
-export type Consent = {
+export interface CleengConsent {
```
kmitrovv: I think it's better to resolve this on service level with formatter methods as the previous functionalities.. We don't need separate types, since there may be new integrations in the future and it will become harder and messy with checks and conditions.
In short, we should avoid using cleeng or jw.

src/components/Account/Account.test.tsx
```diff
@@ -6,12 +6,13 @@ import customer from '#test/fixtures/customer.json';
 import { useAccountStore } from '#src/stores/AccountStore';
 import { renderWithRouter } from '#test/testUtils';
 import type { Consent } from '#types/account';
+import type { ConsentFieldVariants } from '#src/services/inplayer.account.service';
```
kmitrovv: We should probably think of some other way to handle the enums, so we don't have imports from services into the components.
The not recognising exported enums from other files may be a vite problem, but if it's not doable, then maybe switch to using type union or object/array variables.

kmitrovv (commented):
The second step modal should not be showing anymore since the custom fields are handled on the register form.

types/account.d.ts
```diff
@@ -5,6 +5,8 @@ import type { SerializedFavorite } from './favorite';
 
 import type { Config } from '#types/Config';
 
+export type StringedBoolean = 'true' | 'false';
```
kmitrovv: this also can be handled in the formatter method.

types/account.d.ts
mirovladimitrovski: I agree about resolving it on service level, which only means moving this type from this file to the service file where it belongs (in this case `cleeng.account.service.ts`), but I don't think formatter functions are a better solution than keeping a clean type of what a Cleeng consent looks like before being "converted/formatted" into its useful shape, and keep the logic of how it's formatted unexposed and contained in the only place it's ever to be used - the `getPublisherConsents` function. It's a simple straight-forward logic and it wouldn't be used from anywhere else, because that's the reason we're doing the formatting this early on - so that we avoid possible breakage due to differences between `cleeng` consents and `jwp`/`inplayer` register fields. Once that's in place, extracting that logic into a separate function is not only unnecessary, but can trick other contributors into thinking it's meant to be reused from elsewhere, which can potentially lead to messy scenarios.

mirovladimitrovski:
> The second step modal should not be showing anymore since the custom fields are handled on the register form.

What happens to the fields from the second modal then?

src/components/Account/Account.test.tsx
mirovladimitrovski: I hate replacing enums with union types, but it unfortunately seems unavoidable.

mirovladimitrovski:
> The second step modal should not be showing anymore since the custom fields are handled on the register form.

The second step will be shown, but it was decided that `First name` and `Surname` fields will be excluded from step 1 by filtering, and then be displayed in step 2, the way it is now.

src/utils/collection.ts
```diff
@@ -69,6 +69,25 @@ const formatConsentValues = (publisherConsents: Consent[] | null = [], customerC
...
     return {};
   }
+
+  const values: Record<string, string | boolean> = {};
+
+  publisherConsents?.forEach((publisherConsent) => {
+    const consent = customerConsents?.find((customerConsent) => customerConsent.name === publisherConsent.name);
+
+    if (consent) {
+      const value = publisherConsent.provider === 'cleeng' ? consent.state === 'accepted' : consent.value ?? '';
```
darkoatanasovski: can we have a more generic solution here? The idea is to not include the provider's name

types/account.d.ts
```diff
@@ -218,15 +218,22 @@ export type UpdateCustomerArgs = {
   fullName?: string;
 };
 
-export type Consent = {
-  broadcasterId: number;
+export type ConsentProvider = 'cleeng' | 'jwp';
```
darkoatanasovski: do we need a type for the providers?

src/utils/collection.ts
mirovladimitrovski: I'm open to suggestions. The point is that `cleeng` consents have just checkbox boolean values and a lot of the logic and data being recorded to BE is based around that - hence they're either recorded as `accepted` or `declined`. Unlike `jwp` consents*, whose values can be boolean, string or number, and they don't fit into the constraint of being either accepted or declined. The only way I could think of for making that distinction was the provider name.

*`jwp` consents practically exceed what a _consent_ is and can be troublesome/misleading to keep using that term for what are actually `custom register fields`, but that's the terminology used throughout the OTT app codebase. Maybe we need to change that in the future.

types/account.d.ts
mirovladimitrovski: If we find a way to avoid using those values then this type won't be needed.

src/utils/collection.ts
mirovladimitrovski: Done, I replaced `provider: 'cleeng' | 'jwp'` with the more generic `isCustomRegisterField: boolean`.

src/static/json/countries.json
budzworthy, line 1: Can we put these into public/locales/{locale} instead and load them by locale so that they can be translated? We might have to think about how to best setup the i18n so the parser plays nicely and we get the same set of states and countries for each language. (Note: there are some geopolitical things to think about with country lists and different languages and locales (i.e. some countries don't recognize other countries and legally you can't show them for products in those countries), but I don't think we want to get into those details yet.)

src/static/json/index.ts
budzworthy, line 1: We have typically avoided extra boilerplate like index files in the web app

src/components/RegistrationForm/RegistrationForm.tsx
budzworthy, line 1: By putting all of the fields on 1 page, I can't see the form feedback when something is wrong after I click the Continue button, especially on small screens / mobile. Maybe we can just auto-scroll back to the top on errors?

https://github.com/jwplayer/ott-web-app/assets/7417979/8b8c0020-a019-4f03-b191-5b44d8b7ec87

src/static/json/countries.json
mirovladimitrovski: Agreed. So what do we do with `public/locales/es`? Put the exact same copy for now and then deal with the translation later?

src/static/json/countries.json
```diff
@@ -0,0 +1,251 @@
+{
+  "af": "Afghanistan",
```
AntonLantukh: fyi, there is a list of countries stored in `frontend-common` repo. The same is true for states. We can't really reuse it here as they are part of the closed repo. However it is worth keeping in mind that we should better keep them in sync.

package.json
```diff
@@ -128,7 +128,7 @@
     "vite-plugin-pwa": "^0.14.0",
     "vite-plugin-static-copy": "^0.13.0",
     "vite-plugin-stylelint": "^4.0.1",
-    "vitest": "^0.32.3",
+    "vitest": "^0.33.0",
```
AntonLantukh: Do we need to update `vitest` in this PR?

public/locales/es/user.json
```diff
@@ -39,6 +39,7 @@
     "firstname": "Nombre",
     "hide_password": "Ocultar contraseña",
     "lastname": "Apellido",
+    "other_registration_details": "",
```
AntonLantukh: You can add some machine translations for this key and then replace it (or not) with a verified value.

vite.config.ts
```diff
@@ -133,5 +134,5 @@ export default ({ mode, command }: ConfigEnv): UserConfigExport => {
       setupFiles: ['test/vitest.setup.ts'],
       css: true,
     },
-  });
+  } as UserConfigExport);
```
AntonLantukh: Did vitest update lead to this explicit type?

src/components/Radio/Radio.tsx
```diff
@@ -29,10 +29,10 @@ const Radio: React.FC<Props> = ({ name, onChange, header, value, values, helperT
           {!required ? <span>{t('optional')}</span> : null}
         </div>
       ) : null}
-      {values.map((optionValue, index) => (
+      {values.map(({ value: optionValue, label: optionLabel }, index) => (
         <div className={styles.radio} key={index}>
           <input value={optionValue} name={name} type="radio" id={id + index} onChange={onChange} checked={value === optionValue} required={required} />
-          <label htmlFor={id + index}>{optionValue}</label>
+          <label htmlFor={id + index}>{optionLabel ?? optionValue}</label>
```
AntonLantukh: Do we need to show the value here in case label is not defined? Is it a valid case?

src/services/inplayer.account.service.ts
```diff
@@ -158,11 +170,26 @@ export const getPublisherConsents: GetPublisherConsents = async (config) => {
     const { jwp } = config.integrations;
     const { data } = await InPlayer.Account.getRegisterFields(jwp?.clientId || '');
 
-    const result: Consent[] = data?.collection.filter((field) => field.type === 'checkbox').map((consent) => formatPublisherConsents(consent));
-
-    return {
-      consents: [getTermsConsent(), ...result],
-    };
+    const result = data?.collection
+      .filter((field) => !['email_confirmation', 'first_name', 'surname'].includes(field.name))
```
AntonLantukh, line 164: Do we filter these options because we already have them by default? Should better add a comment here.

AntonLantukh:
Hey! I have some comments after testing:

1. I didn't fill 'Gender' field which was not marked as 'optional' but I could still register.
2. I don't really like the way we show 'Optional' fro components without labels. It is not clear for me which field it relates to. Something to think about in the future.

AntonLantukh:
One more thing to note. I used my config with JW integration and signed up successfully. Then I logged out, switched to my Cleeng config but could still see Custom Fields from InPlayer. I agree that customers don't usually switch between different configs but probably we should still invalidate the data when another config is used.

vite.config.ts
mirovladimitrovski: Yes, exactly.

src/components/Radio/Radio.tsx
mirovladimitrovski: Good call, there was such case until I did this later:
image

So I guess we don't need a double fallback.

src/static/json/countries.json
budzworthy: Machine translations for state and country names are probably more reliable than most other translations, or I bet you can even find a pre-defined list you can copy.

src/components/Form/FormSection.tsx
```diff
@@ -52,13 +53,8 @@ export function FormSection<TData extends GenericFormValues>({
...
   const isEditing = sectionId === activeSectionId;
 
-  const onChange = useCallback(
-    function onChange({ currentTarget }: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>) {
-      if (!currentTarget) return;
-
-      const { name, type } = currentTarget;
-      const value = type === 'checkbox' ? (currentTarget as HTMLInputElement).checked : currentTarget.value;
-
+  const onChangeValue = useCallback(
```
darkoatanasovski: is there any specific reason for this change?

src/components/Form/FormSection.tsx
```diff
@@ -155,10 +163,10 @@ export function FormSection<TData extends GenericFormValues>({
       {content &&
         (isEditing ? (
           <form className={styles.flexBox} noValidate onSubmit={(event) => event.preventDefault()}>
-            {content({ values, isEditing, isBusy, onChange, errors: formErrors })}
+            {content({ values, isEditing, isBusy, onChangeEvent, onChangeValue, errors: formErrors })}
```
darkoatanasovski: you can handle CustomRegisterFields changes with onChange, I am not sure that we need separate methods

src/components/CustomRegisterField/CustomRegisterField.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (!optionsObject) {
+      return [];
+    }
+
+    return Object.entries(optionsObject).map(([value, label]) => ({ value, label }));
+  }, [type, props.options]);
+
+  const changeHandler: ChangeEventHandler<HTMLInputElement> = useCallback(
+    ({ currentTarget }) => {
+      const value = type === REGISTER_FIELD_VARIANT.CHECKBOX ? (currentTarget as HTMLInputElement).checked : currentTarget.value;
```
darkoatanasovski: this is also handled in the parent `onChange` method. Can we just pass the values to the parent's `onChange`?

darkoatanasovski (changes requested):

src/components/Form/FormSection.tsx
mirovladimitrovski: The `<DateField />` component inside `<CustomRegisterField />` is the reason for this. The `onChange` callback relies on an actual event object - the React-specific `SyntheticEvent` one, but `<DateField onChange />` doesn't produce such event because internally it doesn't represent a single element which can be an event target, which makes it incompatible with onChange (that's why I renamed it to `onChangeEvent`). I had thought about triggering something like `new Event('change')`, but at the very least the native JS `Event` object is still incompatible with `SyntheticEvent`, and React doesn't offer `new SyntheticEvent('change')`, it's just a typescript type and not an constructor.

src/components/CustomRegisterField/CustomRegisterField.tsx
mirovladimitrovski: As I already answered, the difference is that the parent's `onChange` method requires an event object as its only argument, while the method with the same name here requires two arguments `name` and `value`, and the only reason for that is the `DateField` component which doesn't produce an event object.

src/static/json/countries.json
mirovladimitrovski: Agreed, I've copied the exact same lists for countries and states.

package.json
mirovladimitrovski: I'm resolving this conversation since the comment has become obsolete because `vitest` is already updated on `develop` and the latest changes have been merged in this branch and PR.

vite.config.ts
mirovladimitrovski: With the last update of the `vitest` dependency this change has become unnecessary, so I reverted it.

src/components/CustomRegisterField/CustomRegisterField.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { useTranslation } from 'react-i18next';
+import type { GetRegisterFieldOption } from '@inplayer-org/inplayer.js';
+
+import Checkbox from '#components/Checkbox/Checkbox';
+import TextField from '#components/TextField/TextField';
+import Radio from '#components/Radio/Radio';
+import Dropdown from '#components/Dropdown/Dropdown';
+import DateField from '#components/DateField/DateField';
+import { ConsentFieldVariants, REGISTER_FIELD_VARIANT } from '#src/services/inplayer.account.service';
+import countriesCodes from '#static/countries-codes.json';
```
budzworthy: It looks like this is the only place this import is used... it's probably not worth adding an alias to save 3 characters on 2 lines in a single file.
```suggestion
import countriesCodes from '#src/static/countries-codes.json';
```

.depcheckrc.yaml
```diff
@@ -15,6 +15,7 @@ ignores: [
     '\#types',
     '\#components',
     '\#utils',
+    '\#static',
```
budzworthy:
```suggestion
```

vite.config.ts
```diff
@@ -123,6 +123,7 @@ export default ({ mode, command }: ConfigEnv): UserConfigExport => {
         '#test': path.join(__dirname, 'test'),
         '#test-e2e': path.join(__dirname, 'test-e2e'),
         '#types': path.join(__dirname, 'types'),
+        '#static': path.join(__dirname, 'src/static'),
```
budzworthy:
```suggestion
```

tsconfig.json
```diff
@@ -27,6 +27,9 @@
       ],
       "#types/*": [
         "./types/*"
+      ],
+      "#static/*": [
+        "./src/static/*"
```
budzworthy:
```suggestion
```

src/static/countries-codes.json
budzworthy, line 1: Naming such as `country-codes` is more common

src/static/us-states-codes.json
budzworthy, line 1: Naming `us-state-codes`

src/components/CustomRegisterField/CustomRegisterField.tsx
```diff
@@ -0,0 +1,71 @@
...
+export type CustomRegisterFieldCommonProps = Props;
+
+export const CustomRegisterField: FC<Props> = ({ type, value = '', ...props }) => {
+  const { t } = useTranslation(type);
+
+  const optionsList = useMemo(() => {
+    switch (type) {
+      case REGISTER_FIELD_VARIANT.COUNTRY_SELECT:
+      case REGISTER_FIELD_VARIANT.US_STATE_SELECT: {
+        return (type === REGISTER_FIELD_VARIANT.COUNTRY_SELECT ? countriesCodes : usStatesCodes).map((code) => ({
```
budzworthy: This is kind of a predicament. As much as possible we want to define the country and state codes in 1 place and have the translations automatically pull from that place. Can you update the i18n scripts to read the json files and make sure the expected tags are automatically added to all languages?

src/static/countries-codes.json
budzworthy, line 1: I think these can just go into `src/assets` instead of creating another folder with just a couple things in it.

src/components/CustomRegisterField/CustomRegisterField.tsx
mirovladimitrovski: This is a specific case in which I have to iterate through a list of known tags before the actual translation occurs in runtime. There are two things I can do that can think of right now:

1. Statically import `country.json` and `us_state.json` from `#src/../public/locales/en`. Which immediately makes those files part of the bundle, even though `react-i18next` would later load practically the same data dynamically, resulting into duplication. I don't think it's something either of us would like.
2. Or I can get the tag list in runtime from `const { i18n } = useTranslation()` and then `Object.keys(i18n.getResourceBundle('en', 'country'))`, but that way I stumble to another problem - the `i18n` instance is NOT initialised in test environments, so there's no `.getResourceBundle()` function available, so unit tests for the `CustomRegisterField` fail.

darkoatanasovski (approved):

src/components/Account/Account.tsx
```diff
@@ -8,6 +8,8 @@ import { useMutation } from 'react-query';
 import styles from './Account.module.scss';
 
 import type { FormSectionContentArgs, FormSectionProps } from '#components/Form/FormSection';
+import type { Consent } from '#types/account';
+import { REGISTER_FIELD_VARIANT } from '#src/services/inplayer.account.service';
```
kmitrovv: We should avoid having imports from services/stores inside components. What do you think of creating a new file under utils/registerFields.ts and moving `REGISTER_FIELD_VARIANT` there?
Same goes for `ConsentFieldVariants`.

src/services/cleeng.account.service.ts
```diff
@@ -131,15 +133,37 @@ export async function getUser({ config }: { config: Config }) {
...
 export const getPublisherConsents: GetPublisherConsents = async (config) => {
   const { cleeng } = config.integrations;
   const response = await get(!!cleeng?.useSandbox, `/publishers/${cleeng?.id}/consents`);
 
   handleErrors(response.errors);
 
-  return {
-    consents: response?.responseData?.consents || [],
-  };
+  const consents = ((response?.responseData?.consents || []) as CleengConsent[]).map(
```
kmitrovv: Why do we need to map cleeng's response?

mirovladimitrovski: For consistency between responses from different providers and wherever they're handled. The same kind of mapping is done in `inplayer.account.service` as well. For example, you'll notice that cleeng's initial value of a consent is constrained to a boolean value and is expressed in the property called `enabledByDefault`, while on the other hand an inplayer register field has `default_value`, hence they're both mapped to `defaultValue`. There's also a boolean property called `isCustomRegisterField` that also has its usage.

kmitrovv: Yes but on InPlayer side we map all the responses in order to match the cleeng's ones, not the other way around.. This way we need to make sure nothing is broken on cleeng's side..

PR #311, opened 2023-06-07
chore: add upload content type script

budzworthy:
## Description

Content type upload script for standard web app content types.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code


docs/content-types.md
```diff
@@ -0,0 +1,39 @@
...
+Please refer to [content-types.json](scripts/content-types/content-types.json) and the JWP documentation for the schema format.
+
+To avoid unnecessary duplication the file also allows some basic abstraction.
+You can define reused fields and sections as key-value entries on the `fields` and `sections` properties respectively.
+Then you can include these reusable entities by putting their string key into schemas the same way that you would for inline fields or sections.
+There are many examples in the included [content-types.json](scripts/content-types/content-types.json)
+
+> Note: Although the upload file allows you to define reused fields and sections, when these are uploaded they become distinct copies for each instance in the schemas where they are used.
+> That means that changing fields and sections via the api after they are uploaded must be done individually for each schema.
+> Alternatively, you can
```
AntonLantukh: Looks like some text is missing here.

src/pages/ScreenRouting/mediaScreens/MediaHub/MediaHub.tsx
```diff
@@ -13,7 +13,7 @@ const parsePlaylistIds = (input: unknown): Content[] => {
 };
 
 const MediaHub: ScreenComponent<PlaylistItem> = ({ data }) => {
-  const rows = useMemo(() => parsePlaylistIds(data.playlists), [data.playlists]);
+  const rows = useMemo(() => parsePlaylistIds(data.playlist || data.playlists), [data.playlist, data.playlists]);
```
AntonLantukh: Hm, why do we have both `data.playlist` and `data.playlists` here? Do we expect it to be a custom param?

scripts/content-types/content-types.json
```diff
@@ -0,0 +1,358 @@
...
+          }
+        ]
+      }
+    },
+    "trailer_id": {
+      "param": "trailerId",
+      "label": "Trailer",
+      "details": {
+        "field_type": "media_select",
+        "placeholder": "Select a trailer"
```
AntonLantukh: Do we need any validation rules for default schemas?

scripts/content-types/load-content-types.ts
```diff
@@ -0,0 +1,252 @@
...
+  schemas: Schema<Field | string, Section<Field | string> | string>[];
+}
+
+// TODO: Use yup to validate schema
+interface Field {
+  label: string;
+  param: string;
+  required?: boolean;
+  details: {
+    field_type: 'input' | 'toggle';
```
AntonLantukh: Should we also include other types here?

scripts/content-types/load-content-types.ts
```diff
@@ -0,0 +1,252 @@
...
+    [key: string]: Section<Field | string>;
+  };
+  schemas: Schema<Field | string, Section<Field | string> | string>[];
+}
+
+// TODO: Use yup to validate schema
+interface Field {
+  label: string;
+  param: string;
+  required?: boolean;
```
AntonLantukh: There can be also `description` property here (tooltip)

scripts/content-types/load-content-types.ts
```diff
@@ -0,0 +1,252 @@
...
+
+function parseSchemas(file: string) {
+  try {
+    const schemaDefinitions = JSON.parse(fs.readFileSync(file, 'utf-8')) as SchemaFile;
+
+    if (schemaDefinitions.schemas.length <= 0) {
+      error(`Error! No schemas found in file ${file}`, -4);
+    }
+
+    const mapSchema = (schema: Schema<Field | string, Section<Field | string> | string>) => {
```
AntonLantukh: Can be probably used as a separate type (the same is used in SchemaFile['schemas'])

scripts/content-types/load-content-types.ts
```diff
@@ -0,0 +1,252 @@
+/* eslint @typescript-eslint/no-var-requires: "off" */
+const fs = require('fs');
+const readline = require('readline');
+
+const JWP_HOST = process.env.JWP_HOST || 'https://api.jwplayer.com';
```
AntonLantukh, line 6: Maybe we can also pass it as an argument? Could be useful to create schemas for dev

AntonLantukh:
Could not test how schemas are displayed in UI as we still wait for the release to support renaming.

src/pages/ScreenRouting/mediaScreens/MediaHub/MediaHub.tsx
budzworthy: Yes, because the current hub type only has a single select option until we add multi-select playlist type. I'll add a comment.

scripts/content-types/content-types.json
budzworthy: Not sure what you mean by 'validation rules'. I wanted to err on the side of less schema validation in the script, just to avoid duplicating checks that the API already will do. Data should just pass through as much as possible so we don't risk accidentally mutating it in an unexpected way or having extra script-side validation that gets out of sync with the API validation.

scripts/content-types/load-content-types.ts
budzworthy: I removed all of the properties besides the ones we are using.

scripts/content-types/load-content-types.ts
budzworthy: I removed all of the properties besides the ones we are using.

scripts/content-types/load-content-types.ts
budzworthy: Nice suggestion. Done.

scripts/content-types/load-content-types.ts
budzworthy: I intentionally made this a less obvious param, because it's something only JWP devs will need to care about. You can always set it using:
`JWP_HOST=https://api-dev.jwplayer.com/ yarn load-content-types...`

docs/content-types.md
budzworthy: Updated

scripts/content-types/content-types.json
AntonLantukh: I only mean some validation for the fields we have, like `required`, `minLength` / `maxLength`

scripts/content-types/load-content-types.ts
```diff
@@ -14,31 +16,19 @@ interface SchemaFile {
...
-interface Field {
-  label: string;
-  param: string;
-  required?: boolean;
-  details: {
-    field_type: 'input' | 'toggle';
-    placeholder: string;
-  };
-}
+// Only defining the properties we need here for the script.
```
AntonLantukh: Should we modify this comment in case `Field` type is `unknown` now?

scripts/content-types/content-types.json
budzworthy: Oh I see. Were there any fields you saw that needed constraints?

AntonLantukh: I would say that all `VCH.XXX` props and `scheduleUrl` are required ones. I would also mark `playlist` for `hub` as a required prop because it just won't work without a playlist. Maybe also `markdown` ?

AntonLantukh (approved):

scripts/content-types/content-types.json
budzworthy: Markdown and playlist are good ones. I'll mark those. I think we should leave the VCH properties alone, because it would be a weird scenario where those were missing and required, since they are auto-populated.

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "jw-ott-webapp",
   "version": "4.19.0",
   "main": "index.js",
+  "type": "commonjs",
```
AntonLantukh: Do we need `commonjs` here if we use modules in our script?

scripts/content-types/lib/utils.ts
```diff
@@ -0,0 +1,71 @@
+// @ts-ignore
```
AntonLantukh, line 1: Do we still need ignore here?

scripts/content-types/load-content-types.ts
```diff
@@ -0,0 +1,56 @@
...
+    console.info('Optional argument: source-file - specify a schema file to upload (defaults to ./content-types.json');
+    console.info(
+      'Optional argument: filter - specify a CSV list of schema names in order to limit the schemas to be uploaded to a subset of those found in the source file',
+    );
+
+    process.exit(0);
+  }
+
+  const file = args['source-file'] || `${__dirname}/content-types.json`;
+  const apiKey = await apiKeyPrompt();
```
AntonLantukh, line 32: Probably it makes sense to make it required param.

AntonLantukh:
Got an error when trying to create schemas:

```
 Error creating trailer: 400 - {"errors":[{"code":"invalid_body","description":"Failed validating field sections: {'field_type': 'toggle', 'default': True, 'placeholder': 'Is this item free to watch?'} is not valid under any of the given schemas"}]}
```

Looks like `placeholder` field is not used for `toggle`.

Same for `media_select` and `playlist_select`

package.json
budzworthy: Good catch. Yes, it works without it! Thanks for pointing it out.

scripts/content-types/lib/utils.ts
budzworthy: Yes because the read package doesn't have types available

scripts/content-types/load-content-types.ts
budzworthy: I intentionally removed it because it was a security leak to keep pasting it into the terminal. It's annoying, but more secure to prompt for it each time.

PR #312, opened 2023-06-08
Fix / checkout issues

ChristiaanScheermeijer:
## Description

This PR fixes two problems. The first problem was an infinite render loop when opening the choose offers modal. This was caused by using a default value as react-query data. This becomes an issue when the same variable becomes a dependency in a useEffect/useMemo.

image

The second change fixes a React hook violation. This probably wasn't an issue before because it didn't use any hooks internally, but we've added a `useMemo` hook to ... ironically ... also prevent infinite render loops.

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -95,7 +95,7 @@ module.exports = {
       },
     },
     {
-      files: ['*.jsx', '*.tsx'],
+      files: ['*.jsx', '*.tsx', 'src/hooks/*.ts'],
       plugins: [
         // Enable linting React code
         'react',
--- a/src/hooks/useCheckAccess.ts
+++ b/src/hooks/useCheckAccess.ts
@@ -12,6 +12,7 @@ type intervalCheckAccessPayload = {
   iterations?: number;
   offerId?: string;
 };
+
 const useCheckAccess = () => {
   const intervalRef = useRef<number>();
   const navigate = useNavigate();
@@ -25,6 +26,7 @@ const useCheckAccess = () => {
       if (!offerId && clientOffers?.[0]) {
         offerId = clientOffers[0];
       }
+
       intervalRef.current = window.setInterval(async () => {
         const hasAccess = await checkEntitlements(offerId);
 
@@ -37,7 +39,7 @@ const useCheckAccess = () => {
         }
       }, interval);
     },
-    [intervalRef.current, errorMessage],
+    [clientOffers, navigate, location, t],
   );
 
   useEffect(() => {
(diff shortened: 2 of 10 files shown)


src/hooks/useCheckAccess.ts
```diff
@@ -18,10 +18,10 @@ const useCheckAccess = () => {
   const location = useLocation();
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
   const { t } = useTranslation('user');
+  const { clientOffers } = useClientIntegration();
```
AntonLantukh, line 22: Should we also add it as a dependency to the callback?

src/hooks/useOffers.ts
```diff
@@ -28,16 +28,15 @@ const useOffers = () => {
     return [...(requestedMediaOffers || []).map(({ offerId }) => offerId), ...clientOffers].filter(Boolean);
   }, [requestedMediaOffers, clientOffers]);
 
-  const { data: allOffers = [], isLoading } = useQuery(['offers', offerIds.join('-')], () => checkoutService.getOffers({ offerIds }, sandbox));
+  const { data: allOffers, isLoading } = useQuery(['offers', offerIds.join('-')], () => checkoutService.getOffers({ offerIds }, sandbox));
 
   // The `offerQueries` variable mutates on each render which prevents the useMemo to work properly.
   return useMemo(() => {
```
AntonLantukh, line 34: Do you think we need `useMemo` here?

src/hooks/useCheckAccess.ts
ChristiaanScheermeijer: Yes, it should! I realized that ESLint didn't include hooks due to the file extension. I changed that, but that brought some other changes as well 😅

src/hooks/useOffers.ts
ChristiaanScheermeijer: For now, we need it because we're creating a new `offers` array using some state. A few other callbacks and effects depend on the `offers` variable.

.eslintrc.js
```diff
@@ -95,7 +95,7 @@ module.exports = {
       },
     },
     {
-      files: ['*.jsx', '*.tsx'],
+      files: ['*.jsx', '*.tsx', 'src/hooks/*.ts'],
```
AntonLantukh, line 98: Maybe we should add all `.ts` here. But it can be done later, could probably result in more changes.

AntonLantukh (approved):

kmitrovv (approved):
lgtm

ChristiaanScheermeijer:
@AntonLantukh @kiremitrov123 if we merge this PR, we don't need to change the target branch in #313

PR #313, opened 2023-06-13
Fix / choose offer modal without prices

kmitrovv:
## Description

Choose offer modal improvements so that when someone tries to pay a tvod item and therefore the item doesn't have prices set up, we show them message that there are no prices instead of not doing anything despite clicking the buy button.
image
Here is an example how it looks now and you can navigate to the subscription prices and back, but you will not be able to continue with the flow for the TVOD item since it doesn't have prices the button for continue is disabled.

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/public/locales/es/account.json
+++ b/public/locales/es/account.json
@@ -67,6 +67,7 @@
     "field_required": "Este campo es obligatorio",
     "monthly": "Mensual",
     "monthly_subscription": "Suscripción mensual",
+    "no_pricing_available": "No hay opciones de precios disponibles en este momento.",
     "offer_not_found": "Oferta no encontrada, por favor intenta volver a cargar la página",
     "one_time_only": "Este video",
     "subscription": "Suscripción",
--- a/src/components/ChooseOfferForm/ChooseOfferForm.tsx
+++ b/src/components/ChooseOfferForm/ChooseOfferForm.tsx
@@ -172,9 +172,9 @@ const ChooseOfferForm: React.FC<Props> = ({
           </label>
         </div>
       )}
-      <div className={styles.offers}>{offers.map(renderOfferBox)}</div>
+      {<div className={styles.offers}>{!offers.length ? <p>{t('choose_offer.no_pricing_available')}</p> : offers.map(renderOfferBox)}</div>}
       {submitting && <LoadingOverlay transparentBackground inline />}
-      <Button label={t('choose_offer.continue')} disabled={submitting} variant="contained" color="primary" type="submit" fullWidth />
+      <Button label={t('choose_offer.continue')} disabled={submitting || !offers.length} variant="contained" color="primary" type="submit" fullWidth />
     </form>
   );
 };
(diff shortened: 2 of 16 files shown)


darkoatanasovski (approved):

public/locales/es/account.json
```diff
@@ -67,6 +67,7 @@
     "field_required": "Este campo es obligatorio",
     "monthly": "Mensual",
     "monthly_subscription": "Suscripción mensual",
+    "no_pricing_available": "No hay opciones de precios disponibles en este momento.",
```
AntonLantukh, line 70: We have decided to create a separate PR for Spanish keys (ott-webapp channel). Did you agree on this one with Jose?

src/components/ChooseOfferForm/ChooseOfferForm.tsx
```diff
@@ -172,9 +172,9 @@ const ChooseOfferForm: React.FC<Props> = ({
           </label>
         </div>
       )}
-      <div className={styles.offers}>{offers.map(renderOfferBox)}</div>
+      {<div className={styles.offers}>{!offers.length ? <p>{t('choose_offer.no_pricing_available')}</p> : offers.map(renderOfferBox)}</div>}
```
AntonLantukh, line 175: Does it make sense to show "This Video" tab in case TVOD is not available?

src/hooks/useOttAnalytics.ts
```diff
@@ -9,44 +9,45 @@ const useOttAnalytics = (item?: PlaylistItem, feedId: string = '') => {
...
 
-  const timeHandler = useCallback(({ position, duration }: jwplayer.TimeParam) => {
-    window.jwpltx.time(position, duration);
-  }, []);
+  useEffect(() => {
+    if (!window.jwpltx || !analyticsToken || !player || !item) {
+      return;
+    }
 
-  const seekHandler = useCallback(({ offset, duration }) => {
```
AntonLantukh, line 20: Would be great if we could use a different target branch to avoid changes from another PR.

src/components/ChooseOfferForm/ChooseOfferForm.tsx
kmitrovv: This video won't be shown if TVOD is not available..but if there are no prices I think it makes sense to show that there are no prices even on subscription asset.

src/hooks/useOttAnalytics.ts
kmitrovv: These changes are from another [PR ](https://github.com/jwplayer/ott-web-app/pull/312)
Since the changes are pretty much in the same place that's why I've created the branch from there.

public/locales/es/account.json
kmitrovv: Ah, didn't know what's been decided there. So I just revert these changes and open different PR only for this string, then check with Jose?

AntonLantukh: Yeah, let's try it this way. Thanks!

PR #317, opened 2023-06-15
feat: Spanish translation for `other_registration_details`

mirovladimitrovski:
## Description

There's a new translation text with tag `other_registration_details` and this PR contains the Spanish translation.

### DIFF
--- a/public/locales/en/user.json
+++ b/public/locales/en/user.json
@@ -41,6 +41,7 @@
     "lastname": "Last name",
     "password": "Password",
     "save": "Save",
+    "other_registration_details": "Other registration details",
     "security": "Password",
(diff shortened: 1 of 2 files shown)


public/locales/en/user.json
```diff
@@ -41,6 +41,7 @@
     "lastname": "Last name",
     "password": "Password",
     "save": "Save",
+    "other_registration_details": "Other registration details",
```
AntonLantukh: We can include English translations in the original PR. The additional one is only for Spanish. Sorry if I made it not clear enough in the message.

mirovladimitrovski: No, it makes perfect sense, I didn't think it through.

AntonLantukh (approved):

PR #318, opened 2023-06-16
feat: add machine translations for delete account and social login

naumovski-filip:
## Description

Adds initial translations with Google Translate for ES in delete account and social login components.

### DIFF
--- a/public/locales/en/user.json
+++ b/public/locales/en/user.json
@@ -55,7 +55,7 @@
     "title": "Favorites"
   },
   "login": {
-    "field_required": ""
+    "field_required": "This field is required"
   },
   "nav": {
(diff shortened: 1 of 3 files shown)


public/locales/en/user.json
```diff
@@ -55,7 +55,7 @@
     "title": "Favorites"
   },
   "login": {
-    "field_required": ""
```
AntonLantukh, line 58: You can add English translation in your original PR.

AntonLantukh (approved):

public/locales/en/user.json
naumovski-filip: Original PR is already merged so doing it here.

PR #320, opened 2023-06-19
feat: initial inplayer subscription change implementation

naumovski-filip:
## Description

Testing config: `7jh8ypyh`


darkoatanasovski (dismissed):
According to the [Figma designs](https://www.figma.com/file/qLl8KxBOOr640h18ICx0H6/Web-app?type=design&node-id=1815-54072&mode=design&t=rYuz2WWxoZCXy1xA-0), cancel subscription button should be visible only when the change plan view is active.

src/components/OfferSwitch/OfferSwitch.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useTranslation } from 'react-i18next';
+
+import styles from './OfferSwitch.module.scss';
+
+import type { Offer } from '#types/checkout';
+import Checkbox from '#components/Checkbox/Checkbox';
+import { formatLocalizedDate, formatPrice } from '#src/utils/formatting';
+import { useAccountStore } from '#src/stores/AccountStore';
+
+interface OfferSwitchProps {
```
ChristiaanScheermeijer: All other components are using `type`. Is there a good reason we should be using `interface` for prop definitions?

src/components/OfferSwitch/OfferSwitch.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isCurrentOffer: boolean;
+  pendingDowngradeOfferId: string;
+  offer: Offer;
+  selected: {
+    value: boolean;
+    set: React.Dispatch<React.SetStateAction<string | null>>;
+  };
```
ChristiaanScheermeijer: This is a bit unconventional because you now require the `set` handler to use `React.Dispatch`.

I suggest making it abstract:

```suggestion
  selected: boolean;
  onChange: () => void;
```

src/components/OfferSwitch/OfferSwitch.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const OfferSwitch = ({ isCurrentOffer, pendingDowngradeOfferId, offer, selected }: OfferSwitchProps) => {
+  const { t, i18n } = useTranslation('user');
+  const { customerPriceInclTax, customerCurrency, period } = offer;
+  const expiresAt = useAccountStore((state) => state.subscription?.expiresAt);
```
ChristiaanScheermeijer: Prevent using stores inside components. You can pass the `expiresAt` via props from the parent container. This keeps all store logic central and components "dumb".

src/components/Payment/Payment.tsx
```diff
@@ -69,6 +76,63 @@ const Payment = ({
   const breakpoint = useBreakpoint();
   const isMobile = breakpoint === Breakpoint.xs;
 
+  const { offers } = useOffers();
```
ChristiaanScheermeijer: The same applies here. The `useOffers` hook is also store dependent and should not be used in components directly.

src/components/Payment/Payment.tsx
```diff
@@ -69,6 +76,63 @@ const Payment = ({
...
+    }
+  }, [selectedOfferId, offers, activeSubscription]);
+
+  const updateSubscriptionMetadata = useMutation(updateUser, {
+    onSuccess: () => {
+      useAccountStore.setState({
+        loading: false,
+      });
+    },
+  });
```
ChristiaanScheermeijer: In my opinion, mutations should also be handled outside components.

src/components/Payment/Payment.tsx
```diff
@@ -149,6 +253,38 @@ const Payment = ({
...
+                .filter((o) => o.planSwitchEnabled)
+                .map((offer) => (
+                  <OfferSwitch
+                    key={offer.offerId}
+                    isCurrentOffer={offer.offerId === activeSubscription?.accessFeeId}
+                    pendingDowngradeOfferId={pendingDowngradeOfferId}
+                    offer={offer}
+                    selected={{ value: selectedOfferId === offer.offerId, set: setSelectedOfferId }}
+                  />
+                ))}
```
ChristiaanScheermeijer: Perhaps the `useOffers` can return a `pendingOffers` collection that is passed to this component.

src/hooks/useOffers.ts
```diff
@@ -52,6 +51,20 @@ const useOffers = () => {
...
+        hasTVODOffers: false,
+        hasMultipleOfferTypes: false,
+        isLoading: false,
+        hasPremierOffer: false,
+        defaultOfferId: '',
+        offerType: 'svod' as OfferType,
+        setOfferType: () => null,
+        offers: [],
+        offersDict: {},
+      };
```
ChristiaanScheermeijer: This change feels a bit hacky and is needed because the `useOffers` hook is used for AVOD and AUTHVOD platforms as well. Can't we better prevent using this hook instead of returning default data?

src/components/OfferSwitch/OfferSwitch.tsx
naumovski-filip: oversight, will fix

src/components/Payment/Payment.tsx
naumovski-filip: I'm not sure how else to approach this after giving it some thought. Any ideas?

ChristiaanScheermeijer: You could create a `<PaymentContainer />`. That would also simplify the `User` screen a bit.

The Payment component then only contains UI while all business logic is placed inside the container.

src/components/Payment/Payment.tsx
ChristiaanScheermeijer: This logic can also be moved to the `PaymentContainer` instead of the `useOffers` hook. Although, we might need it elsewhere...

src/components/Payment/Payment.tsx
naumovski-filip: Please take a look at the most recent commits and see if that's what you had in mind

src/containers/PaymentContainer/PaymentContainer.tsx
```diff
@@ -0,0 +1,136 @@
...
+      }
+    } catch (error: unknown) {
+      throw new Error("Couldn't parse receipt. " + (error instanceof Error ? error.message : ''));
+    }
+
+    setIsLoadingReceipt(false);
+  };
+
+  useEffect(() => {
+    if (accessModel !== 'AVOD') {
```
kiremitrov123, line 75:
```suggestion
    if (accessModel === 'SVOD') {
```
The accessModel can be register only as well (AUTHVOD), in which case we don't need anything connected to subscriptions.

test-e2e/utils/payments.ts
```diff
@@ -63,18 +63,30 @@ export async function finishAndCheckSubscription(I: CodeceptJS.I, billingDate: D
   I.see(yearlyPrice);
   I.see('/year');
   I.see('Next billing date is on ' + formatDate(billingDate));
-  I.see('Cancel subscription');
 
-  I.waitForElement('[class*="transactionItem"]');
+  if (providerName?.includes('JW')) {
```
kmitrovv: what do you think of including additional param `canChangeSubscription` or something, instead of using provider name?

src/components/OfferSwitch/OfferSwitch.tsx
```diff
@@ -0,0 +1,55 @@
...
+          </div>
+        )}
+        <div className={styles.offerSwitchPlanContainer}>
+          <div>{t(`user:payment.${period === 'month' ? 'monthly' : 'annual'}_subscription`)}</div>
+          {(isCurrentOffer || isPendingDowngrade) && expiresAt && (
+            <div className={styles.nextBillingDate}>
+              {isCurrentOffer &&
+                !pendingDowngradeOfferId &&
+                t('user:payment.next_billing_date_on', { date: formatLocalizedDate(new Date(expiresAt * 1000), i18n.language) })}
+              {isPendingDowngrade && t('user:payment.downgrade_on', { date: formatLocalizedDate(new Date(expiresAt * 1000), i18n.language) })}
```
kiremitrov123, line 42: `expiresAt` is possibly undefined and no check is provided

src/components/Payment/Payment.tsx
```diff
@@ -39,6 +41,19 @@ type Props = {
   canUpdatePaymentMethod: boolean;
   canRenewSubscription?: boolean;
   canShowReceipts?: boolean;
+  offers?: Offer[];
```
kmitrovv, line 44: This component has grown excessively complex and has become challenging to follow. What are your thoughts on breaking it down into smaller, more manageable pieces? Creating at least one additional component for handling the update, downgrade, and renewal functionality could significantly improve its readability and maintainability.
Considering the possible changes, it might involve relocating some of the component's logic into the container. If you believe that implementing these changes would require a lot of effort, perhaps we could leave it as is for the time being and include it in our future planning.

naumovski-filip: Discussed and will take up refactoring this component after the Profiles feature is finished.

kmitrovv (approved):

PR #321, opened 2023-06-22
Fix: [Placeholder] Message for simultaneous logins logout on web app

borkopetrovicc:
## Description

notion card: https://www.notion.so/jwplayer/Placeholder-Message-for-simultaneous-logins-logout-on-web-app-4328d1f4287d4291ab091e79bf49e126

This PR solves # .

Screenshot 2023-06-22 at 14 44 36

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


naumovski-filip (approved):
lgtm

public/locales/es/account.json
```diff
@@ -89,11 +89,10 @@
     "google": "Inicia sesión con Google",
     "hide_password": "Ocultar contraseña",
     "not_registered": "¿Nuevo/a en {{siteName}}?",
-    "ok": "Ok",
     "password": "Contraseña",
     "sign_in": "Iniciar sesión",
     "sign_up": "Registrarse",
-    "simultaneous_logins": "Se ha cerrado la sesión porque se ha alcanzado el límite de inicios de sesión simultáneos",
+    "simultaneous_logins": "Se ha alcanzado el número máximo de sesiones activas para esta cuenta. Iniciar sesión aquí lo cerrará en otra sesión o dispositivo.",
```
AntonLantukh, line 96: We should better modify Spanish translations in a separate PR (as agreed).

borkopetrovicc: This is machine translations.It wil have separate PR for the final translations.

test-e2e/tests/payments/subscription_test.ts
```diff
@@ -30,8 +30,6 @@ runTestSuite(jwProps, 'JW Player');
 runTestSuite(cleengProps, 'Cleeng');
 
 function runTestSuite(props: ProviderProps, providerName: string) {
-  let paidLoginContext: LoginContext;
```
budzworthy: Removing this means that every test here will get a new user created. It's probably not the end of the world in regard to creating junk data, but it does mean we aren't testing any scenarios for 'existing users'.  Was the simultaneous login feature causing issues?  If so, maybe we can better just add a "Logout" to the end of each test.

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -73,11 +72,10 @@ const AccountModal = () => {
         </div>
       );
     }
+
     switch (view) {
-      case 'simultaneous-logins':
-        return <SimultaneousLoginsNotification />;
       case 'login':
-        return <Login />;
+        return <Login message={message} />;
```
budzworthy: Good that you aren't passing the message to display in full in the query param, but I would suggest making message into an enum that the Login component understands so it can be expanded if needed (i.e. instead of "true" use "simultaneous_warning" or something.) We can parse the value here and have the props on the components be enum values, so it's more intuitive.

src/stores/NotificationsController.ts
```diff
@@ -26,7 +26,7 @@ export const subscribeToNotifications = async (uuid: string = '') => {
           case NotificationsTypes.ACCOUNT_LOGOUT:
             await logout();
             if (notification.resource?.reason === 'sessions_limit') {
-              window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
+              window.location.href = addQueryParams(window.location.href, { u: 'login', message: 'true' });
```
budzworthy: Do we not get a response from the `await login(formData.email, formData.password);` call? It would be nice not to have to rely on this. It's obscure to debug, harder to test, and risks race conditions, no?

test-e2e/tests/live_channel_test.ts
```diff
@@ -90,9 +90,6 @@ Scenario('I can watch the current live program on the live channel screen', asyn
   I.click('Start watching');
   I.seeElement('video');
 
-  // to make sure the back button is visible and can be clicked on
```
budzworthy: It looks like the live channel on the demo app is currently broken. I've reached out to the live channels team to help fix it for us. I think we need to keep this line when it's working.

test-e2e/tests/watch_history/logged_in_test.ts
```diff
@@ -10,41 +9,20 @@ runTestSuite(testConfigs.jwpAuth, testConfigs.jwpAuthNoWatchlist, 'JW Player');
...
 function runTestSuite(config: typeof testConfigs.svod, configNoWatchlist: typeof testConfigs.jwpAuthNoWatchlist, providerName: string) {
-  let loginContext: LoginContext;
-
   Feature(`watch_history - logged in - ${providerName}`).retry(Number(process.env.TEST_RETRY_COUNT) || 0);
 
   Before(({ I }) => {
     I.useConfig(config);
   });
 
-  Scenario(`I can get my watch history when logged in - ${providerName}`, async ({ I }) => {
```
budzworthy: Why did you just delete this test? Did you move the checks somewhere else? I don't see it anymore by name.

test-e2e/tests/watch_history/logged_in_test.ts
```diff
@@ -84,7 +63,7 @@ function runTestSuite(config: typeof testConfigs.svod, configNoWatchlist: typeof
   Scenario(`I do not see continue_watching videos on the home page and video page if there is not such config setting - ${providerName}`, async ({ I }) => {
     I.useConfig(configNoWatchlist);
 
-    await registerOrLogin(I);
+    I.registerOrLogin();
```
budzworthy: Don't you need an await here?

test-e2e/tests/watch_history/logged_in_test.ts
```diff
@@ -10,41 +9,20 @@ runTestSuite(testConfigs.jwpAuth, testConfigs.jwpAuthNoWatchlist, 'JW Player');
...
-
   Scenario(`I can get my watch history stored to my account after login - ${providerName}`, async ({ I }) => {
     I.dontSee(constants.continueWatchingShelfTitle);
 
     await I.openVideoCard(videoTitle);
     I.dontSee(constants.continueWatchingButton);
     I.see(constants.startWatchingButton);
 
-    await registerOrLogin(I);
+    await registerAndPlayVideo(I, videoTitle);
```
budzworthy: This test has been changed subtly but significantly here. In the old version it tests that a user's saved watch history is restored after logging in on a fresh browser. This test only tests that it is saved within the same session. The restore isn't tested.

It is possible to make this test the intended tests but you'll end up duplicating a lot of logic and adding a non-trivial amount of time to the test runs. I'm also not sure how you can actually fully clean the browser session within a single test to make sure the values are actually getting restored and not lingering in session data, cookies, etc.

Maybe it's possible to do something with marking subsequent scenarios as skipped if the previous scenarios fail? Then we can leave the dependency which in this suite we actually make really important use of.

test-e2e/tests/watch_history/logged_in_test.ts
```diff
@@ -62,7 +40,8 @@ function runTestSuite(config: typeof testConfigs.svod, configNoWatchlist: typeof
   Scenario(`I can see my watch history on the Home screen when logged in - ${providerName}`, async ({ I }) => {
     I.dontSee(constants.continueWatchingShelfTitle);
 
-    await registerOrLogin(I);
+    await registerAndPlayVideo(I, videoTitle);
```
budzworthy: You've also changed this test so it's not testing the restoration, only the storing within a session. Looking at the checks, this test can probably be combined with the one above to save some overhead.

test-e2e/tests/payments/subscription_test.ts
```diff
@@ -144,7 +143,7 @@ function runTestSuite(props: ProviderProps, providerName: string) {
   });
 
   Scenario(`I can open the PayPal site - ${providerName}`, async ({ I }) => {
-    paidLoginContext = await I.registerOrLogin(paidLoginContext);
+    await I.registerOrLogin();
```
budzworthy: For all of the tests up to this point the `resgisterOrLogin` logic shouldn't cause any sort of cascading failures or weird dependencies. Really all it's doing is A) minimizing junk data creation and B) minimizing test run time and overhead by skipping the extra 'register' step that is part of creating a new user.

Are there beneficial reasons to refactor these?

test-e2e/tests/payments/subscription_test.ts
```diff
@@ -157,26 +156,13 @@ function runTestSuite(props: ProviderProps, providerName: string) {
...
-
-    await finishAndCheckSubscription(I, addYear(today), today, props.yearlyOffer.price);
+    await registerAndSubscribe(I, props, today);
 
     I.seeAll(cardInfo);
   });
 
   Scenario(`I can cancel my subscription - ${providerName}`, async ({ I }) => {
-    paidLoginContext = await I.registerOrLogin(paidLoginContext);
+    await registerAndSubscribe(I, props, today);
```
budzworthy: Again, these tests have a subtle change. Before your changes it makes sure that subscription data and invoices are properly reloaded when logging back in. With your changes it just tests that it's initially populated properly when the subscription is first created, and you would have to fully clear the browser to test otherwise. I think the former is a more high traffic use case (the initial case only happens once, and the subsequent login and viewing happens every other time.) Honestly it's valuable to test both scenarios.

Maybe you can combine these last 4 scenarios into a single flow (create user, finish subscription, cancel subscription, renew subscription, view invoices.) And run it once with a single login and then a second time with logging out and clearing the browser between each step? Then you have the overhead of 2 registrations and subscription creations instead of 4 and you are actually testing more functionaily.

test-e2e/utils/payments.ts
```diff
@@ -129,3 +130,24 @@ export function overrideIP(I: CodeceptJS.I) {
...
 }
+
+export async function registerAndSubscribe(I: CodeceptJS.I, props: ProviderProps, today: Date) {
+  const user = await I.registerOrLogin();
+
+  await goToCheckout(I);
+
+  I.waitForElement('#card', longTimeout);
+
+  I.payWithCreditCard(
```
budzworthy: It looks like there's a missing await here (and also in the code you moved it from in `I can finish my subscription with credit card`.) That's probably why the tests were failing in the first place and refactoring here probably just happens to luckily tweak the operations so that it masks the race condition.

Do you use an IDE? I realize that this is not your code, but an IDE should flag things like this to help you when you are debugging. I highly recommend jetbrains products. Either way, I will look to improve our code inspection rules to prevent code being committed that has oversights like this.

budzworthy (changes requested):
I understand the learning curve with the tests sharing the same user, but I don't see a strong argument for changing it and in fact it results in adding boilerplate, making tests take longer, and reducing the coverage of actual flows. For the 4 subscription tests that all depend on one another I would be OK combining them provided you follow the suggestion for testing with logout and clearing the browser between steps.

test-e2e/tests/watch_history/logged_in_test.ts
naumovski-filip: I believe going off this part of the documentation that I.* calls don't necessarily need to be awaited (https://codecept.io/basics/#waiting) but correct me if I misunderstood

test-e2e/utils/payments.ts
naumovski-filip: Same as the other comment, I believe that it's not necessary to await I.* calls unless the return value from them is needed

test-e2e/tests/watch_history/logged_in_test.ts
naumovski-filip: Based on the steps and the changes we made to the test below it it looked like the bottom test included everything tested in the top test

test-e2e/tests/live_channel_test.ts
naumovski-filip: The I.click call by itself makes sure the element is visible, no? If it isn't, the test is just going to fail, regardless of what's causing it. There were issues with the actual playback though yes, but I don't believe it makes a difference. The back button is also visible even when there's an error in playback (which could be a real use case) so it should be tested in those cases as well

test-e2e/tests/payments/subscription_test.ts
naumovski-filip: The extra runtime and junk data is imo justified by the modularity and readability gained by having all tests actually be e2e, and it cuts down on future development time. If one of these scenarios changes and you have to change the test, you don't need to wait for 4 tests to run with all the necessary setup just to get to the one test that you're working on. Just my 2c

test-e2e/tests/watch_history/logged_in_test.ts
budzworthy: I'm pretty sure this would only apply if your entire logic flow was a series of `I.` calls, which ours is not.

"If you want to get information from a running test you can use await inside the async function, and utilize special methods of helpers started with the grab prefix."

If you dig down into the methods called inside `registerOrLogin`, you'll eventually find at least 1 `grab` which I think means dropping the await is going to cause problems.

Looks like you can probably remove these 2 awaits:
- https://github.com/jwplayer/ott-web-app/blob/develop/test-e2e/utils/steps_file.ts#L76 - kind of a silly check to worry about. I would say just delete the line
- https://github.com/jwplayer/ott-web-app/blob/develop/test-e2e/utils/steps_file.ts#L291 - this one I'm unsure of exactly how it works properly. I think the usePlaywrightTo method used to allow a return value but no longer does. It would be really nice to find a way to replace this with an injected param instead of checking it each time

Then I think a lot fewer tests will need to use async and await, but when you see it, it's probably a good idea to wait for it.

test-e2e/utils/payments.ts
budzworthy: Also worth noting that these are custom `I` extension methods, so you really have to look at the internals of them before assuming something from the docs applies.

test-e2e/tests/watch_history/logged_in_test.ts
budzworthy: Well sure, but as noted below, that test is no longer testing what it says it is. I think we need to test both behaviors, so please restore this one and revert the changes in the one below.

test-e2e/tests/live_channel_test.ts
budzworthy: The first `click` is clicking on the video frame to make sure the controls are rendered (they fade out after a period of inactivity)

test-e2e/tests/payments/subscription_test.ts
budzworthy: Any of these tests can be run independently or in any order:
- `I can see my payments data `
- `I can see offered subscriptions`
- `I can choose an offer`
- `I can see payment types`
- `I can open the PayPal site`

These tests are the 4 that depend on each other
- `I can finish my subscription with credit card`
- `I can cancel my subscription`
- `I can renew my subscription`
- `I can view my invoices`

We could probably combine those 4 or rewrite them in a way to better support 1-off runs if necessary, but you'd either lose resolution on test results, or still have the same problem that a bug in subscription creation also causes the cancel test to fail, etc. But in this case, since you've made it run the same flow 4 times, you've also added 2 minutes to the total runtime.

budzworthy:
We can go back and forth about the pros and cons of the sequential tests for days, but what I really want to understand is what were the actual bug fixes for the tests that were failing?

Let's scale this PR down to just those and we can take the conversation for how we might make the test scenarios more developer friendly to another forum.

src/containers/AccountModal/AccountModal.tsx
borkopetrovicc: In the paymentFailed component it is approached in the same way see below: 

Screenshot 2023-07-03 at 13 18 58
Screenshot 2023-07-03 at 13 18 40

Should i also refactor this component ?

src/stores/NotificationsController.ts
borkopetrovicc: Can you give me example how to manage these line of code ?

src/containers/AccountModal/AccountModal.tsx
budzworthy: Yes if you have time to do that it would be great, otherwise, at least make sure the new code you are adding is good

src/stores/NotificationsController.ts
budzworthy: I'm not sure what the options are... it depends on the architecture of the inplayer sdk and backend, but I would assume that the login method called here: https://github.com/jwplayer/ott-web-app/blob/develop/src/containers/AccountModal/forms/Login.tsx#L21

Should be able to inspect the return value to see if the login was failed for concurrent sessions or not and set to show the appropriate notification without involving the notification controller.

darkoatanasovski: hey,

to add some context to this:

The simultaneous logins feature manages how many active sessions can have the same account( if the customer set the number greater than 0 in the dashboard ). 
On every login action, we check that number and when the limit is reached, **we invalidate the first session and send WebSocket notification.** 
The only way to know the reason for logout is to check the payload of the `account.logout` WebSocket notification, exactly like this code.

the only change that I am suggesting here is to change the `message` param( true/false ) with something like `reason` parameter;

e.g. https://..../?u=login&reason=sessions_limit
so, we can handle and validate `reason` parameter in the application.

budzworthy: Ah, that makes sense. I didn't have the context and thought that this was the message on the login dialog that shows before your other sessions are logged out. I agree that an enum rather than a boolean or a raw string is good for the param type.

borkopetrovicc: d4bafdf64cf379ef4f3037c94fac29b47a6528ef

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -73,11 +73,10 @@ const AccountModal = () => {
         </div>
       );
     }
+
     switch (view) {
-      case 'simultaneous-logins':
-        return <SimultaneousLoginsNotification />;
       case 'login':
-        return <Login />;
+        return <Login message={message as LoginFormMessage} />;
```
ChristiaanScheermeijer: Casting the message here to `LoginFormMessage` results in more unpredictable behavior. There is no guarantee that `message` is of type `'simultaneous_warning' | undefined` and still means that any string can end up in the `message` prop. However, the `<LoginForm />` doesn't know any better.

There are two ways of fixing this. The most straightforward fix is passing it as `string | undefined`. This doesn't change the implementation, but you're not assuming the `message` prop is a string literal.

The second approach is to validate the message before casting it. If you're only interested in the string literal or undefined, you could do something like this:

```ts
type LoginFormMessage = undefined | 'simultaneous_warning';

// this parser ensures that the type validates to `LoginFormMessage`
const parseLoginFormMessage = (message?: string): LoginFormMessage => {
  return message === 'simultaneous_warning' ? message : undefined;
};

// ....

case 'login':
  return <Login message={parseLoginFormMessage(message)} />;
```

Remember, when adding an `as ...` in your code, there is most likely a better solution.

budzworthy: I agree with the sentiment, but would recommend a slightly different solution. Since we probably want to put the translation tags all into a switch statement anyhow so the i18n parser picks them up properly, why don't we let the message get passed as a string into the Login component and then have a switch statement in there like this:
```
const getMessage = (messageId: string) => {
    if (messageId === 'simultaneous_warning') {
      return t('simultaneous_logins');
    }

    return t('unexpected_login_error');
  }
```

ChristiaanScheermeijer: Yes, I also like that because you won't need to use the `LoginFormMessage` type in several places.

The only misaligned could be that the string literal is not set correctly in the NotificiationsController, but I think you already mentioned (in a different thread) that using a generic constant could solve this problem:

```ts
export const simultaneousLoginsFlag = 'simultaneous_logins';
```

src/components/LoginForm/LoginForm.tsx
```diff
@@ -19,6 +19,14 @@ import { addQueryParam } from '#src/utils/location';
...
 import type { LoginFormData } from '#types/account';
 
+const LOGIN_FORM_MESSAGES = ['simultaneous_logins'] as const;
+
+export type LoginFormMessage = (typeof LOGIN_FORM_MESSAGES)[number];
+
+const isValidLoginFormMessage = (message: string | undefined): message is LoginFormMessage => LOGIN_FORM_MESSAGES.includes(message as LoginFormMessage);
+
+export const getLoginFormMessage = (message: string | undefined): LoginFormMessage | undefined => (isValidLoginFormMessage(message) ? message : undefined);
+
```
ChristiaanScheermeijer: @dbudzins @borkopetrovicc did you perhaps agree on this outside the review comments? I thought the conclusion was to not cast the message to `LoginFormMessage` but just pass it as a string.

borkopetrovicc: No, we didn't consult outside of here. I tried to make my solution better by making it as type safe as possible.

budzworthy (dismissed):
Can you please try to figure out what happened with the snapshots? I can't really review this efficiently with all the no-op changes there. If it's something that the newest version of vitest is adding, please only regenerate the ones that you've actually changed and revert the others, or update all of them and submit that as a separate chore PR.

src/components/LoginForm/LoginForm.tsx
budzworthy: This is a lot of boilerplate that we don't really need and is actually making the components less scalable imo. Strict type enforcement is not always the best choice. Can you please give a go at the solution suggested in this conversation: https://github.com/jwplayer/ott-web-app/pull/321#discussion_r1260197777

borkopetrovicc:
> Can you please try to figure out what happened with the snapshots? I can't really review this efficiently with all the no-op changes there. If it's something that the newest version of vitest is adding, please only regenerate the ones that you've actually changed and revert the others, or update all of them and submit that as a separate chore PR.

I reverted the changes 6656ad790806a97b96a1d755a6c8d4c62b0eb497 and i will make new chore pr with updated snapshots.

src/components/LoginForm/LoginForm.tsx
borkopetrovicc: 967073ed87863665ef7350fe545aba8b5be69efb

budzworthy:
@borkopetrovicc can you do a quick merge from develop into this branch? I'm still seeing all the test files as having changed.

src/stores/NotificationsController.ts
```diff
@@ -26,7 +26,8 @@ export const subscribeToNotifications = async (uuid: string = '') => {
           case NotificationsTypes.ACCOUNT_LOGOUT:
             await logout();
             if (notification.resource?.reason === 'sessions_limit') {
-              window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
+              const message: string = 'simultaneous_logins';
```
budzworthy: Suggest you move this const to a top level export const in LoginForm.tsx

src/services/inplayer.account.service.ts
```diff
@@ -120,7 +120,7 @@ export const logout = async () => {
     InPlayer.Notifications.unsubscribe();
     await InPlayer.Account.signOut();
   } catch {
-    throw new Error('Failed to sign out.');
+    console.error('Failed to sign out.');
```
budzworthy: Why did you change this? Aren't there cases where upstream code will want to give feedback to viewers when the logout request fails?

src/containers/AccountModal/forms/Login.tsx
```diff
@@ -10,7 +10,11 @@ import { removeQueryParam } from '#src/utils/location';
 import type { LoginFormData } from '#types/account';
 import { login } from '#src/stores/AccountController';
 
-const Login = () => {
+type Props = {
+  message?: string;
```
budzworthy: It might be nice to rename this prop to `messageKey` or `messageParam` or something to make it more clear that it's not the actual string that will be displayed, rather a lookup key. You can rename here, in LoginForm, and in AccountModal.

src/stores/NotificationsController.ts
borkopetrovicc: 892317022b80decbf8cb56b9749447673a899c88

src/containers/AccountModal/forms/Login.tsx
borkopetrovicc: 892317022b80decbf8cb56b9749447673a899c88

src/services/inplayer.account.service.ts
borkopetrovicc: We don't catch this error and that's the reason it breaks the simulations logins functionality.

src/components/LoginForm/LoginForm.tsx
```diff
@@ -27,15 +27,32 @@ type Props = {
   values: LoginFormData;
   submitting: boolean;
   siteName?: string;
+  messageKey?: string;
 };
 
-const LoginForm: React.FC<Props> = ({ onSubmit, onChange, values, errors, submitting, siteName }: Props) => {
+export const messageId: string = 'simultaneous_logins';
```
budzworthy:
```suggestion
export const simultaneousLoginWarningKey = 'simultaneous_logins';
```

src/components/LoginForm/LoginForm.tsx
```diff
@@ -27,15 +27,32 @@ type Props = {
...
+
+const LoginForm: React.FC<Props> = ({ onSubmit, onChange, values, errors, submitting, siteName, messageKey }: Props) => {
   const [viewPassword, toggleViewPassword] = useToggle();
   const { t } = useTranslation('account');
   const location = useLocation();
 
+  const getTranslatedErrorMessage = (messageId: string | undefined) => {
+    switch (messageId) {
+      case 'simultaneous_logins':
```
budzworthy:
```suggestion
      case simultaneousLoginWarningKey:
```

src/stores/NotificationsController.ts
```diff
@@ -26,12 +27,12 @@ export const subscribeToNotifications = async (uuid: string = '') => {
           case NotificationsTypes.ACCOUNT_LOGOUT:
             await logout();
             if (notification.resource?.reason === 'sessions_limit') {
-              window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
+              window.location.href = addQueryParams(window.location.href, { u: 'login' });
```
budzworthy: It looks like the simultaneous-logins message param got moved... was that intentional?

src/services/inplayer.account.service.ts
budzworthy: Can you please elaborate on that? I don't like burying errors like this.

src/stores/NotificationsController.ts
borkopetrovicc: No it was mistake 4b1070fd500fadb5b4f62390f3c44b332d18b7f8

src/services/inplayer.account.service.ts
borkopetrovicc: When we get logged out because of simultaneous logins, the token is invalidated and the .signOut() network call fails due to providing an invalid token, causing the function to go into the catch part. If we just throw the error again there is an uncaught error and the redirect doesn't happen

budzworthy: By changing this, you are changing not just the simultaneous logout flow, but also the normal logout flow. The user might try to sign out and have it fail, then there's no longer any way for us to give feedback.

If logout always fails in this scenario why are we even calling it? Can we just call `InPlayer.Notifications.unsubscribe`? Or are there other things happening inside of `InPlayer.Account.signOut`?

darkoatanasovski: this is only the case for simultaneous logins. When the simultaneous login limit is reached on the user login, we invalidate the user's token ( first session for the account ), and then, we send the notification to the user's device. Because the token is already invalidated, the request for logout fails( because of an invalid token ). I will try to find a better solution today.

@borkopetrovicc Maybe we can just update the state on logout notification?!

e.g.
```
         
case NotificationsTypes.ACCOUNT_LOGOUT:
await logout(true);
if (notification.resource?.reason === 'sessions_limit') {
   window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
}
break;
```

and extend the logout method from the controller

e.g.
```
export async function logout(onlyState = false) {
  await useService(async ({ accountService }) => {
    // this invalidates all entitlements caches which makes the useEntitlement hook to verify the entitlements.
    await queryClient.invalidateQueries('entitlements');

    useAccountStore.setState({
      user: null,
      subscription: null,
      transactions: null,
      activePayment: null,
      customerConsents: null,
      publisherConsents: null,
      loading: false,
    });

    await restoreFavorites();
    await restoreWatchHistory();
    if (onlyState) {
      await accountService?.logout();
    }
  });
}
```

filip-naumovski: The case should be if (!onlyState) (with "!"), but other than that this seems like the best solution. Maybe better to call it sendNetworkRequest or something similar and have it be true by default

naumovski-filip: The case should be if (!onlyState) (with "!"), but other than that this seems like the best solution. Maybe better to call it sendNetworkRequest or something similar and have it be true by default

darkoatanasovski: or even better, just move the logout part at the bottom

e.g.

```
case NotificationsTypes.ACCOUNT_LOGOUT:
if (notification.resource?.reason === 'sessions_limit') {
   window.location.href = addQueryParams(window.location.href, { u: 'simultaneous-logins' });
}
await logout();
break;
```

naumovski-filip: Won't logout not be called because of using window.location.href in this case?

darkoatanasovski: correct, but on the page load, the user won't be authenticated because of the invalid session.

borkopetrovicc: I've tested and it is working ok. 95c9032c0c6cbe5a1e786fd357085f0108801488

ChristiaanScheermeijer: I think that with this function order, you can't guarantee that `await logout();` will be finished. This might not be a problem in the current situation, but I don't think relying on flaky behavior is a good idea (even though it still works).

The usage of `window.location.href` should still be discussed because it fully reloads the SPA, which isn't ideal for UX, PWA and Capacitor apps.

borkopetrovicc: 45ec68a6df7215be2dfa5bfcbbaac6383f4aef50

ChristiaanScheermeijer (approved):
I have no other feedback :-)

Just a tip, I would use the [Github Squash & Merge](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#squashing-your-merge-commits) option to prevent cluttering the main branch with a lot of rework commits. 

You can then also rewrite the commit message to `refactor(auth): simultaneous login message from login component` to ensure that the changelog is (more) correct.

src/services/inplayer.account.service.ts
budzworthy: Can we put the logout into an else clause? Then it's more explicit that we are not intending to run logout when we refresh the page for simultaneous logout. A comment could also be nice.

borkopetrovicc: df8d70e2ad529888f73e9d1b9ca1a78db4d1a555

borkopetrovicc:
> snapshots?

This has been fixed, and it is still in change request.

PR #328, opened 2023-07-05
feat(project): use a native fallback for image service

AntonLantukh:
## Description

- Use a native fallback for additional images (via `poster_fallback=1` query)
- Store images as a simple string for media items (instead of ImageData object)
- Add backward compatibility for ImageData structure for channels / programs not to force customers to update all their schedules

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/services/api.service.ts
```diff
@@ -9,31 +9,33 @@ import type { GetPlaylistParams, Playlist, PlaylistItem } from '#types/playlist'
...
   const config = ConfigStore.getState().config;
 
   const offerKeys = Object.keys(config?.integrations)[0];
 
   const transformedMediaItem = {
     ...item,
-    shelfImage: generateImageData(config, ImageProperty.SHELF, item, playlist),
-    backgroundImage: generateImageData(config, ImageProperty.BACKGROUND, item),
-    channelLogoImage: generateImageData(config, ImageProperty.CHANNEL_LOGO, item),
+    shelfImage: generateAlternateImageURL(item, ImageProperty.CARD),
```
ChristiaanScheermeijer: Should we rename the `shelfImage` property as well?

```suggestion
    cardImage: generateAlternateImageURL(item, ImageProperty.CARD),
```

src/services/epg.service.ts
```diff
@@ -20,8 +20,8 @@ export type EpgChannel = {
   id: string;
   title: string;
   description: string;
-  channelLogoImage: ImageData;
-  backgroundImage: ImageData;
+  channelLogoImage: string | ImageData;
+  backgroundImage: string | ImageData;
```
ChristiaanScheermeijer: Is `ImageData` still valid here?

AntonLantukh: We use it in EPG schedule. In case we change the format, it looks like some customers will have to update their data for it to continue working. My idea was to provide backward compatibility.

ChristiaanScheermeijer (commented):
Nice cleanup. I've left two small comments/questions

src/services/epg.service.ts
ChristiaanScheermeijer: The `ImageData` is composed in the `transformProgram` method, so we should be able to remove `ImageData` completely. The API response only contains an `image` property which we wrap in an object.

https://github.com/jwplayer/ott-web-app/blob/4d8ff39439e22f2e7d13ac7e73eb92a12758ddb3/src/services/epg.service.ts#L100

src/components/CardGrid/CardGrid.tsx
```diff
@@ -55,7 +55,7 @@ function CardGrid({
   onCardHover,
 }: CardGridProps) {
   const breakpoint: Breakpoint = useBreakpoint();
-  const posterAspect = parseAspectRatio(playlist.shelfImageAspectRatio);
+  const posterAspect = parseAspectRatio(playlist.aspectRatio || playlist.shelfImageAspectRatio);
```
AntonLantukh: @ChristiaanScheermeijer Do you know any cases where `shelfImageAspectRatio` is used by customers? I would just leave `aspectRatio`.

budzworthy: I don't mind keeping the more verbose param. What was the reason for the name change in the first place? Just `aspectRatio` might imply the sizing of the shelf itself.

AntonLantukh: We use now `cardImage` instead of `shelfImage` as we changed the label to `card` by default (AC). Previously `shelfImage` custom param was used to define the label. I can also use `cardImageAspectRatio`.

ChristiaanScheermeijer: I don't think so. We're using this in the demo app configs, though. But if you are going to change it and remove the fallback, it would be good to mention it in the changelogs.

ChristiaanScheermeijer (approved):

src/components/CardGrid/CardGrid.tsx
AntonLantukh: Left the fallback with`shelfImageAspectRatio` (deprecated), added  `cardImageAspectRatio`

PR #331, opened 2023-07-06
fix(seo): fix seo score

AntonLantukh:
## Description

Fix seo score

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


AntonLantukh:
Looks like firebase adds a `no-index` header for PR envs:

Screenshot 2023-07-06 at 14 10 37

public/locales/en/demo.json
```diff
@@ -1,5 +1,6 @@
 {
   "app_config_not_found": "Invalid or Missing App Config",
+  "app_config_learn_more": "Learn more about the App Config creation",
```
budzworthy:
```suggestion
  "app_config_learn_more": "Learn more about App Configs",
```

public/locales/es/common.json
```diff
@@ -12,6 +12,7 @@
     "confirm": "Sí"
   },
   "default_site_name": "Mi aplicación OTT",
+  "default_description": "",
```
budzworthy, line 15: Did you create a separate PR with these translations?

src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -136,16 +137,24 @@ const DemoConfigDialog = ({ selectedConfigSource, configQuery }: Props) => {
...
+      {!configQuery.isSuccess && (
+        <Helmet>
+          <meta name="description" content="Enter the App Config ID" />
+        </Helmet>
+      )}
       {configQuery.isSuccess && (
         <div className={styles.note}>
           <div>{t('currently_previewing_config', { configSource: selectedConfigSource })}</div>
-          <Link onClick={clearConfig}>{t('click_to_unselect_config')}</Link>
+          <Link href="javascript:;" onClick={clearConfig}>
```
budzworthy: Maybe we should just make this a Button with variant = text?

budzworthy (commented):
It's probably good to fix some things on the demo config dialog, but what we really want to test is the home page with all the shelves. Can you also try to get the lhci action working appropriately with that?

public/locales/en/demo.json
AntonLantukh: Will try to change, but the SEO score was lower because of this text
Screenshot 2023-07-12 at 09 47 22

budzworthy: Looks like "Learn more" specifically is one of the generic flagged phrases: https://developer.chrome.com/en/docs/lighthouse/seo/link-text/

src/components/DemoConfigDialog/DemoConfigDialog.tsx
AntonLantukh: Fixed it with `variant=text`, though styling changed a bit.

budzworthy: Specific styling on that was never anything intentional. I think it looks fine.

AntonLantukh:
@dbudzins I have added Lighthouse workflow to the Preview workflow as a separate job with only one (SEO) required check.

I have also increased `numberOfRuns` to 3 to improve the accuracy. One more thing I spotted is that a server created 'on the go' within Lighthouse workflow is quite slow and doesn't reflect actual performance results.

I decided to use a separate job in the same workflow to simplify the way we can pass the link. I tried several approaches:

1.  I used `workflow_dispatch` in the Lighthouse workflow and triggered it from the Preview workflow manually using `gh` cli. It worked, but the check didn't appear in the PR.

2. I used `workflow_run`, but according to the docs "This event will only trigger a workflow run if the workflow file is on the default branch." 

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

"Default" one is a develop branch and it adds some complexity to debug and develop such workflow as the version of the default branch is taken. And I am not sure that the PR check will be present in this case.

```
  workflow_run:
    workflows: [Firebase Preview]
    types:
      - completed
```

3. I used `lewagon/wait-on-check-action@v1.3.1` package to wait for Preview job in the Lighthouse workflow and continue when preview link is present. I uploaded an artifact in Preview job and wanted to get acces to in Lighthouse job. However it didn't work out of the box. By default, you can only upload / download an artifact within one workflow. There is a custom job to make it possible within different workflows but it already adds too much complexity.

```
    steps:
      - name: Wait for preview link to be created
        uses: lewagon/wait-on-check-action@v1.3.1
        with:
          ref: ${{ github.event.pull_request.head.sha }}
          check-name: build_and_preview
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          wait-interval: 10
          allowed-conclusions: success,skipped,cancelled
          verbose: true
      - name: Grab preview link
        uses: actions/download-artifact@v3
        with:
          name: preview_link
      - name: Set env
        run: echo "PREVIEW_LINK=$(cat preview_link.txt)" >> $GITHUB_ENV
      - uses: actions/checkout@v2
      - name: Install Lighthouse CI
        run: sudo yarn global add @lhci/cli@0.12.x
      - name: Run Lighthouse CI
        run: lhci autorun --collect.url=${{ env.PREVIEW_LINK }}?app-config=gnnuzabk --config=./lighthouserc.js
```

budzworthy:
@AntonLantukh OK, so basically to summarize you are running the lighthouse SEO check on the preview build, right? That's a really clever way to do this. Nice work figuring it out and implementing it!

budzworthy (approved):
Nice work and clever solution to the lighthouse runs.

PR #332, opened 2023-07-06
feat(series): add native support for next episodes

AntonLantukh:
## Description

- `after_id` query param is added
-  complex front-end related logic is removed

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/services/api.service.ts
```diff
@@ -149,13 +149,20 @@ export const getSeriesByMediaIds = async (mediaIds: string[]): Promise<{ [mediaI
  * Get all episodes of the selected series (when no particular season is selected or when episodes are attached to series)
  * @param {string} seriesId
  */
-export const getEpisodes = async (seriesId: string | undefined, pageOffset: number, pageLimit?: number): Promise<EpisodesWithPagination> => {
+export const getEpisodes = async (
+  seriesId: string | undefined | null,
+  pageOffset: number | null,
+  pageLimit: number | null = PAGE_LIMIT,
+  afterId?: string | null,
```
budzworthy: I always like to use named params when there are this many inputs of native types to a function

src/services/api.service.ts
```diff
@@ -149,13 +149,20 @@ export const getSeriesByMediaIds = async (mediaIds: string[]): Promise<{ [mediaI
...
+): Promise<EpisodesWithPagination> => {
   if (!seriesId) {
     throw new Error('Series ID is required');
+  } else if (pageOffset && afterId) {
```
budzworthy: I would suggest we either let the BE handle this or enforce it with typings on the arguments:
```
args: {seriesdId: string} & ({pageOffset: number, pageLimit: number} | {afterId: string})
```

src/services/api.service.ts
AntonLantukh: Was thinking about it, but wanted to follow the approach we already have for other endpoint. Changed it for episodes and seasons endpoints.

src/services/api.service.ts
AntonLantukh: Removed this check

budzworthy (approved):

AntonLantukh:
Waiting for back-end related fixes

PR #340, opened 2023-07-14
Fix / watch history bugs

ChristiaanScheermeijer:
## Description

We've found a problem in the watch history implementation when signed URLs and the inline layout are enabled. The problem was that the wrong item was saved to the watch history. This was because one of the triggers uses the current `item`, but this already contains the "next" item and stores it anyways with the current progress.

This resulted in:

- The watch progress is not being saved for the correct item
- The next item starts playing at the same position as the previous item

While adding this feature I also discovered a bug in the `watchHistoryController#saveItem` function. This currently only saves the latest item. There is a fix for that as well in this PR.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


src/hooks/useWatchHistory.ts
```diff
@@ -0,0 +1,123 @@
...
+  // config
+  const { features } = useConfigStore((s) => s.config);
+  const continueWatchingList = features?.continueWatchingList;
+  const watchHistoryEnabled = !!continueWatchingList;
+
+  // watch History
+  const watchHistoryItem = useWatchHistoryStore((state) => (!!item && watchHistoryEnabled ? state.getItem(item) : undefined));
+
+  // update the queued watch progress on each time update event
+  const handleTimeUpdate = useEventCallback(() => {
```
budzworthy: Do player and item need to be dependencies for this callback? Or are they global / never change?

ChristiaanScheermeijer: I use `useEventCallback` here to prevent the callback from updating when the `item` changes because it will re-trigger the `useEffect` that (un)binds the event listeners. We had problems with that before regarding certain events being missed.

The `useEventCallback` doesn't take any dependencies and will always use the latest state/variables.

src/hooks/useWatchHistory.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!player) return;
+
+    const progress = player.getPosition() / item.duration;
```
budzworthy: Can we get any of this info from the event data? https://docs.jwplayer.com/players/reference/seek-events-1#ontime

It looks like the progress is in the event. Maybe there's a chance that the metadata property will also contain the item being played and then we can have higher confidence that there won't be race conditions.

ChristiaanScheermeijer: Yes, that's a good observation. I also thought about that, but I couldn't remember why we're using the `item.duration` instead of the player duration.

We can use the event params instead. Initially, I tried using the [getPlaylistItem](https://docs.jwplayer.com/players/reference/getplaylistitem_index_) API to get the current playing item, but JW Player sanitizes the item making it incompatible with the PlaylistItem type we're using in the web app.

I also see an undocumented `absolutePosition` property in the event object. I think this can be used for live events.

ChristiaanScheermeijer: I just tested a live event:

image

ChristiaanScheermeijer: I've updated the event handler

src/hooks/useWatchHistory.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  useEffect(() => {
+    return () => {
+      // store watch progress on unmount and when the media item changes
+      maybeSaveWatchProgress();
+    };
+  }, [item?.mediaid, maybeSaveWatchProgress]);
+
+  // add event listeners for unload and visibility change to ensure the latest watch progress is saved
+  useLayoutEffect(() => {
```
budzworthy: From what I can tell, onunload and onvisibilitychange are full of small differences and bugs between browsers, especially on older and mobile versions.

Based on these 2 pages:
- https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/
- https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event

it looks like we might also want to include the pagehide event for fullest coverage. How would you feel about also saving the progress on a timeout every couple minutes or so? Even if it's only every 5 minutes, it would greatly reduce our risk of losing progress for longer content.

src/hooks/useWatchHistory.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * The `useWatchHistory` hook has two responsibilities.
+ *
+ * 1. Determine the `startTime` of the given `item` (return value)
+ * 2. Save the watch progress of the current `item` to the user profile
+ *
+ * __The problem:__
+ *
+ * There are multiple events that trigger a save. This results in duplicate (unnecessary) saves and API calls. Another
+ * problem is that some events are triggered when the `item` to update has changed. For example, when clicking a media
```
AntonLantukh:
```suggestion
 * There are multiple events that trigger a save. This results in duplicate (unnecessary) saves and API calls.
 * Another problem is that some events are triggered when the `item` to update has changed. For example, when clicking a media
```

src/hooks/useWatchHistory.ts
```diff
@@ -0,0 +1,124 @@
...
+    window.addEventListener('visibilitychange', visibilityListener);
+
+    return () => {
+      window.removeEventListener('beforeunload', maybeSaveWatchProgress);
+      window.removeEventListener('visibilitychange', visibilityListener);
+    };
+  }, [maybeSaveWatchProgress]);
+
+  // calculate the `startTime` of the current item based on the watch progress
+  return useMemo(() => {
```
budzworthy, line 23: I feel like this should probably be split into 2 separate hooks. In order to make sure that the listener hook always gets called, it's fine to call it inside of the useWatchHistory hook, but I think it would be nice to have everything from about approximately Lines 49 to 111 encapsulated in a useWatchHistoryListener hook.

src/hooks/useWatchHistory.ts
budzworthy: Jammer that the item isn't available, but I think it's probably safe this way.

src/hooks/useWatchHistory.ts
ChristiaanScheermeijer: Yeah, no problem! I've added the `pagehide` event listener and a progressive save function.

src/hooks/useWatchHistory.ts
ChristiaanScheermeijer: Done!

src/hooks/useWatchHistory.ts
ChristiaanScheermeijer: I found another bug. Due to the `watchHistoryController#saveItem` method always calling the `getSeriesByMediaIds`, all page hide / visibility change events are lost. This is because the XHR request is canceled (or never fulfilled) and the localStorage is never updated.

But I also realize that when using Cleeng/JWP the save call might also be skipped in this scenario...

budzworthy: @AntonLantukh is it OK for us to remove that call to fix this behavior? Shouldn't we be able to get the series ID from somewhere else?

ChristiaanScheermeijer: We could drill the `seriesId`; `MediaSeries` -> `InlinePlayer`/`Cinema` -> `PlayerContainer`

AntonLantukh: Hm, an it solve the problem with calling `persistWatchHistory` which is also an async one? We need a series item to create a 'series' card in watch history instead of an episode's one.

ChristiaanScheermeijer: Perhaps we shouldn't be relying on these events at all? The progress is now saved when pausing and progressively every 5 minutes. 

We should be using the `sendBeacon` API for beforeunload/visibilitychange when the user closes the tab or browser to reliably save the progress. But this requires an API that can handle all scenarios and doesn't work for localStorage.

When I remove the `getSeriesByMediaIds` call, it does work when saving the progress to localStorage. But I don't think it works when using Cleeng/JWP.

budzworthy: I haven't dug deep into this, just a quick google, but maybe we can add keep-alive to the api call?
https://stackoverflow.com/questions/60530716/execute-server-api-call-on-beforeunload-limitations-on-page-unload

Otherwise, periodic updates is fine with me, but I'd make them closer to every minute instead if it's all we have. Or maybe even just fire off one every time we get a time event with some logic to keep them from backing up?

ChristiaanScheermeijer: `keepalive` seems to work for Cleeng 🙌 . For InPlayer we need to modify the Axios config, do you know if that is possible?

budzworthy: Neat! It should be. We've made some other small tweaks to the SDK, but I'll have to check with the team.

ChristiaanScheermeijer:
@dbudzins @AntonLantukh I've refactored the get series logic to prevent async logic while closing the tab/browser. 

The Cleeng `updateCustomer` now uses `keepalive: true`, which successfully saves the progress while closing the tab/browser.

The `beforeunload` event listener is superseded by the `pagehide` and `visibilitychange` listeners. With this change, we're one step closer to supporting the [browser back/forward cache ](https://web.dev/bfcache/). The BroadcastChannel is the last culprit...

TODO:
- Enable keepalive option for InPlayer `updateWatchHistory` and `addToFavorites` calls
- Refactor the InPlayer's `updatePersonalShelves` method so that it only makes a single call

budzworthy:
> @dbudzins @AntonLantukh I've refactored the get series logic to prevent async logic while closing the tab/browser.
> 
> The Cleeng `updateCustomer` now uses `keepalive: true`, which successfully saves the progress while closing the tab/browser.
> 
> The `beforeunload` event listener is superseded by the `pagehide` and `visibilitychange` listeners. With this change, we're one step closer to supporting the [browser back/forward cache ](https://web.dev/bfcache/). The BroadcastChannel is the last culprit...
> 
> TODO:
> 
> * Enable keepalive option for InPlayer `updateWatchHistory` and `addToFavorites` calls
> * Refactor the InPlayer's `updatePersonalShelves` method so that it only makes a single call

So this means that the current JWP / inplayer integration will work with the periodic saves, but won't capture the 'close' events until we do those TODO's, right?

ChristiaanScheermeijer:
That's right! But my guess is that never worked before either...

budzworthy (approved):

AntonLantukh (approved):
Thank you!

PR #345, opened 2023-07-18
chore(project): add anchor tags for better seo

olga-jwp:
## Description

[OWA-6](https://jwplayer.atlassian.net/browse/OWA-6)

This PR solves # .

### Steps completed:

- Add <a> tag inside Shelf.tsx

- Remove onCardClick() from Shelflist.tsx

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

[OWA-6]: https://jwplayer.atlassian.net/browse/OWA-6?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Shelf/Shelf.tsx
```diff
@@ -71,22 +71,24 @@ const Shelf = ({
...
-          onClick={isInView ? () => onCardClick(item, playlist.feedid, type) : undefined}
-          onHover={typeof onCardHover === 'function' ? () => onCardHover(item) : undefined}
-          featured={featured}
-          disabled={!isInView}
-          loading={loading}
-          isLocked={isLocked(accessModel, isLoggedIn, hasSubscription, item)}
-          posterAspect={posterAspect}
-          item={item}
-        />
+        <a href={isInView ? url : undefined} className={styles.cardWrapper}>
```
ChristiaanScheermeijer: I think it is better to update the `Card` component because now it results in a button wrapped by an anchor tag. This negatively impacts the accessibility for keyboard/screenreader users. Besides that, the Card component is also used on the playlist screens, which use the `CardGrid` component.

**Current HTML**
```html
<a href="/m/uB8aRnu6/agent-327?r=dGSUzs9o" class="_cardWrapper_1ge8j_66">
  <div class="_card_1lln7_1" tabindex="0" role="button" aria-label="Play Agent 327">
```

**Expected HTML**
```html
<a href="/m/uB8aRnu6/agent-327?r=dGSUzs9o" class="_card_1lln7_1" aria-label="Agent 327">
```

I also see that the `aria-label` contains the prefix "Play " which can be removed.

olga-jwp: Hey Chris, thanks! I've discussed that with Danny, and I am indeed moving the tag inside of the Card component.

src/components/CardGrid/CardGrid.test.tsx
```diff
@@ -9,15 +10,17 @@ describe('<CardGrid>', () => {
...
-      <CardGrid
-        playlist={placeholderData}
-        onCardHover={vi.fn()}
-        onCardClick={vi.fn()}
-        isLoading={false}
-        accessModel={'SVOD'}
-        isLoggedIn={true}
-        hasSubscription={true}
-      />,
+      <BrowserRouter>
```
budzworthy: Why did you add this here? I don't think we need another BrowserRouter (in fact multiples might not play nicely together) and we already have 1 at the top level in App.tsx

budzworthy: Oh this is a test file, my bad. You can use the wrapper functions in testUtils.tsx

src/components/Card/Card.tsx
```diff
@@ -28,6 +29,7 @@ type CardProps = {
   isCurrent?: boolean;
   isLocked?: boolean;
   currentLabel?: string;
+  redirectUrl?: string;
```
budzworthy: Let's just call this `url`. The `redirect` prefix might imply some sort of redirect vs navigation and be a bit confusing

src/components/Card/Card.tsx
```diff
@@ -86,14 +89,15 @@ function Card({
   };
 
   return (
-    <div
+    <Link
+      to={redirectUrl ?? ''}
```
budzworthy: Is there any reason the url can't be a required prop?

src/components/CardGrid/CardGrid.test.tsx
```diff
@@ -9,15 +10,17 @@ describe('<CardGrid>', () => {
   it('renders and matches snapshot', () => {
     const placeholderData = generatePlaylistPlaceholder();
```
budzworthy: Can you try using playlist fixture data instead of this placeholder? I think then we'll have more data filled into the snapshots like title and ID's: https://github.com/jwplayer/ott-web-app/blob/579cd0cde5d8c48933867ef3371e542252bdbbb7/src/utils/image.test.ts#L9

src/components/Card/Card.tsx
olga-jwp: I set it optional because of this `isInView` flag `url={isInView ? url : undefined}` in Shelf.tsx

src/components/Shelf/Shelf.tsx
```diff
@@ -71,22 +71,23 @@ const Shelf = ({
...
           progress={watchHistory ? watchHistory[item.mediaid] : undefined}
-          onClick={isInView ? () => onCardClick(item, playlist.feedid, type) : undefined}
           onHover={typeof onCardHover === 'function' ? () => onCardHover(item) : undefined}
           featured={featured}
           disabled={!isInView}
           loading={loading}
           isLocked={isLocked(accessModel, isLoggedIn, hasSubscription, item)}
           posterAspect={posterAspect}
           item={item}
+          redirectUrl={isInView ? url : undefined}
```
budzworthy: I don't think we still need this isInView condition. iirc it was to prevent accidental clicks of the cards on the edges under the arrows. Hopefully the links work better with the slider? Can you test and verify? We'll get fuller indexing without special logic if we can add urls to all of the cards. Otherwise, we can also add a bot detection check here and set the url if a bot is detected or is in view.

olga-jwp: Unfortunately, the behavior for the slider is the same
https://github.com/jwplayer/ott-web-app/assets/118803926/d8cec1c7-df6e-44fe-a707-3a2270af906c

budzworthy: OK, then I will create a separate ticket for this w/ bot detection. Thank you for checking!

src/components/Card/Card.tsx
```diff
@@ -28,6 +29,7 @@ type CardProps = {
   isCurrent?: boolean;
   isLocked?: boolean;
   currentLabel?: string;
+  url?: string;
```
budzworthy: What if we make the url property required, and add a 'disabled' property that we pass to the Link when we don't want it to be clickable? An empty string `to` value causes weird behavior (it drops the entire path besides what was matched by the current React Router route and it creates entries in the nav history.)

I also don't like having optional inputs when we expect a value, because it makes it easy for developers to forget to set properties where the component is used.

src/components/Card/Card.tsx
budzworthy: I did some investigating into what it means to set a link to a blank string, and I don't think it is going to do what you expect it to: https://stackoverflow.com/a/76738446/1345424

I'd suggest you set make url required, always set it, and just add a disabled prop when isInView is false.

src/components/Card/Card.test.tsx
```diff
@@ -11,23 +11,23 @@ const itemWithImage = { title: 'This is a movie', duration: 120, cardImage: 'htt
 
 describe('<Card>', () => {
   it('renders card with video title', () => {
-    const { getByText } = renderWithRouter(<Card item={item} onClick={() => ''} />);
+    const { getByText } = renderWithRouter(<Card item={item} onClick={() => ''} url="" />);
```
budzworthy: Let's put some dummy url in so we see it on the snapshots
```suggestion
    const { getByText } = renderWithRouter(<Card item={item} onClick={() => ''} url="https://test.dummy.jwplayer.com" />);
```

src/components/Card/Card.tsx
```diff
@@ -90,7 +90,7 @@ function Card({
 
   return (
     <Link
-      to={url ?? ''}
+      to={disabled ? '/' : url}
```
budzworthy: I think this will only work on the home page where we want to route to the root (`/`).  You should be able to set the `to` prop to the url unconditionally, but then set disabled on the Link itself.
```suggestion
      to={url}
      disabled={disabled}
```

src/pages/ScreenRouting/playlistScreens/PlaylistGrid/PlaylistGrid.tsx
```diff
@@ -48,14 +43,7 @@ const PlaylistGrid: ScreenComponent<Playlist> = ({ data, isLoading }) => {
...
       <main className={styles.main}>
-        <CardGrid
-          playlist={filteredPlaylist}
-          onCardClick={onCardClick}
-          accessModel={accessModel}
-          isLoggedIn={!!user}
-          hasSubscription={!!subscription}
-          isLoading={isLoading}
-        />
+        <CardGrid playlist={filteredPlaylist} accessModel={accessModel} isLoggedIn={!!user} hasSubscription={!!subscription} isLoading={isLoading} />
```
budzworthy: I usually suggest the layout with 1 property per line unless there are only 1-2 props.

src/pages/Search/Search.tsx
```diff
@@ -44,13 +41,11 @@ const Search = () => {
     }
   }, [firstRender, query, searchQuery, updateSearchQuery]);
 
-  const onCardClick = (playlistItem: PlaylistItem) => {
+  const onCardClick = () => {
     useUIStore.setState({
```
budzworthy: Maybe it's cleaner to just do this in on-unmount now that we aren't navigating via card clicks.
https://legacy.reactjs.org/docs/hooks-effect.html#effects-with-cleanup
```suggestion
  useEffect(() => {
    return () => {
      useUIStore.setState({
        searchQuery: '',
        searchActive: false,
      });
    }
  }, []);
```

src/pages/ScreenRouting/playlistScreens/PlaylistGrid/PlaylistGrid.tsx
olga-jwp: Looks like it's corrected by husky

src/components/Card/Card.test.tsx
```diff
@@ -11,23 +11,23 @@ const itemWithImage = { title: 'This is a movie', duration: 120, cardImage: 'htt
 
 describe('<Card>', () => {
   it('renders card with video title', () => {
-    const { getByText } = renderWithRouter(<Card item={item} onClick={() => ''} />);
+    const { getByText } = renderWithRouter(<Card item={item} onClick={() => ''} url="https://test.dummy.jwplayer.com" />);
```
budzworthy: Whoops, there aren't any snapshot tests to actually check the url part of this. Can you add one for the Card component?

src/components/Card/Card.tsx
```diff
@@ -90,9 +90,9 @@ function Card({
 
   return (
     <Link
-      to={url ?? ''}
+      to={url}
       className={cardClassName}
-      onClick={onClick}
+      onClick={disabled ? (e) => e.preventDefault() : onClick}
```
budzworthy: Is there anywhere still using the onClick prop or can we get rid of it?

olga-jwp: https://github.com/jwplayer/ott-web-app/blob/develop/src/components/VideoLayout/VideoLayout.tsx#L126C36-L126C36
That's the only place left

budzworthy: OK, and that's the one we need to dig into series logic for, right? Can you just add a TODO to remove the onClick prop once we don't need it anymore?

olga-jwp: Not really, it's for displaying the related videos. I'll check if this can be removed

src/components/Card/Card.tsx
```diff
@@ -90,13 +88,12 @@ function Card({
...
-      to={url ?? ''}
+      to={url}
       className={cardClassName}
-      onClick={onClick}
+      onClick={disabled ? (e) => e.preventDefault() : undefined}
       onMouseEnter={onHover}
       tabIndex={disabled ? -1 : 0}
-      onKeyDown={(event: KeyboardEvent) => (event.key === 'Enter' || event.key === ' ') && !disabled && onClick && onClick()}
-      role="button"
+      onKeyDown={(event: KeyboardEvent) => (event.key === 'Enter' || event.key === ' ') && !disabled}
```
budzworthy: This method isn't doing anything. I think you can remove it. The disabled links won't be tabbable anyhow.

src/components/CardGrid/CardGrid.test.tsx
```diff
@@ -13,11 +13,11 @@ describe('<CardGrid>', () => {
       <CardGrid
         playlist={playlist}
         onCardHover={vi.fn()}
-        onCardClick={vi.fn()}
         isLoading={false}
         accessModel={'SVOD'}
         isLoggedIn={true}
         hasSubscription={true}
+        getUrl={() => '/https://test.dummy.jwplayer.com'}
```
budzworthy: You can increase the test coverage a little bit by having this function take a playlist item and set the url with a param or something based on it.
```getUrl={(item: PlaylistItem) => `https://test.dummy.jwplayer.com?media_id=${item.id}`}```

File: src/pages/LegacySeries/LegacySeries.tsx [hunk 16]
```diff
@@ -79,8 +79,10 @@ const LegacySeries = () => {
 
   // Handlers
   const goBack = () => episode && navigate(legacySeriesURL({ episodeId: episode.mediaid, seriesId, play: false, playlistId: feedId }));
-  const onCardClick = (toEpisode: PlaylistItem) =>
-    seriesPlaylist && navigate(legacySeriesURL({ episodeId: toEpisode.mediaid, seriesId, play: false, playlistId: feedId }));
+  const getUrl = (toEpisode: PlaylistItem) => {
+    return (seriesPlaylist && legacySeriesURL({ episodeId: toEpisode.mediaid, seriesId, play: false, playlistId: feedId })) || '';
```
budzworthy:
```suggestion
    return seriesPlaylist ? legacySeriesURL({ episodeId: toEpisode.mediaid, seriesId, play: false, playlistId: feedId })) : '';
```

src/pages/ScreenRouting/mediaScreens/MediaSeries/MediaSeries.tsx
```diff
@@ -182,6 +182,10 @@ const MediaSeries: ScreenComponent<PlaylistItem> = ({ data: seriesMedia }) => {
 
   if (!seriesMedia || !series) return <ErrorPage title={t('series_error')} />;
 
+  const getURL = (toEpisode: PlaylistItem) => {
+    return mediaURL({ media: seriesMedia, episodeId: toEpisode.mediaid });
```
budzworthy: This one is going to be a little different, but we still want to replace the `onCardClick`method.
```suggestion
    return addQueryParam(location, 'e', toEpisode.mediaid);
```

`location` should come from a useLocation hook, which you can put up by the `useSearchParams` hook.

src/pages/ScreenRouting/playlistScreens/PlaylistLiveChannels/PlaylistLiveChannels.tsx
```diff
@@ -201,6 +201,7 @@ const PlaylistLiveChannels: ScreenComponent<Playlist> = ({ data: { feedid, playl
         shareButton={shareButton}
         trailerButton={null}
         favoriteButton={null}
+        getURL={() => ''} // needs to be added with <Link> to <Epg>
```
budzworthy:
```suggestion
        getURL={() => ''} // TODO: OWA-31 - Add link for EPG item
```

src/components/CardGrid/CardGrid.test.tsx
```diff
@@ -13,11 +13,11 @@ describe('<CardGrid>', () => {
       <CardGrid
         playlist={playlist}
         onCardHover={vi.fn()}
-        onCardClick={vi.fn()}
         isLoading={false}
         accessModel={'SVOD'}
         isLoggedIn={true}
         hasSubscription={true}
+        getUrl={(item: PlaylistItem) => `https://test.dummy.jwplayer.com?media_id=${item.id}`}
```
budzworthy: Sorry, this should have been `mediaid`, not just `id`. Did you notice all the undefined values in the snapshot?
```suggestion
        getUrl={(item: PlaylistItem) => `https://test.dummy.jwplayer.com?media_id=${item.mediaid}`}
```

budzworthy (approved):

PR #365, opened 2023-08-21
Webapp bugs

borkopetrovicc:
## Description

https://www.notion.so/jwplayer/Webapp-bugs-51c77f5be0b34cdc9789b88bf597f0a6

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/src/containers/AccountModal/forms/EditPassword.tsx
+++ b/src/containers/AccountModal/forms/EditPassword.tsx
@@ -41,7 +41,7 @@ const ResetPassword: React.FC = () => {
         }
         await changePasswordWithToken(emailParam || '', password, resetToken, passwordConfirmation);
       }
-      await logout(true);
+      await logout();
       navigate(addQueryParam(location, 'u', 'login'));
     } catch (error: unknown) {
       if (error instanceof Error) {
--- a/src/containers/AccountModal/forms/EditPassword.tsx
+++ b/src/containers/AccountModal/forms/EditPassword.tsx
@@ -41,7 +41,7 @@ const ResetPassword: React.FC = () => {
         }
         await changePasswordWithToken(emailParam || '', password, resetToken, passwordConfirmation);
       }
-      await logout();
+      await logout({ includeNetworkRequest: false });
       navigate(addQueryParam(location, 'u', 'login'));
     } catch (error: unknown) {
       if (error instanceof Error) {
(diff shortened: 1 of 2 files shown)


src/containers/AccountModal/forms/EditPassword.tsx
```diff
@@ -40,8 +41,8 @@ const ResetPassword: React.FC = () => {
         }
         await changePasswordWithToken(emailParam || '', password, resetToken, passwordConfirmation);
       }
-      await logout();
-      navigate(addQueryParams(window.location.origin, { u: 'login' }));
+      await logout(true);
```
budzworthy: Why do we not actually log the user out here? Have you tested this change with the Cleeng integration?

src/stores/AccountController.ts
```diff
@@ -146,7 +146,7 @@ export const login = async (email: string, password: string) => {
     useAccountStore.setState({ loading: false });
   });
 };
-export async function logout() {
+export async function logout(onlyState = false) {
```
budzworthy: I would suggest splitting out the `clearLoginState` logic to its own method if it turns out we need it. An optional flag is kind of hard to intuit, especially when the logic is inverted where `logout(false)` means to actually logout and `logout(true)` means not to.

src/containers/AccountModal/forms/EditPassword.tsx
borkopetrovicc: 075a5d6bde7bc10485a74af46852c40e0e35bf9b

src/stores/AccountController.ts
borkopetrovicc: 075a5d6bde7bc10485a74af46852c40e0e35bf9b

kmitrovv (approved):

src/stores/AccountController.ts
```diff
@@ -146,26 +146,36 @@ export const login = async (email: string, password: string) => {
...
+}
+
+type LogoutOptions = {
+  includeNetworkRequest: boolean;
+};
 
-    await restoreFavorites();
-    await restoreWatchHistory();
-    await accountService?.logout();
+export async function logout(logoutOptions: LogoutOptions = { includeNetworkRequest: true }) {
```
kmitrovv: You can delete the type defined above not really needed for just on param
```suggestion
export async function logout(logoutOptions: { includeNetworkRequest: boolean }) {
```

kiremitrov123, line 169:
```suggestion
```

PR #368, opened 2023-09-08
Dbudzins/inversify classes

budzworthy:
## Description

This refactor implements the classes as identifiers for the ioc container

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/stores/ProfileController.ts
```diff
@@ -1,38 +1,42 @@
...
+    return this.profileService?.enterProfile({ id, pin }, this.getSandbox());
+  };
+
+  deleteProfile = ({ id }: ProfileDetailsPayload) => {
+    return this.profileService?.deleteProfile({ id }, this.getSandbox());
+  };
+
+  getProfileDetails = ({ id }: ProfileDetailsPayload) => {
+    return this.profileService?.getProfileDetails({ id }, this.getSandbox());
+  };
```
ChristiaanScheermeijer, line 37: Won't this result in the return type being a union `undefined | Promise<Result>`?

src/stores/ProfileController.ts
```diff
@@ -1,38 +1,42 @@
...
+import { ProfileService } from '#src/services/profile.service';
+import { useConfigStore } from '#src/stores/ConfigStore';
+
+@injectable()
+export default class ProfileController {
+  private readonly profileService: ProfileService;
+
+  constructor(profileService: ProfileService) {
+    this.profileService = profileService;
+  }
```
ChristiaanScheermeijer, line 13: I guess the ProfileService isn't a concrete injection? 

https://github.com/inversify/InversifyJS/blob/master/wiki/constructor_injection.md

src/stores/ProfileController.ts
budzworthy: Yes, that's very much intentional. Downstream code needs to understand that there might not be a profile service / profile controller / profile data. This is true elsewhere for things like checkout too.

src/stores/ProfileController.ts
budzworthy: Is there a bug here? Maybe it needs to be optional (`?`)?It's possible, as the profile service was added after the fact and only came in when I merged latest from dev. I tried to refactor it quickly, but I might have missed something.

src/stores/ProfileController.ts
ChristiaanScheermeijer: Ah I understand. So this can replace the feature flags 😄

src/stores/ProfileController.ts
ChristiaanScheermeijer: No, I think it is correct like this. If the service was concrete, I would prefer using the concrete injection method so we can remove all optional chaining. But your other comment also answered this question 😄

PR #370, opened 2023-09-15
feat(project): per media ads

AntonLantukh:
## Description

- add per media ads functionality

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/components/Player/Player.tsx
```diff
@@ -59,7 +60,8 @@ const Player: React.FC<Props> = ({
   const startTimeRef = useRef(startTime);
   const setPlayer = useOttAnalytics(item, feedId);
 
-  const { adScheduleData } = useConfigStore((s) => s);
+  const { adSchedule, adScheduleUrls } = useConfigStore((s) => s.config);
```
budzworthy: Any reason not to put useConfigStore directly into the useAds hook?

test-e2e/tests/live_channel_test.ts
```diff
@@ -78,7 +78,8 @@ Scenario('I can navigate to live channels from the header', ({ I }) => {
   I.see('On Channel 1', locate('div').inside(videoDetailsLocator));
 });
 
-Scenario('I can watch the current live program on the live channel screen', async ({ I }) => {
+// eslint-disable-next-line codeceptjs/no-skipped-tests
```
budzworthy, line 82: Would be good to at least put a "TODO" here to remind us to re-enable this. I've also done stuff with date checks to temporarily pass/skip the test in code until X date, then start testing it again (that way we'll see it start failing if the stream doesn't get fixed)

src/hooks/useAds.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const useAds = (adScheduleId: string | null | undefined, urls: AdScheduleUrls | undefined, mediaId: string) => {
+  const hasAdConfig = !!urls?.json;
+
+  // Fetch ad-schedule only when ad-config is not set (`adScheduleUrls.json` prop is not set) + adScheduleId is present
+  const { data: adSchedule, isLoading: isAdScheduleLoading } = useAdSchedule(adScheduleId, !!(adScheduleId && hasAdConfig));
+  // adScheduleUrls.json prop exists when ad-config is attached to the App Config
+  const { isLoading: isMediaAdsLoading, data: mediaAds } = useMediaAds(urls?.json, mediaId, hasAdConfig);
+
+  return {
```
budzworthy, line 55: These ORs are kind of confusing. I would probably not worry about splitting the 2 ad lookups into separate methods and just use 1 useQuery with if/else logic inside of it.

AntonLantukh: My idea was to provide an easy way of using just ad-schedule or ad-config if partners need id.

AntonLantukh: + we also need a key to cache the results. For ad-config it should be a media item, but for ad-schedule we don't need it.

budzworthy: Oh I see. That's a nice consideration for our sample repo. Is there a way you can implement a boolean switch to make sure only 1 query is active at a time and that the results from the intended active query are returned based on that switch as well?  My only reservation with the current approach is worrying that somehow both queries will get triggered and return some weird combined or unexpected results

test-e2e/tests/live_channel_test.ts
AntonLantukh: I have created a ticket in the backlog and added TODO to the skipped test. I will ask the team to create a stream when we get OTT params feature in `prd`.

src/hooks/useAds.ts
AntonLantukh: I added a ternary operator to differentiate between two flows. In case we have a url in the config, then we consider the app to use per media ads. If it is not present - then we use old flow with an ad-schedule.

src/components/Player/Player.tsx
```diff
@@ -161,7 +161,13 @@ const Player: React.FC<Props> = ({
 
       // Player options are untyped
       const playerOptions: { [key: string]: unknown } = {
-        advertising: adScheduleData,
+        advertising: {
```
budzworthy, line 164: Would it work if we change these settings in the default player instead of hardcoding showCountdown and showAdMarkers? Then customers will still have the flexibility to control them from the dashboard if they want.

src/hooks/useAds.ts
```diff
@@ -0,0 +1,56 @@
...
+  );
+
+  return {
+    isLoading,
+    data,
+  };
+};
+
+export const useAds = ({ mediaId }: { mediaId: string }) => {
+  const { adSchedule: adScheduleId, adScheduleUrls } = useConfigStore((s) => s.config);
```
budzworthy, line 47: This confused me, because I forgot that we called the app config delivery property "AdScheduleUrls" and then if we have adScheudleUrls we disable the hook that is named "useAdSchedule". I think we can clear it up with just some renamings.

Maybe we should consider calling the perMedia use something more generic like "appBasedAds". That will allow it to make sense for customers who don't use per media, but do want to use our simplified ads instated of a stand alone ad schedule and also set us up to add more function to that system like ad frequency based on subscriptions for example without calling it all "perMediaAds".  Maybe the other hook can be something like useStandaloneAdSchedule?

src/hooks/useAds.ts
budzworthy: Much better! Thanks!

src/hooks/useAds.ts
AntonLantukh: I agree. The naming is not perfect. Should we also deprecate useAdSchedule / useStandaloneAdSchedule hook? We hide this setting on the App Config page when Ad Config is active.

budzworthy: We can call it `useLegacyStandaloneAds` or something. Unfortunately I think we'll need to leave it in for a while because otherwise we might break customers who have deployments with it.

src/components/Player/Player.tsx
budzworthy: Never mind, those settings aren't in the dashboard

src/hooks/useAds.ts
AntonLantukh: Fixed it

budzworthy (approved):
Looks good!

budzworthy (approved):

PR #353, opened 2023-07-31
Feat/User Profiles

naumovski-filip:
## Description

Testing config: **maqet3qh**

### TODO:
- ~~Move Profiles logic from account service and controller to Profiles service and controller.~~
- ~~Move edit, delete and create profile handlers to hook~~
- Unit and E2E tests

### Steps completed:

According to our definition of done, I have completed the following steps:

- [X] Acceptance criteria met
- [ ] Unit tests added
- [ ] Docs updated (including config and env variables)
- [X] Translations added
- [X] UX tested
- [X] Browsers / platforms tested
- [X] Rebased & ready to merge without conflicts
- [X] Reviewed own code

PR #371, opened 2023-09-18
Fix: Additional Profiles fixes

naumovski-filip:
## Description

Fixes leftover comments from https://github.com/jwplayer/ott-web-app/pull/353.

Testing config: **eehypvlc**

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


borkopetrovicc (approved):

src/pages/User/User.tsx
```diff
@@ -85,7 +86,7 @@ const User = (): JSX.Element => {
                     to={`my-profile/${profile?.id}`}
                     label={profile?.name ?? t('nav.profile')}
                     variant="text"
-                    startIcon={<img className={styles.profileIcon} src={profile?.avatar_url} alt={profile?.name} />}
+                    startIcon={<img className={styles.profileIcon} src={profile?.avatar_url || defaultAvatar} alt={profile?.name} />}
```
budzworthy: It looks like similar fallback logic is used in several places. Would it make sense to have the profile returned from the store hooks above be a dummy one with the default image by default?

src/hooks/useProfiles.ts
budzworthy, line 1: The code in here feels like it should be part of a controller more so than just a hook

src/containers/Layout/Layout.tsx
```diff
@@ -154,6 +156,8 @@ const Layout = () => {
           currentProfile={profile ?? undefined}
```
budzworthy: There are a lot of profile properties added here. Can we combine them in a single object to make passing around cleaner?

src/containers/Profiles/Profiles.tsx
```diff
@@ -40,42 +46,68 @@ const Profiles = ({ editMode = false }: Props) => {
...
-            <Button onClick={() => navigate('/u/profiles/edit')} label={t('account.manage_profiles')} variant="outlined" size="large" fullWidth={isMobile} />
-          ) : (
-            <Button onClick={() => navigate('/u/profiles')} label="Done" variant="outlined" size="large" fullWidth={isMobile} />
-          )}
-        </div>
-      )}
-    </div>
+      <Alert
+        isSuccess
+        open={creationSuccess}
```
budzworthy, line 93: Is there any way to do this with state instead of URL? I don't think there's a security risk here or anything, but it would be sort of weird if anyone bookmarked this url with the params in it and saw the "Created" message each time.

naumovski-filip: I'm not sure what the best way to approach it using state would be. We'd have to either expand on one of the stores or add context which might end up being bulky. Is there some other way I'm not seeing?

src/containers/Layout/Layout.tsx
naumovski-filip: https://github.com/jwplayer/ott-web-app/pull/371/commits/df8266634ac7eaba18352129b93fdf5d38761a25

src/pages/User/User.tsx
naumovski-filip: https://github.com/jwplayer/ott-web-app/commit/df8266634ac7eaba18352129b93fdf5d38761a25#diff-6c741f0c03d4c3fce6df09209928c5456ed6c41be862f1042159c4ecca8b5efc

Wdyt of this approach?

src/containers/Profiles/Profiles.tsx
budzworthy: I was thinking just a local useState hook

naumovski-filip: They are separate components on separate routes though

src/containers/Profiles/Form.tsx
```diff
@@ -41,7 +41,12 @@ const Form = ({ initialValues, formHandler, selectedAvatar, showCancelButton = t
   ];
 
   const validationSchema: SchemaOf<{ name: string }> = object().shape({
-    name: string().required(t('profile.validation.name.required')).min(2, t('profile.validation.name.too_short')),
+    name: string()
```
kmitrovv, line 44: recheck the validation.. seems like it has missing translation, as well as misleading validation message. 

image
image

kmitrovv (commented):
When you try to create a profile with an existing profile name. We should add the correct validation error here.
image

src/hooks/useProfiles.ts
```diff
@@ -105,5 +83,20 @@ export const useProfiles = (
   if (!canManageProfiles && query.data?.responseData.canManageProfiles) {
     useAccountStore.setState({ canManageProfiles: true });
   }
-  return { ...query, profilesEnabled: query.data?.responseData.canManageProfiles && canManageProfiles };
+
+  return {
```
kmitrovv, line 98: If this is like this just for the avatar, maybe it can be split somehow..I don't see the issue now, it's just a little hard to read, but in the future if we need to handle more properties here, it may be problematic.

src/stores/ProfileController.ts
```diff
@@ -21,7 +33,20 @@ export const updateProfile = async ({ id, name, adult, avatar_url, pin }: Profil
 
 export const enterProfile = async ({ id, pin }: EnterProfilePayload) => {
   return await useService(async ({ profileService, sandbox }) => {
-    return await profileService?.enterProfile({ id, pin }, sandbox ?? true);
+    const response = await profileService?.enterProfile({ id, pin }, sandbox ?? true);
+    const profile = response?.responseData;
+    if (profile?.credentials?.access_token) {
```
kmitrovv: This can be handled on service level, and that way you will not need the condition as well.
You can use `InPlayer.Account.setToken`, something similar like in initialize account service.
Or maybe somehow combine it with initializeAccount since it's called right after this is done.

src/hooks/useProfiles.ts
ChristiaanScheermeijer: The same can be achieved a bit more clean. It's not a good idea to spread the `query` object because query-query will track all observed properties and trigger a rerender when these tracked properties update. I don't know if this is enabled by default in this version, but it is in the latest.

I would also use the `onSuccess` and `select` properties to set the `canManageProfiles` state once (without checking this every render) and restructure the data.

PS Can we prevent changing the store from this query? It is a weird side-effect, IMHO...

```tsx
  return useQuery(['listProfiles'], listProfiles, { 
    ...options, 
    enabled: !!user,
    onSuccess: (data) => {
      // update the canManageProfiles in state based on this response
      if (!useAccountStore.getState().canManageProfiles && data.responseData.canManageProfiles) {
        useAccountStore.setState({ canManageProfiles: true });
      }
    },
    select: (data) => ({
      ...data,
      responseData: {
        ...data.responseData,
        collection: data.responseData.collection.map((profile) => ({
          ...profile,
          avatar_url: profile?.avatar_url || defaultAvatar,
        })),
      },
    }),
  });
```

Even better would be to move the default avatar_url to the service call because then it will be cached by react-query, and we don't need to create a new object every render.

src/components/Header/Header.tsx
```diff
@@ -49,10 +50,15 @@ type Props = {
...
-  currentProfile?: Profile;
-  profiles?: Profile[];
-  profilesEnabled?: boolean;
+
   accessModel?: AccessModel;
+  profilesData?: {
+    currentProfile?: Profile;
+    profiles?: Profile[];
+    profilesEnabled?: boolean;
+    selectProfile?: UseMutateFunction<unknown, unknown, { id: string; avatarUrl: string }, unknown>;
```
ChristiaanScheermeijer: Both minor feedback:

I would try to prevent passing a full mutate function to a component. Generally, a component shouldn't care about what or how things are happening. It should get all the information via simple props.

This makes testing and mocking easier.

E.g.

```tsx
profiles?: {
  enabled?: boolean;
  currentProfile?: Profile;
  profiles?: Profile[];
  onProfileSelect?: (profile: { id: string; avatarUrl: string; }) => void;
  isSelectingProfile?: boolean;
}
```

Should all properties of the `profiles` prop be optional, or can we assume that all properties are given when the `profiles` prop is set?

E.g.

```tsx
profiles?: {
  currentProfile: Profile;
  profiles: Profile[];
  onProfileSelect: (profile: { id: string; avatarUrl: string; }) => void;
  isSelectingProfile: boolean;
}

// ...
  profiles,
}) => {
  const { t } = useTranslation('menu');
  const profilesEnabled = !!profiles;
  const { currentProfile, profiles, onProfileSelect, isSelectingProfile } = profiles || {};
```

src/components/Header/Header.tsx
```diff
@@ -133,7 +137,11 @@ const Header: React.FC<Props> = ({
     return isLoggedIn ? (
       <React.Fragment>
         <IconButton className={classNames(styles.iconButton, styles.actionButton)} aria-label={t('open_user_menu')} onClick={openUserMenu}>
-          {currentProfile?.avatar_url ? <ProfileCircle src={currentProfile.avatar_url} alt={currentProfile.name} /> : <AccountCircle />}
+          {profilesEnabled && currentProfile ? (
+            <ProfileCircle src={currentProfile.avatar_url} alt={currentProfile?.name ?? t('profile_icon')} />
```
ChristiaanScheermeijer: 1. You don't need the optional chaining here
2. In this case, you don't want to use the nullish coalescing operator to prevent an empty alt attribute

```suggestion
            <ProfileCircle src={currentProfile.avatar_url} alt={currentProfile.name || t('profile_icon')} />
```

src/components/UserMenu/UserMenu.tsx
```diff
@@ -26,14 +25,24 @@ type Props = {
   currentProfile?: Profile;
   profilesEnabled?: boolean;
   profiles?: Profile[];
+  isSelectingProfile?: boolean;
+  selectProfile?: UseMutateFunction<unknown, unknown, { id: string; avatarUrl: string }, unknown>;
```
ChristiaanScheermeijer:
```suggestion
  onProfileClick?: (selectedProfile: { id: string; avatarUrl: string }) => void;
```

src/hooks/useProfiles.ts
```diff
@@ -24,27 +17,14 @@ export const useSelectProfile = () => {
...
-      }
+    onSuccess: async () => {
+      useProfileStore.setState({ selectingProfileAvatar: null });
+      navigate('/');
     },
     onError: () => {
       useProfileStore.setState({ selectingProfileAvatar: null });
-      throw new Error('Unable to enter profile.');
+      navigate('/u/profiles');
+      console.error('Unable to enter profile.');
```
ChristiaanScheermeijer: This is only useful for dev?

```suggestion
      logDev('Unable to enter profile.');
```

src/hooks/useProfiles.ts
```diff
@@ -96,14 +73,37 @@ export const useDeleteProfile = (options?: UseMutationOptions<ServiceResponse<Co
...
+    ) => void,
+  ) => {
+    const formError = e as ProfileFormSubmitError;
+    if (formError.message.includes('409')) {
+      setErrors({ name: t('profile.validation.name.already_exists') });
+      return;
+    }
+    setErrors({ form: t('profile.form_error') });
+  };
+};
```
ChristiaanScheermeijer: You want to ensure that the `unknown` error is of type `ProfileFormSubmitError`. If you have control, even better would be to throw a custom error so that you can use `instanceof` checks instead.

There is already a ProfileFormValues type, why not use that in the `setErrors` argument?

```suggestion
export const isProfileFormSubmitError = (error: unknown): error is ProfileFormSubmitError => {
  return !!error && typeof error === 'object' && 'message' in error;
} 

export const useProfileErrorHandler = () => {
  const { t } = useTranslation('user');

  return (error: unknown, setErrors: (errors: Partial<ProfileFormValues & GenericFormErrors>) => void) => {
    if (isProfileFormSubmitError(error) && error.message.includes('409')) {
      setErrors({ name: t('profile.validation.name.already_exists') });
      return;
    }
    setErrors({ form: t('profile.form_error') });
  };
};
```

src/hooks/useProfiles.ts
ChristiaanScheermeijer: The `query` object is still spread in the `return` statement, is it possible to prevent that?

src/stores/AccountController.ts
```diff
@@ -13,22 +25,11 @@ import { restoreWatchHistory, serializeWatchHistory } from '#src/stores/WatchHis
...
-  GetCustomerConsentsResponse,
-  GetPublisherConsentsResponse,
-} from '#types/account';
 import type { Offer } from '#types/checkout';
-import { unpersistProfile } from '#src/hooks/useProfiles';
 
 const PERSIST_PROFILE = 'profile';
 
-export const initializeAccount = async () => {
+export const initializeAccount = async ({ profile }: { profile?: Profile } = {}) => {
```
ChristiaanScheermeijer: The initialize account function intends to initialize all account-related stores and services only once. A service could bind event listeners or initialize actions that shouldn't be performed multiple times.

It is entangled with profile logic and called when switching the profile.

Would something like this work?

```mermaid
---
title: Account/Profile service flow
---
flowchart TD
    START[/Startup/] --> CONFIG(Load config)
    CONFIG --> INITP[Profile.initialize]
    INITP --> INITA[Account.initialize]
    INITA --> LOADU[Account.loadUser]
    LOADU --> EMPTY[[Empty favorites/history]]
    EMPTY --> AUTH[[getAuthData]]
    AUTH --> END[[account/watchHistory/favorites]]
    SELECT[/Select profile/] --> SWITCH[Profile.switchProfile]
    SWITCH --> PERSIST[Profile.persistProfile]
    PERSIST --> LOADU
```

ChristiaanScheermeijer (commented):
Nice fixes @naumovski-filip! I've left some additional feedback 😄 Not everything is super important, but I think the profile initialization can use some refactoring.

src/stores/AccountController.ts
naumovski-filip: You're right, there will be constant unnecessary calls to initializeAccount. I've separated the logic a bit here: https://github.com/jwplayer/ott-web-app/pull/371/commits/d358ddbf0613ff5e0d0bb8e483770219761cec8d

Not exactly like your example but I think it should achieve the same result.

naumovski-filip: I've also tried to address the rest of the comments in the same commit

src/components/UserMenu/UserMenu.tsx
```diff
@@ -45,14 +53,13 @@ const UserMenu = ({ showPaymentsItem, small = false, onClick, accessModel, curre
 
   return (
     <ul className={styles.menuItems}>
-      {accessModel === 'SVOD' && profilesEnabled && (
+      {accessModel === 'SVOD' && profilesEnabled && selectProfile && (
```
darkoatanasovski: what if the `accessModel` is **AVOD** or any other? Should we show the user menu in that case?

src/containers/Profiles/avatarUrls.json
darkoatanasovski, line 1: should we move this file to `./src/assets`?

src/stores/ProfileController.ts
```diff
@@ -1,9 +1,71 @@
...
+  persist.setItemStorage('inplayer_token', {
+    expires: profile.credentials.expires,
+    token: profile.credentials.access_token,
+    refreshToken: '',
+  });
+};
+
+export const isValidProfile = (profile: unknown): profile is ProfilesData => {
+  return (
+    typeof profile === 'object' &&
```
darkoatanasovski: can we use `yup` validator here?

kmitrovv (approved):

borkopetrovicc (approved):

PR #376, opened 2023-09-22
Feat: Custom Registration Field (rollback of the rollback + bugfixes and e2e coverage)

mirovladimitrovski:
## Description

Everything introduced in [this PR](https://github.com/jwplayer/ott-web-app/pull/310) is included here, plus various fixes to bugs discovered in the meantime.

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


src/services/inplayer.account.service.ts
```diff
@@ -86,8 +86,33 @@ export const login: Login = async ({ config, email, password }) => {
...
-export const register: Register = async ({ config, email, password }) => {
+export const register: Register = async ({ config, email, password, customFields }) => {
   try {
+    const metadata = {
+      ...customFields,
+      ...Object.entries(customFields).reduce((acc, [name, val]) => {
+        const value = (() => {
+          if (name === 'us_state') {
+            if (customFields.country === 'us') {
+              return val === 'n/a' ? '' : val;
```
budzworthy: There must be a better way to deal with this (Filtering out 'n/a' values) at the component level. This is implementation detail that should stay in tsx and not get all the way down into services.

mirovladimitrovski: Unfortunately it's very specific to Inplayer and that's why I thought it belonged there. Unless I extract the logic in a formatter function, but it'll still be in the same file, as I believe there are already few such functions in it already.

mirovladimitrovski: Done. Disregard my previous comment.

src/components/CustomRegisterField/CustomRegisterField.tsx
```diff
@@ -62,7 +62,7 @@ export const CustomRegisterField: FC<Props> = ({ type, value = '', ...props }) =
     case 'datepicker':
       return <DateField {...props} value={value as string} />;
     default:
-      return <Checkbox {...props} checked={value === true} />;
+      return <Checkbox {...props} checked={[true, 'on'].includes(value)} />;
```
budzworthy:
```suggestion
      return <Checkbox {...props} checked={isTruthyCustomParam(value)} />;
```

budzworthy: We already have a helper that does a good job of this. Probably it's no big deal if there are some extra truthy values, but if so, then we can create a copy of the method for INP/JWP params.

src/utils/collection.ts
```diff
@@ -142,6 +142,30 @@ const checkConsentsFromValues = (publisherConsents: Consent[], consents: Record<
   return { customerConsents, consentsErrors };
 };
 
+const isNotEmptyStringEntry = <T>([, value]: [string, T]) => value !== '';
+
+const formatCrfEntry = <T>([name, value]: [string, T], _: number, collection: [string, T][]) => {
```
budzworthy: Please avoid obscure abbreviations like "Crf". Prefer to just call it `formatRegisterFieldEntry`

src/components/Account/Account.tsx
```diff
@@ -262,14 +262,17 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
           formSection({
             label: t('account.terms_and_tracking'),
             saveButton: t('account.update_consents'),
-            onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, { ...values.consentsValues, terms: true })),
+            onSubmit: (values) => {
+              const cleanConsentValues = Object.fromEntries(Object.entries(values.consentsValues).filter(isNotEmptyStringEntry).map(formatCrfEntry));
+              return updateConsents(formatConsentsFromValues(publisherConsents, { ...cleanConsentValues, terms: true }), cleanConsentValues);
```
budzworthy: Why is `cleanConsentValues` being passed in twice to this method in different ways?

src/components/Account/Account.tsx
```diff
@@ -262,14 +262,17 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
           formSection({
             label: t('account.terms_and_tracking'),
             saveButton: t('account.update_consents'),
-            onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, { ...values.consentsValues, terms: true })),
+            onSubmit: (values) => {
+              const cleanConsentValues = Object.fromEntries(Object.entries(values.consentsValues).filter(isNotEmptyStringEntry).map(formatCrfEntry));
```
budzworthy: Looks like this logic is copied at least 1 other place. Good to abstract it into a method.

src/components/Account/Account.tsx
```diff
@@ -282,7 +285,11 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
             formSection({
               label: t('account.other_registration_details'),
               saveButton: t('account.update_consents'),
-              onSubmit: (values) => updateConsents(formatConsentsFromValues(publisherConsents, values.consentsValues)),
+              onSubmit: (values) => {
```
budzworthy: Seems like the whole onSubmit logic is duplicated twice too, with just the `terms` boolean different. Could easily be abstracted into 1 function.

src/containers/AccountModal/forms/PersonalDetails.tsx
```diff
@@ -117,7 +119,10 @@ const PersonalDetails = () => {
             value: questionValues[question.key],
           } as CaptureCustomAnswer),
       );
-      await updateCaptureAnswers(removeEmpty({ ...formData, customAnswers }));
+
+      const metadata = removeEmpty({ ...user?.metadata });
+
+      await updateCaptureAnswers(removeEmpty({ ...formData, customAnswers, metadata }));
```
budzworthy: Looks like there are hidden type problems here. The controller `updateCaptureAnswers` function only expects an arg of type `Capture`, which doesn't have a `metadata` property. Typescript only allows it because the `removeEmpty` function converts it to a Record type, but if you remove it temporarily you'll see the type error. `metadata` also doesn't exist on `UpdateCaptureStatusArgs` on the service, so it's actually only when it gets to the call to `updateCustomer` in the inplayer service that it is using the right type. This is too fragile of an approach. It can easily be broken with code cleanup and it will be quite hard to debug.

src/containers/AccountModal/forms/Registration.tsx
```diff
@@ -57,9 +57,11 @@ const Registration = () => {
         return;
       }
 
-      await register(email, password);
+      const cleanConsentValues = Object.fromEntries(Object.entries(consentValues).filter(isNotEmptyStringEntry).map(formatCrfEntry));
```
budzworthy: Should you be using `customerConsents` here? Or do the JWP fields not behave the same way where `publisherConsents` will be a super set of `consentValues`?

src/services/inplayer.account.service.ts
```diff
@@ -210,10 +211,15 @@ export const getCustomerConsents: GetCustomerConsents = async (payload) => {
 
 export const updateCustomerConsents: UpdateCustomerConsents = async (payload) => {
   try {
-    const { customer, consents } = payload;
+    const { customer, consents, consentValues } = payload;
```
budzworthy: I'm confused as to why you added another field that has the same data (`consents` and `consentValues`.) Shouldn't it have been possible to just mutate the existing `consents` data?

src/containers/AccountModal/forms/Registration.tsx
```diff
@@ -57,9 +57,11 @@ const Registration = () => {
...
 
-      await updateConsents(customerConsents).catch(() => {
+      await register(email, password, cleanConsentValues);
+
+      await updateConsents(customerConsents, cleanConsentValues).catch(() => {
```
budzworthy: Isn't this redundant for the JWP service since consents are set with the `register` call? Since there is only 1 use of register, I think it would make sense to call `updateConsents` within the implementation of `register` for the cleeng service that needs this separately. Then you can maintain a common interface and avoid redundant calls.

src/stores/AccountController.ts
```diff
@@ -228,7 +228,10 @@ export const updatePersonalShelves = async () => {
   });
 };
 
-export const updateConsents = async (customerConsents: CustomerConsent[]): Promise<ServiceResponse<CustomerConsent[]>> => {
+export const updateConsents = async (
+  customerConsents: CustomerConsent[],
+  consentValues: Record<string, string | boolean>,
```
budzworthy: Again, isn't this just the same data as `customerConsents` in a different format?

types/account.d.ts
```diff
@@ -256,6 +260,7 @@ export type UpdateCustomerConsentsArgs = {
   config: Config;
   customer: Customer;
   consents: CustomerConsent[];
+  consentValues: Record<string, string | boolean>;
```
budzworthy: We need to try to make the cleeng and jwp services use 1 format, not just create a second property with similar but slightly different formatted data. Why can't we mutate consentValues into an array of `CustomerConsents`?

budzworthy (dismissed):
Unfortunately I think this PR needs some work before it moves forward. There is too much diverging of cleeng and jwp implementations. We need to mutate the data to use the same formats instead of just adding new properties to hold the jwp version of things.

src/components/Account/Account.tsx
mirovladimitrovski: Because cleeng requires the values to be in a `[{ name, version, state: 'accepted' | 'declined', value }[]` format and inplayer/jwp needs them to be shaped as `{ [name]: value }`. I had the formatting logic isolated inside the `updateConsents` method in `inplayer.account.service.ts` and now I have it extracted in the `isNotEmptyStringEntry` and `formatCrfEntry` functions. I can either do it this way or use those inside `updateConsents` again.

src/containers/AccountModal/forms/Registration.tsx
mirovladimitrovski: Yeah, it's gotten redundant as bugs got discovered and as I apply fixes, but it is the way it's currently implemented and such refactor, no matter how small its scope, would still be out of the scope of this PR. I suggest leaving this for a separate task and move this PR forward if everything else is fine.

src/components/Account/Account.tsx
```diff
@@ -65,15 +67,33 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
...
   );
 
-  const consentValues = useMemo(() => formatConsentValues(publisherConsents, customerConsents), [publisherConsents, customerConsents]);
+  const [termsConsents, nonTermsConsents] = useMemo(() => {
+    const terms: Consent[] = [];
+    const nonTerms: Consent[] = [];
+
+    publisherConsents?.forEach((consent) => {
+      if (consent?.type === 'checkbox') {
+        terms.push(consent);
```
AntonLantukh, line 76: What if we have a custom field with a `checkbox` type?

src/utils/collection.ts
```diff
@@ -60,6 +60,25 @@ const formatConsentValues = (publisherConsents: Consent[] | null = [], customerC
   if (!publisherConsents || !customerConsents) {
     return {};
   }
+
+  const values: Record<string, string | boolean> = {};
+
+  publisherConsents?.forEach((publisherConsent) => {
```
AntonLantukh: Maybe `reduce` function could help here?

src/utils/collection.ts
```diff
@@ -60,6 +60,25 @@ const formatConsentValues = (publisherConsents: Consent[] | null = [], customerC
...
+    const consent = customerConsents?.find((customerConsent) => customerConsent.name === publisherConsent.name);
+
+    if (consent) {
+      const value = publisherConsent.isCustomRegisterField === true ? consent.value ?? '' : consent.state === 'accepted';
```
AntonLantukh:
```suggestion
      const value = publisherConsent.isCustomRegisterField ? consent.value ?? '' : consent.state === 'accepted';
```

src/components/CustomRegisterField/CustomRegisterField.tsx
```diff
@@ -0,0 +1,70 @@
...
+      }
+      default: {
+        if (props.options) {
+          return Object.entries(props.options).map(([value, label]) => ({ value, label }));
+        }
+
+        return [];
+      }
+    }
+  }, [t, type, props.options, i18n]);
```
AntonLantukh:
```suggestion
  const optionsList = useMemo(() => {
    if (type === 'country' || type === 'us_state') {
     return Object.keys(i18n.getResourceBundle(i18n.language, type)).map((code) => ({
          value: code,
          label: t(`${type}:${code}`),
        }));
      }
      
     return Object.entries(options || []).map(([value, label]) => ({ value, label }));
    }
  }, [t, type, options, i18n]);
```

src/components/CustomRegisterField/CustomRegisterField.test.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import { CustomRegisterField } from './CustomRegisterField';
+
+describe('<CustomRegisterField>', () => {
+  test('renders and matches snapshot <Checkbox>', () => {
+    const { container } = render(<CustomRegisterField type="checkbox" label="label" name="name" value="value" onChange={vi.fn()} />);
+
+    expect(container).toMatchSnapshot();
```
AntonLantukh, line 10: Does it make sense for us to check here that expected `testId` is renedered? That would mean that the component is one which we expect to have.

src/utils/collection.ts
AntonLantukh: Could you please also help me to understand why we check `consent.state === 'accepted' here and assign it as a value?

src/components/DateField/DateField.tsx
```diff
@@ -105,22 +133,22 @@ const DateField: React.FC<Props> = ({ className, label, error, helperText, value
...
-    <div className={DateFieldClassName}>
+    <div className={DateFieldClassName} {...rest} data-testid={testId} id={id}>
       <label htmlFor={id} className={styles.label}>
         {label}
-        {!rest.required ? <span>{t('optional')}</span> : null}
+        {!required ? <span>{t('optional')}</span> : null}
       </label>
       <div className={styles.container}>
+        {/* don't be tempted to make it type="hidden", onChange will practically be ignored that way */}
+        <input ref={hiddenInputRef} id={`${id}-hidden`} className={styles.hiddenInput} name={name} onChange={onChange} />
```
AntonLantukh, line 151: Why do we need `hiddenInput` here?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -37,14 +38,15 @@ const Dropdown: React.FC<Props & React.AriaAttributes> = ({
...
   ...rest
 }: Props & React.AriaAttributes) => {
   const { t } = useTranslation('common');
   const id = useOpaqueId();
 
   return (
-    <div className={classNames(styles.container, { [styles.fullWidth]: fullWidth, [styles.error]: error }, styles[size], className)}>
-      {label && (
+    <div className={classNames(styles.container, { [styles.fullWidth]: fullWidth, [styles.error]: error }, styles[size], className)} data-testid={testId}>
+      {(label || !required) && (
```
AntonLantukh, line 49: Does `label` tag make sense without the actual label?

src/components/Radio/Radio.tsx
```diff
@@ -9,30 +9,30 @@ import useOpaqueId from '#src/hooks/useOpaqueId';
...
 };
 
-const Radio: React.FC<Props> = ({ name, onChange, header, value, values, helperText, error, required }: Props) => {
+const Radio: React.FC<Props> = ({ name, onChange, header, value, values, helperText, error, required, ...rest }: Props) => {
   const { t } = useTranslation('common');
   const id = useOpaqueId('radio', name);
 
   return (
-    <div className={error ? styles.error : undefined}>
-      {header ? (
```
AntonLantukh, line 26: Do we need to show header without the content?

test-e2e/utils/steps_file.ts
```diff
@@ -77,6 +80,125 @@ const stepsObj = {
...
+      this.seeElement(constants.customRegFields.topContainerSelector);
+    });
+  },
+  hasTermsAndConditionField: async function (this: CodeceptJS.I) {
+    return tryTo(() => {
+      this.seeElement(constants.customRegFields.termsAndConditionsField);
+    });
+  },
+  fillCustomRegistrationFields: async function (this: CodeceptJS.I) {
+    if (!(await this.hasCustomRegistrationFields())) {
```
AntonLantukh, line 94: Is it possible for us to check account with and without custom fields separately? It looks like in case we remove custom fields from the account used in e2e tests, then we won't even see / catch it because we won't check them in this case.

src/utils/common.ts
```diff
@@ -83,9 +83,9 @@ export function getOverrideIP() {
     .trim();
 }
 
-export const isTruthyCustomParamValue = (value: unknown): boolean => ['true', '1', 'yes'].includes(String(value)?.toLowerCase());
+export const isTruthyCustomParamValue = (value: unknown): boolean => ['true', '1', 'yes', 'on'].includes(String(value)?.toLowerCase());
```
AntonLantukh, line 86: I think we agreed that for custom params we consider only three values to be `true`: 'true', '1', 'yes'. We use them for example in case a customer wants to define free content using `free` custom param.

src/services/inplayer.account.service.ts
```diff
@@ -158,11 +164,35 @@ export const getPublisherConsents: GetPublisherConsents = async (config) => {
...
+      .filter((field) => !['email_confirmation', 'first_name', 'surname'].includes(field.name) && ![terms].includes(field))
+      .map(
+        (field): Consent => ({
+          type: field.type as CustomRegisterFieldVariant,
+          isCustomRegisterField: true,
+          name: field.name,
+          label: field.label,
+          placeholder: field.placeholder,
+          required: field.required,
+          // todo: field.option type in SDK is incorrect, remove the type casting after fixing that
```
AntonLantukh: Could be great to create a ticket to fix it

src/components/Checkbox/Checkbox.tsx
```diff
@@ -20,12 +20,12 @@ type Props = {
...
 };
 
-const Checkbox: React.FC<Props> = ({ label, name, onChange, header, checked, value, helperText, disabled, error, required }: Props) => {
+const Checkbox: React.FC<Props> = ({ label, name, onChange, header, checked, value, helperText, disabled, error, required, ...rest }: Props) => {
   const { t } = useTranslation('common');
   const id = useOpaqueId('check-box', name);
 
   return (
-    <div className={classNames(styles.checkbox, { [styles.error]: error })}>
+    <div className={classNames(styles.checkbox, { [styles.error]: error })} {...rest}>
```
AntonLantukh, line 28: Do we need any extra props for this `div`?

mirovladimitrovski: Yes, `data-test`.

src/services/inplayer.account.service.ts
mirovladimitrovski: It's fixed now.

src/utils/common.ts
mirovladimitrovski: I already had a comment about this by @dbudzins  when I had it with `['on', true].includes(value)` and was explicitly suggested to use the `isTruthyCustomParamValue` function instead and add `on` to it.

test-e2e/utils/steps_file.ts
mirovladimitrovski: If custom fields are removed, that means there are no such fields to fill out. I'm sure e2e tests can be improved with more coverage, but this PR has been stuck in review for such a long time, I suggest we move it forward and leave such improvements for later.

src/components/Radio/Radio.tsx
mirovladimitrovski: Yes, because when the field isn't `required` we want to show `(Optional)` even if `header` happens to be empty.

src/utils/collection.ts
mirovladimitrovski: Because consents have two possible answers: `accepted` and `declined`. While custom register fields have multiple possibilities for an answer: `on` and `off` (which are interchangeable with `true` and `false`) and practically any string value.

src/components/Account/Account.tsx
mirovladimitrovski: Any checkbox gets treated as a "consent" for something. It's the way it was before, so I figured that's the way it's supposed to be.

src/components/CustomRegisterField/CustomRegisterField.tsx
mirovladimitrovski: This is a matter of personal preference.

- Sometimes I like to create a separate var, such as `const codes`, so that its name would be descriptive enough and would make it more apparent to other developers (and to my future self) what I'm doing - in this case I'm trying to make it more obvious why I need `Object.keys(i18n.getResourceBundle(i18n.language, type))` - it's for the (country or US state) codes. So I'd rather have us keep that const.

- On the `if` vs `switch` state - I'm in favour of either `switch` or the `if ([].includes())` pattern in this case.

- On `Object.entries(options || []).map()` vs `if (!options) return [];` - the latter would prevent `.map()` to be run at all if `!options`. A teeny tiny optimisation.

darkoatanasovski (approved):

src/components/Dropdown/Dropdown.tsx
mirovladimitrovski: Same here - it's for showing `(Optional)` if the field is not `required`, even if `label` is empty.

src/components/DateField/DateField.tsx
mirovladimitrovski: To pose for an event-triggering element. See also the `triggerChangeEvent` function.

test-e2e/utils/randomizers.ts
```diff
@@ -0,0 +1,9 @@
+const random = (min = 0, max = 0) => min + Math.round(Math.random() * (max - min));
+
+const randomDay = () => random(1, 28);
+
+const randomMonth = () => random(1, 12);
+
+const randomYear = () => random(2050, 2100);
+
+export const randomDate = () => [randomDay(), randomMonth(), randomYear()];
```
budzworthy, line 9: Won't this sometimes result in invalid dates (i.e. February 30?)

mirovladimitrovski: The `DateField` component does allow that. But this line of code can't produce that kind of date because `const randomDay = () => random(1, 28);`.

PR #377, opened 2023-09-25
perf(project): accounts flow requests

kmitrovv:
## Description

As an outcome from the `envision subscription at scale` design sprint, we've came up with some `easy` and immediate to fix bottlenecks, in order to have some improvements in our system. 

This change will decrease the `accounts` requests at least 70% since it won't fire on every page load, authentication or registration anymore. 

The initial data will be decoded and loaded from the jwt token, and only when a user navigates to the account, the data will be fetched from the endpoint.

Notion card: [Improve requests flow on web app](https://www.notion.so/jwplayer/Improve-requests-flow-on-web-app-a81badbaa4ed4bfa9d8f5f316ac8c4f5)

**Note** This performance updates will be needed for other places as well, such as: payments, subscriptions, profiles, and probably different approach will be used, something like persistQuery from react-query, but since it's a bigger scope and affects pretty much the entire web app, it will need grooming and discussions with the OTT team.


src/components/Account/Account.tsx
```diff
@@ -48,24 +50,39 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
...
+  // update account with fresh data when visited
+  useEffect(() => {
+    const getUserInfo = async () => {
+      if (fetchOnVisit) {
+        const { user } = await accountService.getUser({ config, shouldFetchData: true });
+        useAccountStore.setState({ user, fetchOnVisit: false });
+      }
+    };
+    getUserInfo();
+  }, [customer, fetchOnVisit, accountService, config]);
```
AntonLantukh: Two comments here:

1. Can we use `useQuery` here instead of `useEffect` hook + `fetchOnVisit` flag? It will be triggered only once when cache settings are set up properly (cacheTime / staleTime)
2. We should avoid calling services in the UI and prefer controllers instead (controllers will decide whether to use cleeng ot inplayer under the hood). You can create a method in the controller (if it is not there yet) and use it.

test-e2e/tests/account_test.ts
```diff
@@ -349,10 +351,17 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string, res
...
 
-  function editAndCancel(I: CodeceptJS.I, editButton: string, fields: { name: string; startingValue: string; newValue: string; expectedError?: string }[]) {
+  async function editAndCancel(
+    I: CodeceptJS.I,
+    editButton: string,
+    fields: { name: string; startingValue: string; newValue: string; expectedError?: string }[],
+  ) {
     I.amOnPage(constants.accountsUrl);
-    I.click(editButton);
+    await waitForElementToBePopulated(I, `//*[text() = "${editButton}"]`, normalTimeout);
```
AntonLantukh: Hm, does it make sense for us just to wait for the text and then click on the button?
`waitForText` supports timeout, `click` function should also support it:
https://playwright.dev/docs/api/class-page#page-click

AntonLantukh:
@kiremitrov123 I also see blank name and surname when navigating to the account page directly (for a second or two).

src/components/Account/Account.tsx
kmitrovv: I can update this using useQuery, but we still need the fetchOnVisit (service level flag). We don't want this with cleeng. 
And also it will require creating new controller method, only for calling the getUser and updating the useAccountStore.. We do have user connected methods in the controller but they also do bunch of other calls we don't need, and I don't think it's a good idea to add additional checks there.

kmitrovv:
> @kiremitrov123 I also see blank name and surname when navigating to the account page directly (for a second or two).

@AntonLantukh , fixed

test-e2e/tests/account_test.ts
kmitrovv: Actually the issue is not for the edit button, but the fields that are being populated from the accounts endpoint.. all good, updated on other place, removing this.

darkoatanasovski (approved):

src/components/Account/Account.tsx
kmitrovv: updated

src/components/Account/Account.tsx
```diff
@@ -65,6 +66,9 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
     shallow,
   );
 
+  // refetch user data on visit (this is only JW integration case)
+  const { isLoading } = useQuery(['userInfo'], () => getUserInfo());
```
ChristiaanScheermeijer, line 70: This should be moved up one level. Stores/controllers should only be used in pages and containers.

src/stores/AccountController.ts
```diff
@@ -140,6 +140,20 @@ export const getAccount = async () => {
   });
 };
 
+export const getUserInfo = async () => {
+  return await useService(async ({ accountService }) => {
+    try {
+      const user = await accountService.getAccountInfo({ shouldFetchData: true });
+      if (user) {
+        useAccountStore.setState({ user });
+        return user;
```
ChristiaanScheermeijer, line 149: Maybe don't return the user from this method? It's assumed that the function returns the user data when called, but this will never be the case for the Cleeng integration.

The return data is not being used as of now, but we may make this mistake in the future.

kmitrovv: Yeah, there was an issue if the data was not returned, the isLoading state was not working properly or something, cant quite remember :D.
Thanks @ChristiaanScheermeijer  for the input, but this is on hold for now, will convert it to draft.

PR #385, opened 2023-10-05
Fix / notifications handler

kmitrovv:
## Description

This PR
-Improves overall the payment flow because the refresh page is eliminated after notification is received.
-Improves the notifications handling
-Fixes bugs about message handling with TVOD items
Jira cards: 
-[Improve TVOD flow for web app](https://jwplayer.atlassian.net/jira/software/c/projects/IDM/issues/IDM-5)
-[TVOD offer bugs (JW)](https://jwplayer.atlassian.net/browse/OWA-37)

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


darkoatanasovski (approved):

src/utils/formatting.ts
```diff
@@ -54,6 +54,17 @@ export const addQueryParams = (url: string, queryParams: { [key: string]: string
   return `${urlWithoutSearch}${queryString ? `?${queryString}` : ''}`;
 };
 
+export function removeQueryParam(key: string): string {
```
AntonLantukh: Looks like we already have it in `#src/utils/location`

src/stores/AccountController.ts
```diff
@@ -522,8 +521,6 @@ async function afterLogin(user: Customer, customerConsents: CustomerConsent[] |
     customerConsents,
   });
 
-  subscribeToNotifications(user.uuid);
```
AntonLantukh, line 525: I like that we removed this cross controllers dependency.

src/utils/formatting.ts
kmitrovv: that one uses location as a param, this one is intended to work with plain url

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -60,6 +61,22 @@ const AccountModal = () => {
     }
   }, [viewParam, loading, isPublicView, user, toLogin]);
 
+  // handle websocket notifications
+  useEffect(() => {
+    if (user) {
+      const handleNotification = async () => {
+        if (user) {
+          await subscribeToNotifications(user.uuid, async (navigateTo) => {
```
AntonLantukh: Do we automatically close socket connection on the logout? My idea was to do it when the user is not defined anymore.

kmitrovv: Yes, we do it along with the logout

kmitrovv: I don't think it's possible though.. If the user is not defined, we cannot call anything from the InPlayer sdk

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -60,6 +61,22 @@ const AccountModal = () => {
     }
   }, [viewParam, loading, isPublicView, user, toLogin]);
 
+  // handle websocket notifications
+  useEffect(() => {
```
budzworthy: I think this whole thing with NotificationsController could be refactored into a hook that could have hte useEffect and useNavigate built into it.

src/stores/NotificationsController.ts
```diff
@@ -1,30 +1,39 @@
...
+  CARD_SUCCESS = 'payment.card.success',
   SUBSCRIBE_REQUIRES_ACTION = 'subscribe.requires.action',
   FAILED = '.failed',
   SUBSCRIBE_FAILED = 'subscribe.failed',
+  SUBSCRIBE_SUCCESS = 'subscribe.success',
   ACCOUNT_LOGOUT = 'account.logout',
 }
 
-export const subscribeToNotifications = async (uuid: string = '') => {
+export const subscribeToNotifications = async (uuid: string = '', navigateTo: ({ url }: { url: string }) => void) => {
```
budzworthy: Can you just use the return type of useNavigate here?
```suggestion
export const subscribeToNotifications = async (uuid: string = '', navigateTo: NavigateFunction => void) => {
```

src/containers/AccountModal/AccountModal.tsx
budzworthy: I understand why this is in the Account Modal, but it might be nicer to put closer to / in the Root.tsx since it is such a 'global' thing.

src/utils/formatting.ts
budzworthy: Why not add a useLocation hook (assuming hook refactoring) and then you can use the existing function.

For the most part we are trying to prefer location, navigation, and searchParam hooks instead of direct `window.location.href` usage.

src/containers/AccountModal/AccountModal.tsx
kmitrovv: Yeah, that crossed my mind, but then when using it, we will need just to call it like useNotifications in the AppRoutes or somewhere, and not really use something from it, which is I guess not standard for hooks. wdyt?

src/stores/NotificationsController.ts
kmitrovv: that is a nice suggestion! thanks

src/containers/AccountModal/AccountModal.tsx
budzworthy: idk, I don't think it's that weird, but I would add a comment to the line where it's used.

kmitrovv: The dilemma was between AppRoutes and Account, but I guess in the account feels more isolated and all the needed stuff was already there.. It can go in the Root as well, but we need to add user store checks, navigation imports,.. etc.

src/utils/formatting.ts
kmitrovv: yeah, initially i was going with location, but I guess because it's inside of a promise, the location state is not set as latest (prorably stores the location from the initial load when connecting to websocket notifications) and all the query params get deleted when you try to add or remove the `u` query param.

src/containers/AccountModal/AccountModal.tsx
```diff
@@ -60,6 +61,20 @@ const AccountModal = () => {
...
+      const handleNotification = async () => {
+        if (user) {
+          await subscribeToNotifications(user.uuid, navigate);
+        }
+      };
+      handleNotification();
+    }
+    // we don't need to subscribe on each navigate change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [user]);
```
ChristiaanScheermeijer: When combining @dbudzins suggestion, you can shorten this to:

```suggestion
  useEffect(() => {   
    if (user) subscribeToNotifications(user.uuid, navigate);
    // we don't need to subscribe on each navigate change
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [user]);
```

I don't know if this should be a custom hook or `useEffect`. A custom hook would be nicer, and it would be totally fine if it doesn't return anything;

```tsx
useAccountSubscriptions(user, navigate);
```

I do have worries that this change also affects the Cleeng integration. This part is really InPlayer specific.

Are update payment details still working after this change?

kmitrovv: This shouldn't affect Cleeng at all, it's for the websocket notifications, where cleeng's method only returns true and doesn't do anything else.

src/utils/formatting.ts
budzworthy: Good point. Could it use `setSearchParams` from `useSearchParams`?

src/containers/AccountModal/AccountModal.tsx
budzworthy: If its a hook then all of those dependencies can go into the hook

ChristiaanScheermeijer: @dbudzins, perhaps we can change this in the services/inverify refactor (or a different cleanup PR 😄). It would be great to remove all integration-specific code from the UI and work only with generic Account, Checkout, Profile, and Subscription controllers.

For example, if we allow the account service to dispatch generic events, we can align the Cleeng and InPlayer integrations at service level. The Adobe SDK, for example, also works with global (window) callbacks that can be aligned via this method.

ChristiaanScheermeijer (approved):

ChristiaanScheermeijer (commented):
Wait... I think the hook needs a little change

src/containers/AccountModal/AccountModal.tsx
kmitrovv: updated with hook

darkoatanasovski:
@dbudzins @ChristiaanScheermeijer @AntonLantukh @kiremitrov123 Do we need to refactor the notifications feature in this PR, or should we focus on fixing the bug and improving the redirect after receiving the notification? If we want to enhance the feature, let's plan it next week. However, if we only need to address the bug in this PR, let's proceed with the merge because we don't have time today to refactor the entire feature. What do you think?

src/utils/formatting.ts
kmitrovv: Yes, makes sense, but I guess we should leave it for another cleanup update, because we have window.location in multiple places, and this bug fixing already end up with multiple unplanned changes.

PR #389, opened 2023-10-12
chore: fix changelog format, update preset, and improve release action

budzworthy:
## Description

This PR updates the github release actions.

First, the create release action is updated to use CHANGELOG as an input file, which preserves the existing content and appends just the new release, instead of regenerating the whole file each time. It also changes the preset to conventionalcommits because the angular preset uses h1 headings in markdown for some versions instead of always using h2 which breaks tools like changelog readers.

Accordingly, I manually updated the changelog to fix all of the h1 headings to be h2.

Lastly, I updated the release and tag action to read the current version and changelog changes from package.json and CHANGELOG.md respectively, instead of trying to rerun the changelog generator. This prevents the version from double incrementing and also means when you commit to the release branch, the release details created matches what you see in the release candidate branch.

### Steps completed:

~- [ ] Acceptance criteria met~
~- [ ] Unit tests added~
~- [ ] Translations added~
~- [ ] UX tested~
- [x] Reviewed own code

### DIFF
--- a/.github/workflows/release-tag-and-build.yml
+++ b/.github/workflows/release-tag-and-build.yml
@@ -28,7 +28,9 @@ jobs:
 
       - name: Get current package version
         id: package-version
-        uses: martinbeentjes/npm-get-version-action@v1.3.1
+        run: |
+          version=$( jq -r .version package.json)
+          echo "current-version=${version}" >> "$GITHUB_OUTPUT"
 
       - name: Parse Changelog Entry
         id: changelog-reader
--- a/.github/workflows/release-tag-and-build.yml
+++ b/.github/workflows/release-tag-and-build.yml
@@ -1,6 +1,7 @@
 name: Release - Tag and Build Artifacts
 
 on:
+  workflow_dispatch:
   push:
     branches: ['release']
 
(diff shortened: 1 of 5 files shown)


.github/workflows/release-tag-and-build.yml
```diff
@@ -32,18 +26,32 @@ jobs:
           tar czf ott-web-app-build-${RELEASE_VERSION}.tar.gz public
           zip -r ott-web-app-build-${RELEASE_VERSION}.zip public
 
+      - name: Get current package version
+        id: package-version
+        uses: martinbeentjes/npm-get-version-action@v1.3.1
```
AntonLantukh: We can probably use `jq` to parse the file:

```suggestion
      - name: Get current package version
        id: package-version
        run: |
          version=$( jq -r .version package.json)
          echo "current-version=${version}" >> "$GITHUB_OUTPUT"
```

And then use

AntonLantukh (approved):

.github/workflows/release-tag-and-build.yml
budzworthy: Good point, I guess we don't need a dependency for this

Issue #334, opened 2023-07-11
'Password strength score does not match standards' error is not handled by reset password form


ChristiaanScheermeijer:
When resetting your JWP password and entering a password with a "fair" strength, the request silently fails, and the user is not notified about the error.

The response from the reset password API is:

```json
{
    "code": 422,
    "message": "Password strength score does not match standards, please use a stronger password"
}
```

1. The error should be shown to users in the form
2. The error strength indicator should not show "fair" when not satisfying the password strength rules on the backend

**Edit**: the 'strong' and 'very strong' password strengths also fail the same restriction.

This restriction is not present when creating a new account in the registration flow.

budzworthy:
This is due to a disconnect between the FE feedback and BE validation. Need to decide on the best way to either unify the 2 or make one more primary.

kmitrovv:
Right, on BE we use a library for strong validation, which may have a different understanding of what constitutes 'strong' compared to the frontend. 
I believe addressing this on FE is a better approach, as modifying backend functionality could be more complex and potentially impact GDPR compliance.

ChristiaanScheermeijer:
@kiremitrov123 I believe this issue has been solved 🎉 . Can this be closed?

kmitrovv:
> @kiremitrov123 I believe this issue has been solved 🎉 . Can this be closed?

@ChristiaanScheermeijer Sure thing! 🎉

PR #395, opened 2023-11-09
fix(user): password strength on reset

kmitrovv:
Jira ticket: [Password strength error not handled by reset password form](https://jwplayer.atlassian.net/browse/IDM-25)
This was raised as an [issue](https://github.com/jwplayer/ott-web-app/issues/334).

Now we have corresponding validation message if the FE validations passes and the BE does not.
Preview example:
image

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/services/inplayer.account.service.ts
```diff
@@ -237,8 +237,9 @@ export const changePasswordWithOldPassword: ChangePasswordWithOldPassword = asyn
       errors: [],
       responseData: {},
     };
-  } catch {
-    throw new Error('Failed to change password.');
+  } catch (error: unknown) {
+    const { response } = error as InPlayerError;
+    throw new Error(response.data.message);
```
ChristiaanScheermeijer: I would check if the error is really an InPlayerError instead of type asserting it.

```suggestion
  } catch (error: unknown) {
    if (error is InPlayerError) {
      throw new Error(error.response.data.message);
    }
    throw new Error('Failed to change password');
```

Chances are small, but consider this example:

```ts
try {
  throw new InPlayerError(nofn(''));
} catch(error: unknown) {
  const { response } = error as InPlayerError;
  throw new Error(response.data.message);
}
```

The `nofn` call will trigger a ReferenceError, and the catch assumes it's an `InPlayerError`.

ChristiaanScheermeijer (approved):
Nice fix! I have one suggestion 😄

src/services/inplayer.account.service.ts
kmitrovv: Yeah, good point! Thanks for the suggestion!

borkopetrovicc (approved):

PR #363, opened 2023-08-14
feat(project): services modularization

AntonLantukh:
## Controllers and Services 

Controllers and Services can both be used to provide services (objects) that can be injectable into parts of the application.

**Services** - domain related entities. Business logic is stored there. We use services to communicate with the back-end, to process the data we receive. Services help to manage different dependencies. For example, we can use them to support several integration providers or streaming schedule providers. If this is the case we should also create a common interface and make dependant entities use the interface instead of the actual implementation. This is how inversion of control principle can be respected. Then when we inject services into controllers we use interfaces types instead of the implementation classes. 

- They contain the actual business logic to perform the use case;
- They can be injected into controllers which orchestrate different services;
- They should not used directly in the View / UI;
- One service can use different implementation. For example, we can split it into cleeng and jw implementation (account, checkout and so on).

**Controllers** - binding different parts of the application, using services, store and providing methods to operate with business logic in the UI and in the App. If we need to share code across controllers then it is better to promote the code to the next level. Then it is possible to modify both controllers to call the same (now shared) code. We should try to avoid controllers calling each other because it leads to circular dependencies and makes the code messy. However now they do it sometimes.

- They can be called from the View part of the application;
- They can use a Store to find the necessary data;
- They use the data from the Store and from the UI to operate different injected services;
- They use a Store to persist the entities again;
- They return data back to the UI if needed.

**Store**  - If the code is related to storage/retrieval, it should go in the Store. Both controllers and UI / View can use Store in accordance with their needs. _Something to think about_: use classes and inject them into controllers.

----------

## InversifyJS

InversifyJS library is used to provide IOC container and to perform dependency injection for both services and controllers. Injection happens automatically with the help of the reflect-metadata package.

`initDefaultServices` function is used to init default services and controllers which don't depend on any integration provider.

In the `initIOCData` function we initialise controllers based on the selected integration provider and inject necessary services.

### Steps completed:

According to our definition of done, I have completed the following steps:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Docs updated (including config and env variables)
- [ ] Translations added
- [ ] UX tested
- [ ] Browsers / platforms tested
- [ ] Rebased & ready to merge without conflicts
- [ ] Reviewed own code

PR #396, opened 2023-11-09
refactor(project): separate integration from services

ChristiaanScheermeijer:
@AntonLantukh @dbudzins I was trying some things out after discussing #363. I will leave it here as input or for inspiration 😄.

Things I like:

1. Root.tsx is becoming more and more simple (we can go further)
2. Integrations can be added without adding or changing multiple parts of the application
3. We can eliminate all the integration-specific code (access model calculation, etc.)
4. container.ts is placed in the src folder

Things I'm not sure about:

1. A new custom hook is introduced, `useBootstrapApp`
2. AppController might be renamed to ConfigController because it doesn't orchestrate the app bootstrap anymore (intended)
3. ...?


src/integrations.ts
```diff
@@ -0,0 +1,40 @@
...
+import SubscriptionJWService from '#src/services/integrations/jwp/inplayer.subscription.service';
+import CleengProfileService from '#src/services/integrations/cleeng/cleeng.profile.service';
+import InplayerProfileService from '#src/services/integrations/jwp/inplayer.profile.service';
+import ProfileService from '#src/services/profile.service';
+import { container, getModule } from '#src/container';
+import AppController from '#src/stores/AppController';
+
+const appController = getModule(AppController);
+
+appController.registerIntegration({
```
budzworthy: tbh, I don't see how this is any better than having all the bindings in 1 place  with some if's. You've gotta look in 2 files (this one and container.ts) to see all the bindings and a third file to see where those bindings are actually bound. It's more files and lines of code to call the exact same bind methods.

src/container.ts
```diff
@@ -0,0 +1,51 @@
...
+import FavoritesController from '#src/stores/FavoritesController';
+import AppController from '#src/stores/AppController';
+import { SettingsController } from '#src/stores/SettingsController';
+import type { interfaces } from 'inversify';
+
+export const container = new Container({ defaultScope: 'Singleton', skipBaseClassChecks: true });
+
+// resolve shortcut
+export const getModule = <T>(constructorFunction: interfaces.Newable<T>): T => {
+  return container.resolve(constructorFunction);
```
budzworthy: You might be right that we don't need all of our custom code for the resolution. @AntonLantukh can we try simplifying it and see if you branch still works?

src/hooks/useBootstrapApp.ts
```diff
@@ -0,0 +1,49 @@
...
+  config: Config;
+  configSource: string | undefined;
+  settings: Settings;
+};
+
+export const useBootstrapApp = (onReady: () => void) => {
+  const { data, isLoading, error, isError, isSuccess } = useQuery<Resources, Error>(
+    'config-init',
+    async () => {
+      const settings = await settingsController.initialize();
```
budzworthy: This is a nice simplification of the multiple queries into 1 sequential flow. I think you'll still have the problem of identifying a settings error from a config error, but that can be handled by seeing if settings is undefined or config is undefined.

src/stores/SettingsController.ts
budzworthy, line 1: I would call this a service because it's doing IO

src/hooks/useBootstrapApp.ts
```diff
@@ -0,0 +1,49 @@
...
+      const configSource = settingsController.getConfigSource(settings);
+      const config = await applicationController.initApp(configSource);
```
budzworthy: Can you combine these further? AppController already depends on a config service, but can also depend on the settings service. Actually config service could potentially depend on the settings service, so if you want to build one that doesn't care about the settings concept you could.

src/components/Root/Root.tsx
```diff
@@ -10,46 +9,18 @@ import DevConfigSelector from '#components/DevConfigSelector/DevConfigSelector';
...
-    refetchInterval: false,
-    retry: 1,
-    cacheTime: CACHE_TIME,
-    staleTime: STALE_TIME,
-  });
-
-  // Modify query string to add / remove app-config id
-  useTrackConfigKeyChange(data?.settings, data?.configSource);
-
+  const { data, isLoading, error, isSuccess } = useBootstrapApp(onReady);
```
budzworthy, line 21: I think with the cleanup you've done, leaving the single useQuery inline here is OK. A single use hook just hides a few lines of code.

src/services/integrations/cleeng/cleeng.account.service.ts
budzworthy, line 1: I think I do like this file layout better, but let's wait and move these after we merge the bulk of the module refactor. The merge conflicts are going to be bad enough without renames. I so desperately wish there was a way for JS to use reflection to lookup the files in each of these folders in order to dynamically bind an integration. Crazy long term idea, but maybe we could make a code generator to do that and run at build time?

src/container.ts
AntonLantukh: Hm, we might still want to avoid calling and creating controllers which are not supposed to exist / work (like CheckoutController for StartWatchingButton with AVOD).

budzworthy: Ah, good point!

ChristiaanScheermeijer: But that is different from what the code does, right? It checks the uniqueness of a service but does not conditionally return a service.

I would expect something like this then:

```ts
export const getModule = <T>(type: interfaces.Newable<T>): T => {
  const all = container.getAll(type);

  // don't resolve when there are no services registered
  if (!all.length) {
    return undefined;
  }

  return container.resolve(type);
};
```

The issue with the above is that all `getModule` calls will return `T | undefined`, and we will have to optional chain every service call.

IMO, we should prevent rendering components/containers that depend on account services conditionally, but I haven't thought that one through well.

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: Yes, that's correct. We can also use custom errors to make it more concrete.

src/stores/SettingsController.ts
ChristiaanScheermeijer: Yes, but this would only be allowed if we move the `setState` call to the AppController, right?

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: Yes, we could! Then the AppController would also make more sense as it orchestrates more than just the config logic.

src/components/Root/Root.tsx
ChristiaanScheermeijer: It's a small win 😄

src/services/integrations/cleeng/cleeng.account.service.ts
ChristiaanScheermeijer: Hmm, interesting thought...

src/integrations.ts
ChristiaanScheermeijer: Do you suggest merging the container.ts and integrations.ts files? The only reason for separating the two is that the integrations logic depends on the AppController. However, it is still possible to resolve it from the container file...

src/container.ts
budzworthy: Maybe the easiest way to do that is to allow getModule to return optionally undefined, and in the components that shouldn't render, simply short-circuit with 
```
if (!module) { return null; }
```

src/hooks/useBootstrapApp.ts
budzworthy: My preference would be the former instead of creating custom error classes. For just this use case that seems like overkill.

src/stores/SettingsController.ts
budzworthy: Maybe we should call it a Servtroller? :-)

I feel like maybe we should just add the handful of 'settings' into the config store. They are mostly tightly related to config already and that solves this problem for us and removes 1 extra store. What do you think?

src/integrations.ts
budzworthy: Yes, keeping all the bindings together in 1 file, but also, I think the registerIntegration logic is just adding unnecessary indirection. It would only make sense if we needed to dynamically create integrations at runtime, which we don't / can't because JS lacks any way to do the reflection or auto-discovery.

src/container.ts
ChristiaanScheermeijer: What about having a separate `getOptionalModule` to clarify that a service/controller is optional? For most services, we expect it to be present.

src/integrations.ts
ChristiaanScheermeijer: I am okay with combining the files.

About the registration part, this helps publishers register custom identity providers that aren't productized without changing the implementation code, for example, for an Adobe TVE integration. You're right that this can't be used runtime, but it does help with the implementation when publishers have forked the repository and want to keep sync.

I'm open to better suggestions! 😄

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: Agree!

src/stores/SettingsController.ts
ChristiaanScheermeijer: Hehe! That sounds good to me!

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: I've updated the PR with the changes proposed. What do you think?

src/stores/SettingsController.ts
ChristiaanScheermeijer: I've updated the PR with the changes proposed. What do you think?

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: Actually, there already are custom errors: `ConfigError` and `SettingsError` 😅

AntonLantukh: Yeah, that is what I started with. We have discussed it here:

https://github.com/jwplayer/ott-web-app/pull/363#discussion_r1386553458

We can use `undefined` for config / settings, but then we won't be able to differentiate between different config's errors (is it missing or is api broken?). Though it is maybe not as important.

src/services/integrations/cleeng/cleeng.account.service.ts
AntonLantukh: Yeah, that was my first intention to create some folders and also a separate folder for store / controllers. However it didn't work well when I had to merge new changes from `develop`...

src/container.ts
AntonLantukh: Danny suggested to use optional `required` property and to throw the error in case it is required and not found:

https://github.com/jwplayer/ott-web-app/pull/363#discussion_r1386573736

src/hooks/useBootstrapApp.ts
AntonLantukh: Nice! Looks like we can also make SettingsService optional, right?

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: Ah, yes, I see... There is even more discussion overlap. Sorry about that!

src/container.ts
ChristiaanScheermeijer: Yes, I like that solution as well! 😄

src/hooks/useBootstrapApp.ts
ChristiaanScheermeijer: We could, but the SettingsService is now responsible for determining the config id (`configSource`). We would need to refactor that. Do you have suggestions? 😄

Thinking about this, we could swap the `SettingsService` for production builds with a static service. For example;

```ts
class DemoSettingsService extends SettingsService {
  // current SettingsService with URL search param logic
}

class SettingsService {
  initialize = async () {
    return {
      defaultConfigSource: import.meta.env.APP_DEFAULT_CONFIG_SOURCE,
      playerId: import.meta.env.APP_PLAYER_ID || OTT_GLOBAL_PLAYER_ID,
      playerLicenseKey: import.meta.env.APP_PLAYER_LICENSE_KEY,
    };
  }

  getConfigSource(settings: Settings | undefined) {
    if (!settings) {
      return '';
    }

    return settings?.defaultConfigSource;
  }
}
```

But that's for later I guess 😅

src/integrations.ts
budzworthy: I think your latest proposal might already solve this, but just to finish the thread - I would be OK with integrations having individual files where they register themselves (so the main app references that 1 file where the integration binds its own modules.) I don't think we need to add an extra layer of decoupling with the `appController.registerIntegration` logic.

src/services/integrations/cleeng/cleeng.account.service.ts
ChristiaanScheermeijer: Yes, that's a risk indeed. We can do this after we've merged most PRs

src/integrations.ts
ChristiaanScheermeijer: Yes, I agree! I did split the container initialization from the registration/bindings to fix some import issues we will be having later. The core only cares about the container instance and `getModule` helper but shouldn't care how the dependencies are bound.

PR #405, opened 2023-11-20
feat(project): add ViewNexa epg provider

AntonLantukh:
## Description

- add ViewNexa epg provider

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


src/services/epg/epgClient.service.ts
```diff
@@ -18,25 +17,16 @@ export const isFulfilled = <T>(input: PromiseSettledResult<T>): input is Promise
...
 @injectable()
-export default class EpgService {
+export default class EpgClientService {
+  private viewNexaProvider: EpgProviderService;
+  private jwProvider: EpgProviderService;
+
+  public constructor(@targetName(EPG_TYPE.JW) jwProvider: EpgProviderService, @targetName(EPG_TYPE.VIEW_NEXA) viewNexaProvider: EpgProviderService) {
+    this.viewNexaProvider = viewNexaProvider;
+    this.jwProvider = jwProvider;
+  }
```
AntonLantukh: @dbudzins @ChristiaanScheermeijer not ideal, but maybe you have ideas how we could dynamically choose which provider to use depending on the media item.

ChristiaanScheermeijer: This can work the same as for the integration, right?

```ts
public loadProvider(item: PlaylistItem) {
  this.epgService = getModule(EpgService, getEpgName(item));
}
```

However, I would prefer something other than determining the EPG provider per media item but rather do it on the app config. With this change, only one EPG provider can be active at the time and can be loaded in the constructor instead.

The service modularization doesn't pay off completely if we still need to update the `epgClient` to add or replace certain services.

Ideally, when you want to implement a custom EPG service, only an update of the `container/register.ts` file would be needed 😄

AntonLantukh: Yeah, that was my first intention to use `getModule` in the client class, but I would still prefer a native way of DI than using this trick. App Config definition could be an option, though in this case both implementations can't be used together (will clarify if it could be a problem).

AntonLantukh: It is a possible use case when a customer adds two channels of different providers into one playlist.
So we will have to keep it on the media level.

ChristiaanScheermeijer: In this case, it would still make sense to use the same mechanism as that used for integrations. But ... currently, the same issue exists there;

If you want to introduce your own EPG provider or Integration, you must make changes in multiple files. I'm hoping we can find a solution that only requires adding the service, and the rest of the application knows how to process this.

I'm thinking about making the integration configuration extendable. Something like this:

```ts
type Config = {
  ...
  integrations: {
    cleeng?: Cleeng;
    jwp?: JWP;
    [key: string]: unknown;
  };
};
```

We know what to expect from the Cleeng and JWP configurations, but when integrating Adobe, we don't know what properties are needed. Validation of these properties should be done in the service.

The rest of the application shouldn't deal with the integration configuration. The integration services should provide all the information needed by the application.

Because the `integrations` type contains basically a wildcard for any configuration, we could look up the first key in the `integrations` object and use that as the "named resource".

This would also work for EPG where the `scheduleType` or `providerName` is used to request the correct EPG provider with a fallback to the JWP provider.

src/services/epg/epgClient.service.ts
budzworthy, line 1: I'm confused about the renaming on this class. What is a client service? (I think of client to mean lowest level api wrapper. I think what is currently called api.service and cleeng.service used to be or were at one point api.client and cleeng.client for example.)  Can we just call this an epg controller?

src/services/epg/epgClient.service.ts
```diff
@@ -18,25 +17,16 @@ export const isFulfilled = <T>(input: PromiseSettledResult<T>): input is Promise
...
-  ),
-});
-
 @injectable()
-export default class EpgService {
+export default class EpgClientService {
+  private viewNexaProvider: EpgProviderService;
+  private jwProvider: EpgProviderService;
+
+  public constructor(@targetName(EPG_TYPE.JW) jwProvider: EpgProviderService, @targetName(EPG_TYPE.VIEW_NEXA) viewNexaProvider: EpgProviderService) {
```
budzworthy: I think this is the perfect use case for getting multiple modules by type of EpgProviderService. This class can just get all of the EpgProviderService's that are defined and put them into a dictionary by 'type' (assuming we require each one to define a 'type' property.)

src/services/epg/epgClient.service.ts
```diff
@@ -152,6 +102,10 @@ export default class EpgService {
     } as EpgChannel;
   };
 
+  getScheduleProvider = (item: PlaylistItem) => {
+    return item?.scheduleType === EPG_TYPE.VIEW_NEXA ? this.viewNexaProvider : this.jwProvider;
```
budzworthy: If we have a dictionary of epg services by type, this code can just grab the matching one (we probably also need fallback logic to someone define the 'default' one when none is set.)

src/services/epg/epgClient.service.ts
ChristiaanScheermeijer: That should work, but then we do want to prevent any imports/initialization from ending up in the constructor. For example, loading/importing a ViewNexa SDK.

src/services/epg/epgClient.service.ts
budzworthy: I'm not sure if we need to worry about supporting multiple epg types simultaneously. If so, I think grabbing them all and putting them in a dictionary by type is a solid pattern (see above.) But honestly, I doubt customers would try to use 2 different providers, which means we can define it at an app level (probably just as an app config custom param.) Then we can do a similar thing to what we've already done with the integration type.

src/services/epg/epgProvider.service.ts
budzworthy, line 1: I would remove the word 'provider' from this class name, because provider can mean other things architecturally speaking. Just epg.service is probably fine.

src/services/epg/epgClient.service.ts
budzworthy: @ChristiaanScheermeijer I think you're onto an interesting thought with your integration suggestion. I would take it maybe a half step further. What if we take every key from 'integrations' and use it to dynamically load a named class from an 'integrations' directory? Something like this in `AppController.initializeApp`:

```
    await Promise.allSettled(
      Object.entries(config.integrations).map(async ([integrationName, settings]) => {
        const integration = await import(`#src/integrations/${integrationName}.integration.ts`);

        integration.init(settings);
    }));
```

Honestly I didn't think this would work, but crazily enough it does.
Benefits:
* Apps can define as many integrations as they want and as long as they result in a valid composition, the app code will 'just work'
* Each integration can fully encapsulate it's own initialization code
* Each integration can expect whatever settings it needs and they will get passed generically through
* Easy to add new integrations without touching any other files

Drawbacks:
* Need to refactor the modularization code already
* 'Common' module initialization and integration module initialization will be separated
* A little bit magical / non-deterministic

src/modules/container.ts
```diff
@@ -15,10 +13,10 @@ export function getModule<T>(constructorFunction: interfaces.ServiceIdentifier<T
   return module;
 }
 
-export function getNamedModule<T>(constructorFunction: interfaces.ServiceIdentifier<T>, integration: IntegrationType | null, required: false): T | undefined;
-export function getNamedModule<T>(constructorFunction: interfaces.ServiceIdentifier<T>, integration: IntegrationType | null, required: true): T;
-export function getNamedModule<T>(constructorFunction: interfaces.ServiceIdentifier<T>, integration: IntegrationType | null): T;
-export function getNamedModule<T>(constructorFunction: interfaces.ServiceIdentifier<T>, integration: IntegrationType | null, required = true): T | undefined {
+export function getNamedModule<T>(constructorFunction: interfaces.ServiceIdentifier<T>, integration: string | null, required: false): T | undefined;
```
budzworthy:
```suggestion
export function getNamedModule<T>(constructorFunction: interfaces.ServiceIdentifier<T>, name: string | null, required: false): T | undefined;
```

src/services/epg/viewNexaEpg.service.ts
```diff
@@ -0,0 +1,79 @@
...
+@injectable()
+export default class ViewNexaEpgService extends EpgProviderService {
+  /**
+   * Validate the given data with the viewNexaProgramSchema and transform it into an EpgProgram
+   */
+  transformProgram = async (data: unknown): Promise<EpgProgram> => {
+    const program = await viewNexaEpgProgramSchema.validate(data);
+
+    return {
+      id: program['episode-num']['#text'],
```
budzworthy: Do we need to worry about null refs here (i.e. if `program['episode-num']` is null) or does the parser handle that?

src/services/epg/epgClient.service.ts
AntonLantukh: Of course! Don't know why I did it differently here.

src/services/epg/viewNexaEpg.service.ts
AntonLantukh: I would expect all properties to be present in the xml response. We have a `yup` validation schema which will be triggered when some of required fields are missing.

AntonLantukh: Made `description`, `cardImage` and `backgroundImage` optional using chaining (that is what we expect in the EpgProgram type)

budzworthy: OK, and does the yup schema validation throw a fatal error or do we just not show the epg for broken schemas?

src/services/epg/epgClient.service.ts
AntonLantukh: Used `multiInject` to add services.

src/services/epg/viewNexaEpg.service.ts
AntonLantukh: The page itself is still shown. We filter items where the validation failed (`parseSchedule` method)

src/services/epg/epgClient.service.ts
AntonLantukh: I think the main reason is that part of the business logic is stored here (`parseSchedule` where we transform items and filter them). So this method can be potentially reused on other platforms where we don't have access to controllers.

src/services/epg/epgClient.service.ts
AntonLantukh: Used multiInject decorator here.

src/stores/EpgController.ts
```diff
@@ -102,8 +99,15 @@ export default class EpgClientService {
     } as EpgChannel;
   };
 
-  getScheduleProvider = (item: PlaylistItem) => {
-    return item?.scheduleType === EPG_TYPE.VIEW_NEXA ? this.viewNexaProvider : this.jwProvider;
+  getEpgService = (item: PlaylistItem) => {
+    const scheduleType = item?.scheduleType || EPG_TYPE.JWP;
+    const service = this.epgServices.find((service) => service.type === scheduleType);
```
budzworthy: What do you think about removing the type field and just using the name registered in the DI container as the type lookup?

AntonLantukh: Used `getNamedModule` instead of multiInject. Please have a look

src/stores/EpgController.ts
```diff
@@ -152,6 +94,17 @@ export default class EpgService {
     } as EpgChannel;
   };
 
+  getEpgService = (item: PlaylistItem) => {
+    const scheduleType = item?.scheduleType || EPG_TYPE.JWP;
+    const service = getNamedModule(EpgService, scheduleType);
```
budzworthy: Should we make this case insensitive?

src/stores/EpgController.ts
```diff
@@ -152,6 +94,17 @@ export default class EpgService {
...
+
+    if (!service) {
+      throw Error(`No epg service was added for the ${scheduleType} schedule type`);
```
budzworthy: Will this cause the whole app to crash if someone mis-types the epg type? If so, can we make it more resiliant and just log a warning adn return an empty schedule?

AntonLantukh: Makes sense to me

budzworthy (approved):

PR #416, opened 2023-12-05
fix(series): fix 'hasMore' param

AntonLantukh:
## Description

- fix 'hasMore' param

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

### DIFF
--- a/src/components/CardGrid/CardGrid.tsx
+++ b/src/components/CardGrid/CardGrid.tsx
@@ -11,7 +11,6 @@ import type { AccessModel } from '#types/Config';
 import type { Playlist, PlaylistItem } from '#types/playlist';
 import { parseAspectRatio, parseTilesDelta } from '#src/utils/collection';
 import InfiniteScrollLoader from '#components/InfiniteScrollLoader/InfiniteScrollLoader';
-import { isDefined } from '#src/utils/common';
 
 const INITIAL_ROW_COUNT = 6;
 const LOAD_ROWS_COUNT = 4;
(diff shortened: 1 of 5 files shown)


src/components/CardGrid/CardGrid.tsx
```diff
@@ -90,9 +91,14 @@ function CardGrid({
   };
 
   return (
-    <InfiniteScroll pageStart={0} loadMore={loadMore || defaultLoadMore} hasMore={hasLoadMore || defaultHasMore} loader={<InfiniteScrollLoader key="loader" />}>
+    <InfiniteScroll
+      pageStart={0}
+      loadMore={isDefined(loadMore) ? loadMore : defaultLoadMore}
```
budzworthy: Is there any difference to this?
```suggestion
      loadMore={loadMore ?? defaultLoadMore}
```

src/components/CardGrid/CardGrid.tsx
```diff
@@ -90,9 +91,14 @@ function CardGrid({
...
+      hasMore={isDefined(hasMore) ? hasMore : defaultHasMore}
+      loader={<InfiniteScrollLoader key="loader" />}
+    >
       <div className={classNames(styles.container, styles[`cols-${visibleTiles}`])} role="grid">
-        {playlist.playlist.slice(0, rowCount * visibleTiles).map(renderTile)}
+        {(isDefined(loadMore) ? playlist.playlist : playlist.playlist.slice(0, rowCount * visibleTiles)).map(renderTile)}
```
budzworthy: The logic on this looks backwards to me. Can you explain / maybe comment?

AntonLantukh: When `loadMore` function is defined, then we use the accumulated data we received from the back-end and don't need to hide some of the cards manually. We hide them manually for playlists where we receive all the data at once (no pagination), but don't want to render cards which are not visible.

budzworthy: Oh OK I understand. Maybe worth a quick comment in the code. This is the 'fix' for the bug, yes?

AntonLantukh: Will do. Not only this bug, we also used `defaultHasMore` when we getting `false` from the series endpoint.

src/components/CardGrid/CardGrid.tsx
AntonLantukh: It checks both `null` and `undefined`. My idea was to filter `undefined` props, but I don't really have strong preference here.

budzworthy: Ooh, I see. OK. I guess it's potentially nice to be able to set it to null, though I'm not sure what the use case would be here specifically.

AntonLantukh: Yeah, that is why I removed it here and used nullish coalescing operator instead.

budzworthy (approved):

PR #418, opened 2023-12-06
feat(watchhistory): change max items limit

AntonLantukh:
## Description

- 48 items for Cleng, JWP and local storage

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


kmitrovv (approved):
lgtm

src/config.ts
```diff
@@ -23,8 +23,11 @@ export const VideoProgressMinMax = {
 
 export const PLAYLIST_LIMIT = 25;
 
-// The externalData attribute of Cleeng can contain max 5000 characters
-export const MAX_WATCHLIST_ITEMS_COUNT = 48;
+export const MAX_WATCHLIST_ITEMS_COUNT = {
+  CLEENG: 38, // The 'externalData' attribute of Cleeng can contain max 4000 characters
+  JWP: 50, // Limit of media_ids length passed to the /apps/watchlists endpoint
+  DEFAULT: 38, // Local storage
+};
```
ChristiaanScheermeijer: Shouldn't this be defined in the integration? I'm hoping we can remove all integration references from the non-service code.

docs/features/user-watchlists.md
```diff
@@ -127,10 +127,16 @@ Example data format
...
+#### JWP
 
-The length of one stringified object of History equals to 52 symbols, one Favorites object equals to 22 symbols. Taking into account only History objects, we get 5000 / 52 = ~96, so 48 for Favorites and 48 for History. We also leave some extra space for possible further updates.
+For JWP the limit is 50 items for Favorites and 50 for Watch History.
+
+#### Cleeng
+
+Cleeng customer `externalData` attribute has maxsize of 4000 symbols.
+
+The length of one stringified object of History equals to 52 symbols, one Favorites object equals to 22 symbols. Taking into account only History objects, we get 4000 / 52 = ~76.92, so 38 for Favorites and 38 for Watch History. We also leave some extra space for possible further updates.
```
ChristiaanScheermeijer: Existing platforms might also use the `externalData` property to persist specific data. Will there be enough room for this (knowing that favorites consume less than 52 characters)?

The size of a serialized watch history item is not static. This is because of the rounding of the progress.

I see an optimization we can do here. Limit the progress decimal places to 5.

This still results in a millisecond precision for a 2-hour video:

```ts
const serialized = '{"mediaid":"awWEFyPu","progress":0.02227576362112322}';

7200 * 0.02227576362112322;
// 160.3854980720872

7200 * 0.0222;
// 160.34400000000002

serialized.length;
// 53

'{"mediaid":"awWEFyPu","progress":0.02227}'.length; 
// 41
```

What will happen if the `externalData` property already contains 2000 characters and we add more on top?

ChristiaanScheermeijer: We can make `AccountService#updatePersonalShelves` throw an exception when it's "full" and show a message? Then the integration decides when to stop adding items and this will also work for the watch history. 

In the CleengAccountService we can even do this based on the current `externalData` length.

AntonLantukh: Nice optimisation, thanks! I will add it.

4000 / 41 = ~97.56, so 48 for Favorites and 48 for Watch History
We leave some extra space using Favorites here:
1056 (48 Favorites items) + 1968 (48 WatchHistory items) = 3024

src/config.ts
AntonLantukh: Will add it to integrations.

docs/features/user-watchlists.md
AntonLantukh: It doesn't look like there is any error handling right now for the cases when `externalData` is full. I would add it in a separate PR. It doesn't look like this one should be a message in the UI. Customer can't really do anything with it. We can log it to the console.

budzworthy (approved):

PR #420, opened 2023-12-20
feat: add user_id and profile_id for CDN analytics

olga-jwp:
## Description

[OWA-52](https://jwplayer.atlassian.net/browse/OWA-52)

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

[OWA-52]: https://jwplayer.atlassian.net/browse/OWA-52?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Player/Player.tsx
```diff
@@ -161,6 +172,23 @@ const Player: React.FC<Props> = ({
 
       playerRef.current = window.jwplayer(playerElementRef.current) as JWPlayer;
 
+      // Add user_id and profile_id for CDN analytics
+      const { sources } = item;
+
+      sources?.map((source) => {
```
AntonLantukh: I think we should make these calculations part of the service / controller, either ApiService or part of the new AnalyticsController responsible for the calculations. That would help to reuse this logic for other platforms in the future. @dbudzins do you have a preference?

budzworthy: Analytics controller sounds good to me, but also a simple util method would be fine. Whatever fits better with the architecture.

src/components/Player/Player.tsx
```diff
@@ -161,6 +172,23 @@ const Player: React.FC<Props> = ({
...
+        const url = new URL(source.file);
+        const isVOD = url.pathname.includes('manifests');
+        const isBCL = url.pathname.includes('broadcast');
+        const isDRM = url.searchParams.has('exp') || url.searchParams.has('sig');
```
budzworthy: Should we use the app config drm property for this instead of this check which may be a bit flaky?

src/components/Player/Player.tsx
```diff
@@ -161,6 +172,23 @@ const Player: React.FC<Props> = ({
 
       playerRef.current = window.jwplayer(playerElementRef.current) as JWPlayer;
 
+      // Add user_id and profile_id for CDN analytics
+      const { sources } = item;
+
+      sources?.map((source) => {
+        const url = new URL(source.file);
+        const isVOD = url.pathname.includes('manifests');
```
budzworthy: Do we need to worry about case sensitivity here?

src/components/Player/Player.tsx
```diff
@@ -161,6 +172,23 @@ const Player: React.FC<Props> = ({
...
+        const isBCL = url.pathname.includes('broadcast');
+        const isDRM = url.searchParams.has('exp') || url.searchParams.has('sig');
+
+        if ((isVOD || isBCL) && !isDRM && userId) {
+          url.searchParams.set('user_id', userId);
+          if (isJwIntegration && profileId) url.searchParams.append('profile_id', profileId);
```
budzworthy: I generally recommend to avoid single line if's and if's w/o brackets

src/components/Player/Player.tsx
```diff
@@ -161,6 +173,10 @@ const Player: React.FC<Props> = ({
 
       playerRef.current = window.jwplayer(playerElementRef.current) as JWPlayer;
 
+      // Inject user_id and profile_id into the CDN analytics
+      const { sources, mediaid } = item;
+      attachAnalyticsParams(sources, mediaid, isJwIntegration, userId, profileId);
```
budzworthy: Is there any reason you left the store lookups in this file and added them all as parameters? I think it would be safe for you to move everything into the util and just have this be a 1 line call like `attachAnalyticsParams(item)`

src/utils/player.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    // Attach user_id and profile_id only for VOD and BCL SaaS Live Streams
+    const isVOD = url.href === `https://cdn.jwplayer.com/manifests/${mediaid}.m3u8`;
+    const isBCL = url.href === `https://content.jwplatform.com/live/broadcast/${mediaid}.m3u8`;
+
+    if ((isVOD || isBCL) && userId) {
+      url.searchParams.set('user_id', userId);
+
+      if (isJwIntegration && profileId) {
+        url.searchParams.append('profile_id', profileId);
```
budzworthy: I think you're safer to you `set` here again. There shouldn't be other query params on this item called profile_id, but if so, we should only keep the latest.
```suggestion
        url.searchParams.set('profile_id', profileId);
```

src/utils/player.ts
budzworthy, line 1: I would probably call this file analtyics.ts

budzworthy (commented):
A couple small comments, but looks like it's almost there!

src/components/Player/Player.tsx
olga-jwp: Actually, the store lookups  can't be called inside of the `attachAnalyticsParams()` as it's called inside of the `useEffect` and hooks can only be called inside of the body of a function component.

budzworthy: I think you can, zustand is just a little tricky and obscure about it: https://github.com/pmndrs/zustand?tab=readme-ov-file#readingwriting-state-and-reacting-to-changes-outside-of-components

You want something like `useConfigStore.getState()`.  Like you said, it won't be in the hook so it won't trigger re-renders, but integration should change and user id / profile id only change when the user logs out which probably can't happen while watching a video unless it's a background thingy triggered by inplayer sdks, in which case we probably would just want to keep logging with the user id and profile id captured at the start.

src/utils/analytics.ts
```diff
@@ -0,0 +1,23 @@
+import type { PlaylistItem, Source } from '#types/playlist';
+
+export const attachAnalyticsParams = (item: PlaylistItem, isJwIntegration: boolean, userId?: string, profileId?: string) => {
+  const { sources, mediaid } = item;
+
+  return sources.map((source: Source) => {
+    const url = new URL(source.file);
```
dbudzins, line 7: Just to be safe with case sensitivity:
```suggestion
    const url = new URL(source.file.toLowerCase());
```

src/utils/analytics.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    // Attach user_id and profile_id only for VOD and BCL SaaS Live Streams
+    const isVOD = url.href === `https://cdn.jwplayer.com/manifests/${mediaid}.m3u8`;
+    const isBCL = url.href === `https://content.jwplatform.com/live/broadcast/${mediaid}.m3u8`;
```
budzworthy: Sorry for the noise, but slightly different suggestion to avoid case sensitivity w/o changing the source URL's which are case sensitive:
```suggestion
    const isVOD = url.href.toLowerCase() === `https://cdn.jwplayer.com/manifests/${mediaid.toLowerCase()}.m3u8`;
    const isBCL = url.href.toLowerCase() === `https://content.jwplatform.com/live/broadcast/${mediaid.toLowerCase()}.m3u8`;
```

src/components/Player/Player.tsx
olga-jwp: I'll try it, thnx!

budzworthy (approved):

PR #435, opened 2024-01-18
Feat / Restructure and a11y

ChristiaanScheermeijer:
## Description

Sorry in advance for dropping this PR 😅... 

I initially planned to rebase the changes from the develop branch for a linear history, but unfortunately, there were too many commits and conflicts.

We talked about restructuring the OTT Web App for multi-platform purposes a few weeks ago. A lot of code can be reused for different platforms, but because it is currently a single package (web platform), it's hard to extract the code for a, let's say React Native app.

Most of the structure changes you've seen and reviewed already. We've worked hard for the last 3 weeks to finalize the restructure by cleaning up dependencies and ensuring the common package doesn't use browser typings or React.

Besides that, we've also addressed a lot of accessibility bugs. We aim to achieve at least WCAG 2.1 AA, but we're not there yet. In the next sprint we will address even more accessibility issues and submit this later as PR(s). The accessibility fixes are mainly applied to components and SCSS files.

**The most significant changes made since you last reviewed the restructure are:**

1. Remove React and browser typings from the @jwp/ott-common package
2. Move all SVG icons to the @jwp/ott-theme package
3. Refactor the query param utils (we had multiple overlapping utils for the same purpose)
4. Remove react-router from the @jwp/ott-common and @jwp/ott-hooks-react packages
5. Remove react-query from the @jwp/ott-common package (see below)
6. Refactor the integration registration to allow registering custom integrations more easily
7. Show a nice player error when the media is GEO blocked for the user

**We also addressed some issues found while testing:**

1. Cleeng login is really slow due to an expected failing `listProfiles` query with a few retries
2. Paywall not working correctly for TVOD offers in the inline player variant
3. User loading state never reset after logout or account deletion

We have found a few more problems but didn't have the time to fix these in this sprint. We've confirmed that these bugs are currently also present in the latest release.

There is 1 more PR open (almost ready to merge) that fixes a problem we've had while testing but was harder to fix, unfortunately.

https://github.com/Videodock/ott-web-app/pull/31

PS. You can also view each PR (with reviews) separately: https://github.com/Videodock/ott-web-app/pulls?q=is%3Apr+is%3Aclosed

In the next sprint, we'll continue working on more a11y issues and refactoring some services, controllers, containers, and hooks. Specifically, we want to move most logic from components, containers and hooks to services/controllers.


.github/workflows/release-build-tag-release.yml
budzworthy, line 1: I may have already asked / suggested this, but does it make sense to and are we able to use subfolders in actions? I don't know what's possible, but the 2 patterns that come to mind are:
1) Subfolders inside of .github/workflows for each platform (i.e. /web for the web build/test/release actions)
2) If it's supported by github, have separate .github/actions directories in each of the platform directories
3) At the very least if the above 2 options don't work, let's prefix the actions by platform ('web-test-e2e')

docs/initialization-file.md
budzworthy, line 1: Does it make sense to organize the docs by platform? Or at least mark certain sections as "Web Only" or something?

vitest.workspace.ts
```diff
@@ -0,0 +1 @@
+export default ['packages/*/vitest.config.ts', 'platforms/*'];
```
budzworthy, line 3: Nice to add a comment on ths file

budzworthy (approved):
This one actually wasn't too bad to review. I added a few small questions just to add more comments and details around things.  Otherwise, I think I would ask if we can start a doc file to summarize the monorepo workspace approach. A couple things to describe there would be:

- Overall structure, workspace layout, approach for monorepo
- the eslintrc architecture and reasons
- syncpack - usage and reasons
- vite workspace
- tsconfig heirarchy

ChristiaanScheermeijer:
Thank you very much for reviewing @dbudzins! I'm glad to hear that it wasn't too bad 😄 

I think all your comments make sense. I will work on that soon. Do you want to do that before merging this PR, or can I submit a separate PR to develop?

.github/workflows/release-build-tag-release.yml
ChristiaanScheermeijer: I haven't looked into this for long, but according to this [discussion](https://github.com/orgs/community/discussions/18055), subfolders are not possible (automatically).

Workflows are only read from the root `.github/workflows` directory. However, workflows can reference composite actions in a different folder.

> Reusable workflows can only be defined in the `.github/workflows` directory. Subdirectories are not supported.

So we could do something like this:

**.github/workflows/test-e2e.yml**
```yml
name: Test - End to End

on:
  pull_request:
  push:
    branches: [ 'develop', 'release' ]
  schedule:
    - cron: '30 3 * * 1-5'
  workflow_dispatch:

jobs:
  test-e2e:
    runs-on: ubuntu-latest

    strategy:
      fail-fast: false
      matrix:
        node-version: [18.x]
        platform: [desktop, mobile]

    steps:
      - uses: actions/checkout@v4
      - id: web-test
        uses: ./platforms/web/actions/test-e2e
        with: 
          platform: ${{ matrix.config }}
          node-version: ${{ matrix.node-version }}
```

**./platforms/web/actions/test-e2e.yml** or **.github/actions/web/test-e2e.yml**
```yml
name: "Web - test e2e"
description: "Run e2e tests for the web platform"
inputs:
  platform:
    description: "Desktop or mobile"
    required: true
  node-version:
    description: "The node version to use"
    required: true
runs:
  using: "composite"
  working-directory: ./platforms/web
  steps:
    - name: Use Node.js ${{ inputs.node-version }}
      uses: actions/setup-node@v3
      with:
        node-version: ${{ inputs.node-version }}
    - name: Install dependencies
      run: |
        yarn
        yarn global add wait-on
    - name: Start preview server
      run: yarn start:test &
    - name: Run tests
      run: wait-on -v -t 60000 -c ./scripts/waitOnConfig.js http-get://localhost:8080 && yarn codecept:${{ matrix.config }}
      env:
        TEST_RETRY_COUNT: 2
        WORKER_COUNT: 2
    - name: Uploading artifact
      if: always()
      uses: actions/upload-artifact@v3
      with:
        name: allure-report-${{ matrix.config }}
        path: ./platforms/web/test-e2e/output/${{ matrix.config }}
        retention-days: 7
```

The only "link" between the main workflows and platform workflows is the reference to the composite action and its inputs. It seems impossible to have the workflows completely separated per platform.

I do think the workflows are meant for JWP releases only. Most users will have their own CI/CD or use different workflows anyway (if this is one of the reasons).

Perhaps prefixing it is the easiest fix for now because creating composite actions will take quite some time to get up and running. What do you think?

ChristiaanScheermeijer:
@dbudzins I've merged develop and also created a PR in our fork with some documentation updates:

https://github.com/Videodock/ott-web-app/pull/56

.github/workflows/release-build-tag-release.yml
budzworthy: Prefixing is easy and fine

budzworthy:
> Thank you very much for reviewing @dbudzins! I'm glad to hear that it wasn't too bad 😄
> 
> I think all your comments make sense. I will work on that soon. Do you want to do that before merging this PR, or can I submit a separate PR to develop?

I think it would be best if we have at least a rough version of an overview doc included with the first merge, in case anyone tries to grab the latest and start working with it, they'll be able to navigate the basics. It doesn't have to be a full, in depth doc, we can always add and refine later.

ChristiaanScheermeijer:
@dbudzins, the docs are updated to reflect the directory changes. I've also added an initial version of the workspaces documentation.

I've prefixed three workflows that are exclusively web only. The only questionable workflow is `release-build-tag-release.yml` which creates web build artifacts. But this workflow should only be used once because it also creates a tag and release.

Let me know what you think! 😄

docs/workspaces.md
```diff
@@ -0,0 +1,202 @@
...
+- Easy linking of packages/dependencies
+
+## The solution
+
+Based on the re-usability of parts of the existing codebase, we've created separate packages using Yarn Workspaces.
+This will combine all similar code and prevent installing redundant or conflicting dependencies.
+
+For example, all components, containers, and pages are be combined into the `packages/ui-react` package, which depends
+on
+react and react-dom. To create a React Native app, you could add an `packages/ui-react-native` package and configure
```
AntonLantukh:
```suggestion
For example, all components, containers, and pages are be combined into the `packages/ui-react` package, which depends on react and react-dom. To create a React Native app, you could add an `packages/ui-react-native` package and configure
```

docs/workspaces.md
```diff
@@ -0,0 +1,202 @@
+# Workspaces
+
+## Why workspaces?
+
+The JW OTT Web App is an open-source repository that showcases an OTT app implementing JWP services. The OTT Web App, as
+the name implies, originates as a web only repository. But much of the source-code can be re-used for many different
+platforms. Like; Capacitor, React Native, and other frameworks based on TypeScript.
```
AntonLantukh:
```suggestion
platforms, like Capacitor, React Native, and other frameworks based on TypeScript.
```

i18next-parser.config.js
```diff
@@ -1,6 +1,7 @@
 const fs = require('fs');
 
-const localesEntries = fs.readdirSync('./public/locales');
+// @TODO: make it work with all packages and the web platform
```
AntonLantukh, line 3: Should we have a list of TODOs in issues?

packages/common/.depcheckrc.yaml
```diff
@@ -0,0 +1,8 @@
+ignores: [
+  # Installed by eslint-config-jwp
+  '@typescript-eslint/parser',
+  '@typescript-eslint/eslint-plugin',
+  'eslint-plugin-import',
+
+  ''
```
AntonLantukh:
```suggestion
```

packages/common/src/services/epg/ViewNexaEpgService.ts
```diff
@@ -2,11 +2,10 @@ import { object, string } from 'yup';
 import { parse } from 'date-fns';
 import { injectable } from 'inversify';
 
-import EpgService from './epg.service';
-
-import type { PlaylistItem } from '#types/playlist';
-import { logDev } from '#src/utils/common';
-import type { EpgProgram } from '#types/epg';
+import EpgService from '../EpgService';
+import type { PlaylistItem } from '../../../types/playlist';
```
AntonLantukh, line 6: Can we use import aliases form the common folder like `#common/types...`?

packages/common/src/services/integrations/AccountService.ts
```diff
@@ -31,18 +31,21 @@ export type AccountServiceFeatures = {
   readonly canUpdatePaymentMethod: boolean;
   readonly canShowReceipts: boolean;
   readonly hasSocialURLs: boolean;
-  readonly hasProfiles: boolean;
```
AntonLantukh, line 34: Do you think we can remove it? We don't have profiles for Cleeng.

packages/common/src/services/integrations/cleeng/CleengAccountService.ts
```diff
@@ -67,16 +76,29 @@ export default class CleengAccountService extends AccountService {
...
-  private getLocales: GetLocales = async (sandbox) => {
-    return this.cleengService.getLocales(sandbox);
+  private getLocales: GetLocales = async () => {
+    return this.cleengService.getLocales();
   };
 
-  initialize = async (config: Config, logoutCallback: () => Promise<void>) => {
-    await this.cleengService.initialize(!!config.integrations.cleeng?.useSandbox, logoutCallback);
+  initialize = async (config: Config, _url: string, logoutCallback: () => Promise<void>) => {
+    const cleengConfig = config?.integrations?.cleeng;
```
AntonLantukh, line 88: Should we add a Config dependency?

packages/common/src/stores/AppController.ts
```diff
@@ -0,0 +1,120 @@
...
+import type { CalculateIntegrationType } from '../../types/calculate-integration-type';
+import { DETERMINE_INTEGRATION_TYPE } from '../modules/types';
+
+import AccountController from './AccountController';
+import WatchHistoryController from './WatchHistoryController';
+import FavoritesController from './FavoritesController';
+import { useConfigStore } from './ConfigStore';
+
+@injectable()
+export default class AppController {
```
AntonLantukh, line 19: Should we move controllers to the separate folder now?

packages/common/src/stores/ProfileController.ts
```diff
@@ -27,45 +27,48 @@ const profileSchema = yup.object().shape({
...
   listProfiles = async () => {
     assertModuleMethod(this.profileService?.listProfiles, 'listProfiles is not available in profile service');
 
-    const res = await this.profileService.listProfiles(undefined, this.getSandbox());
-
-    const { canManageProfiles } = useProfileStore.getState();
-
-    if (!canManageProfiles && res?.responseData.canManageProfiles) {
-      useProfileStore.setState({ canManageProfiles: true });
-    }
```
AntonLantukh, line 48: This one is tricky. There can be JWP account service with profiles enabled, however request to the profiles endpoint could return 403 or 400 meaning that the feature is not enabled.

packages/eslint-config-jwp/.depcheckrc.yaml
```diff
@@ -0,0 +1,8 @@
+ignores: [
```
AntonLantukh: Do you think we could store linting configs somewhere in packages/configs?

packages/common/src/modules/register.ts
```diff
@@ -0,0 +1,87 @@
...
+
+// EPG services
+container.bind(EpgService).to(JWEpgService).whenTargetNamed(EPG_TYPE.jwp);
+container.bind(EpgService).to(ViewNexaEpgService).whenTargetNamed(EPG_TYPE.viewNexa);
+
+// Functions
+container.bind(INTEGRATION_TYPE).toDynamicValue(getIntegrationType);
+
+// Cleeng integration
+container.bind(DETERMINE_INTEGRATION_TYPE).toConstantValue(isCleengIntegrationType);
```
AntonLantukh, line 76: Maybe something like `INTEGRATION_TYPE_CONFIGURATIONS` ?

i18next-parser.config.js
ChristiaanScheermeijer: Or generate a TODO report: https://www.npmjs.com/package/leasot

packages/common/src/services/epg/ViewNexaEpgService.ts
ChristiaanScheermeijer: I haven't configured aliases yet. But we can look into this later. This will change (and break) all open PRs that are in progress.

packages/common/src/services/integrations/AccountService.ts
ChristiaanScheermeijer: `ProfileService` is named and only registered for the JWP integration. So when using Cleeng, the service isn't registered. The ProfileController knows this and exposes an `isEnabled` method to prevent unnecessary calls and waiting for disabled queries.

packages/common/src/services/integrations/cleeng/CleengAccountService.ts
ChristiaanScheermeijer: What do you mean exactly?

packages/common/src/stores/AppController.ts
ChristiaanScheermeijer: Yes,  that would be nice, but I also suggest doing this later. This will change many imports and break open PRs.

packages/common/src/stores/ProfileController.ts
ChristiaanScheermeijer: This was redundant because the `useProfiles` hook exposes the `canManageProfiles` if the endpoint returns an error.

packages/common/src/services/integrations/cleeng/CleengAccountService.ts
AntonLantukh: Never mind, was thinking about store being accessible as a dependency.

packages/eslint-config-jwp/.depcheckrc.yaml
ChristiaanScheermeijer: Can you elaborate?

packages/common/src/stores/ProfileController.ts
AntonLantukh: What if we we open the home page, go to the account page and then back to the home page? Won't profiles be fetched several times in this case?

packages/eslint-config-jwp/.depcheckrc.yaml
AntonLantukh: I mean having one folder `configs` in the packages with `eslint`, `postcss` and other linting configs inside.

```
/packages/configs
    /eslint-config-jwp
    /postcss-config-jwp
    /...
```

Or maybe on the top level:

```
/configs
    /eslint-config-jwp
    /postcss-config-jwp
    /...
```

packages/common/src/modules/register.ts
ChristiaanScheermeijer: Doesn't that imply it has something to do with the integration configuration? We have coined `INTEGRATION_TYPE` for the type. This is the function that returns that type if it should use that integration. What do you think?

AntonLantukh: Well, yeah. Also makes sense to me. @dbudzins any other suggestions?

AntonLantukh (approved):

packages/eslint-config-jwp/.depcheckrc.yaml
ChristiaanScheermeijer: The second suggestion is easier because of the workspaces configuration in the package.json.

packages/common/src/modules/register.ts
ChristiaanScheermeijer: It's a minor change, we can update this later (but before the next release).

Issue #38, opened 2022-04-20
Add i18next scanner


budzworthy:
It seems like i18n has been left in a half finished state. There are resource files and the translation object is used in most places, but there’s no pipeline to extract and maintain the resources. To achieve this add the scanner to the package, update the yarn commands, and add documentation:
https://github.com/i18next/i18next-scanner

ChristiaanScheermeijer:
I think this is already implemented 😅 

https://github.com/jwplayer/ott-web-app/blob/9337f451a6020bb99d41808c46bcf03e7faa9ed3/package.json#L15

You can run the following command to extract all translations from code. 

```shell
$ yarn i18next
```

There are some quirks though when dealing with dynamic translations like `` t(`key_${variable}`) ``, you will have to define the possible translations in comments like so:

https://github.com/jwplayer/ott-web-app/blob/9337f451a6020bb99d41808c46bcf03e7faa9ed3/src/components/CheckoutForm/CheckoutForm.tsx#L59-L62

PR #441, opened 2024-01-24
Fix / Prevent double id attribute

langemike:
Two e2e tests  failed because all the read-only`<input>`s on the page had the same ID. Causing Playwright to fail.

Happened because of this call:
```js
cardInfo.forEach(([label, value]) => I.seeInField(label, value));
```
Applying a name fixed this, but I  also marked the `name` property required, to prevent this issue from happening in the future. Double IDs can also lead to issues with clicking on a `<label>` which is linked to the wrong `<input>`.

Caused by: PR #38


langemike:
Mistake

PR #444, opened 2024-01-30
Feat / Accessible sidebar

langemike:
WIP

This change will make it possible to use the sidebar correctly using keyboard and screen reader navigation. It uses `inert` primarily to achieve this. I reordered the HTML and encapsulated all the main elements (header, main-content, footer) in one container, so I only need to apply `inert` to one element instead of multiple seperatly. This change didn't impact the presentation (styling) surprisingly enough.

_More changes are coming..._

TODO:
- [ ] Fix scrolling issue in sidebar on small viewports
- [ ] Apply <main> element (since this PR already includes a restructure, this should be little effort)


langemike:
Mistake. I close my PR.

PR #450, opened 2024-02-08
Feat/small features and refactoring

ChristiaanScheermeijer:
## Description

**Features**

- Dynamic GTM tag injection via an environment variable
- Custom footer text using markdown via an environment variable
- Initiative for replaceable components
- Automatic favicon/PWA image generator using a single base icon
- Dynamic PWA manifest generation

**Refactoring**

- Make routing paths static
- Typing refactoring (.d.ts -> .ts), which revealed some typing problems
- Update the reload subscriptions mechanism
- Move controllers to its own directory
- Move logic from containers/hooks to controllers (progress has been made, but more is planned)

**Fixes**

- Prevent a crash when history or favorites fails to load (duplicate IDs)
- Inline player not supporting TVOD (will always show the paywall)
- TVOD subscriptions not reloaded on non-SVOD platforms
- Customer IP not defined when creating a Cleeng order
- Error when deleting an InPlayer account
- Unexpected errors (crashes) not showing the ErrorPage
- Start watching button shown for TVOD items in AVOD platform (no integration configured)


packages/common/src/FormValidationError.ts
```diff
@@ -0,0 +1,11 @@
+type FormValidationErrors = Record<string, string[]>;
+
+export class FormValidationError extends Error {
+  public errors: FormValidationErrors;
+
+  constructor(errors: FormValidationErrors) {
+    super(Object.values(errors)[0][0]);
```
budzworthy: Would it be better to do a big csv join of all errors here? Otherwise either something dynamic or static. I worry about the edge case where there's an empty `errors` passed in and the indexing `Object.values(errors)[0][0]` null refs.

packages/common/src/controllers/AccountController.ts
```diff
@@ -188,20 +166,27 @@ export default class AccountController {
...
+        if (error.message.toLowerCase().includes('invalid param email')) {
+          throw new FormValidationError({ email: [i18next.t('account:login.wrong_email')] });
+        } else {
+          throw new FormValidationError({ form: [i18next.t('account:login.wrong_combination')] });
+        }
+      }
     }
 
-    useAccountStore.setState({ loading: false });
+    return useAccountStore.setState({ loading: false });
```
budzworthy: Does this need to be in a finally block so that we don't get stuck in 'loading'?

packages/common/src/controllers/AccountController.ts
```diff
@@ -188,20 +166,27 @@ export default class AccountController {
...
-
     useAccountStore.setState({ loading: true });
 
-    const response = await this.accountService.login({ config, email, password, referrer });
+    try {
+      const response = await this.accountService.login({ email, password, referrer });
 
-    if (response) {
-      await this.afterLogin(response.user, response.customerConsents, accessModel);
+      if (!response) throw new Error("Couldn't login");
```
budzworthy: Can we refactor this to avoid throwing errors to control flow? Maybe leave the `if (response)` check and move the return statement up to the end of the try block. Then let the catch block catch and rethrow for the known special cases, and finally let the flow cascade down to the end of the method where you can throw the `wrong_combination` error if nothing else has happened.

packages/common/src/controllers/AccountController.ts
```diff
@@ -212,38 +197,48 @@ export default class AccountController {
...
+          throw new FormValidationError({ password: [i18next.t('account:registration.invalid_password')] });
+        } else {
+          // in case the endpoint fails
+          throw new FormValidationError({ password: [i18next.t('account:registration.failed_to_create')] });
+        }
+      }
     }
 
-    await this.updatePersonalShelves();
+    return;
```
budzworthy: What happens when there's no response, but also no error? Is this handled downstream or should we try to refactor like login above?

packages/common/src/controllers/AccountController.ts
```diff
@@ -258,86 +253,68 @@ export default class AccountController {
...
     const { config } = useConfigStore.getState();
 
-    const response = await this.accountService.getPublisherConsents(config);
-
-    useAccountStore.setState({ publisherConsents: response.consents });
+    useAccountStore.setState({ loading: true });
+    const consents = await this.accountService.getPublisherConsents(config);
 
-    return response;
+    useAccountStore.setState({ publisherConsents: consents, loading: false });
```
budzworthy: Why the loading false here? Seems out of place...

packages/common/src/controllers/AccountController.ts
```diff
@@ -409,22 +386,30 @@ export default class AccountController {
...
+          this.reloadSubscriptions().finally(resolve);
         }, delay);
       });
     }
 
+    // For non-SVOD platforms, there could be TVOD items, so we only reload entitlements
+    if (accessModel !== 'SVOD') {
+      await this.refreshEntitlements?.();
+
+      return useAccountStore.setState({ loading: true });
```
budzworthy: This also looks out of place. If it is supposed to be here, please add a comment as to why.

packages/common/src/controllers/AccountController.ts
```diff
@@ -550,9 +530,9 @@ export default class AccountController {
       selectingProfileAvatar: null,
     });
 
-    this.profileController?.unpersistProfile();
+    this.profileController.unpersistProfile();
 
-    await this.favoritesController?.restoreFavorites();
-    await this.watchHistoryController?.restoreWatchHistory();
+    await this.favoritesController.restoreFavorites().catch(logDev);
```
budzworthy: Is there anywhere where we want these methods to throw errors? If not, does it make sense to add try/catch inside of the restoreFavorites and restoreWatchHistory implementations instead of catching on each call?

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -76,19 +79,30 @@ export default class CheckoutController {
...
+    try {
+      const response = await this.checkoutService.updateOrder(updateOrderPayload);
+
+      if (response.errors.length > 0) {
+        // clear the order when the order doesn't exist on the server
+        if (response.errors[0].includes(`Order with ${order.id} not found`)) {
+          useCheckoutStore.getState().setOrder(null);
+        }
+
+        throw new FormValidationError({ order: [response.errors[0]] });
```
budzworthy: You have space, why not keep all the errors?
```suggestion
        throw new FormValidationError({ order: response.errors });
```

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -76,19 +79,30 @@ export default class CheckoutController {
...
+        useCheckoutStore.getState().setOrder(response.responseData?.order);
+      }
+    } catch (error: unknown) {
+      if (error instanceof FormValidationError) {
+        throw error;
+      }
 
-    if (response.responseData?.order) {
-      useCheckoutStore.getState().setOrder(response.responseData?.order);
+      throw new FormValidationError({ couponCode: [i18next.t('account:checkout.coupon_not_valid')] });
```
budzworthy: Is coupon code invalid really the right 'catch all' error to show?

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -76,19 +79,30 @@ export default class CheckoutController {
...
+        throw new FormValidationError({ order: [response.errors[0]] });
       }
 
-      throw new Error(response.errors[0]);
-    }
+      if (response.responseData?.order) {
+        useCheckoutStore.getState().setOrder(response.responseData?.order);
+      }
+    } catch (error: unknown) {
+      if (error instanceof FormValidationError) {
```
budzworthy: Suggest trying to refactor this again to avoid the awkward throw-catch-rethrow & the empty return at the end of the method.

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -174,18 +191,26 @@ export default class CheckoutController {
...
+    cancelUrl,
+    errorUrl,
+    couponCode = '',
+  }: {
+    successUrl: string;
+    waitingUrl: string;
+    cancelUrl: string;
+    errorUrl: string;
+    couponCode: string;
+  }): Promise<string> => {
```
budzworthy: If we remove the response type, can we at least create an object type for the response here (i.e. `{ redirectUrl: string})`? Returning a raw string is a little hard to interpret for callers of this method.

packages/common/src/controllers/EpgController.test.ts
budzworthy, line 1: What do you think about organizing the modules by concern instead of by architecture type?

Something like:
/src
  /modules
    /Account
      /AccountController.ts
      /AccountStore.ts
      /JwpAccountService.ts
      /CleengAccountService.ts

packages/common/src/paths.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+// Get a nested path without the parent prefix (for nested routes)
+// I.E.: getNestedPath('/u/*', '/u/my-account') => 'my-account'
+const getNestedPath = (parentPath: string, fullPath: string) => fullPath.replace(parentPath.replace('*', ''), '');
+
+export const NESTED_PATH_USER_ACCOUNT = getNestedPath(PATH_USER, PATH_USER_ACCOUNT);
+export const NESTED_PATH_USER_MY_PROFILE = getNestedPath(PATH_USER, PATH_USER_MY_PROFILE);
+export const NESTED_PATH_USER_FAVORITES = getNestedPath(PATH_USER, PATH_USER_FAVORITES);
+export const NESTED_PATH_USER_PAYMENTS = getNestedPath(PATH_USER, PATH_USER_PAYMENTS);
+export const NESTED_PATH_USER_PROFILES = getNestedPath(PATH_USER, PATH_USER_PROFILES);
```
budzworthy:
```suggestion
export const RELATIVE_PATH_USER_ACCOUNT = 'my-account';
export const RELATIVE_PATH_USER_MY_PROFILE = 'my-profile/:id';
export const RELATIVE_PATH_USER_FAVORITES = 'favorites';
export const RELATIVE_PATH_USER_PAYMENTS = 'payments';
export const RELATIVE_PATH_USER_PROFILES = 'profiles';
```

packages/common/src/paths.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const PATH_MEDIA = '/m/:id/*';
+export const PATH_PLAYLIST = '/p/:id/*';
+export const PATH_LEGACY_SERIES = '/s/:id/*';
+
+export const PATH_SEARCH = '/q/*';
+export const PATH_ABOUT = '/o/about';
+
+export const PATH_USER = '/u/*';
+export const PATH_USER_ACCOUNT = '/u/my-account';
```
budzworthy: I think building paths is a little easier to follow than splitting paths
```suggestion
export const PATH_USER_BASE = '/u'
export const PATH_USER = `${PATH_USER_BASE}/*`;
export const PATH_USER_ACCOUNT = `${PATH_USER_BASE}/${RELATIVE_PATH_USER_ACCOUNT}`;
```

packages/common/src/services/ConfigService.ts
```diff
@@ -28,9 +28,7 @@ export default class ConfigService {
     content: [],
     menu: [],
     integrations: {},
-    styling: {
-      footerText: '',
```
budzworthy: We need to keep footerText from here. It is still in app config, just not exposed in the JW dashboard. We can make the env values override it if you want.

packages/common/src/utils/configSchema.ts
```diff
@@ -41,7 +41,6 @@ const stylingSchema: SchemaOf<Styling> = object({
   backgroundColor: string().nullable(),
   highlightColor: string().nullable(),
   headerBackground: string().nullable(),
-  footerText: string().nullable(),
```
budzworthy: Keep

packages/common/types/config.ts
```diff
@@ -59,7 +59,6 @@ export type Styling = {
   backgroundColor?: string | null;
   highlightColor?: string | null;
   headerBackground?: string | null;
-  footerText?: string | null;
```
budzworthy, line 62: Keep

packages/common/src/FormValidationError.ts
ChristiaanScheermeijer: Yes, good point. We removed the `string[]` type of the errors object in a different PR, but currently, this could fail indeed.

https://github.com/Videodock/ott-web-app/pull/86/files#diff-9caec313adc28576ac57907ce90dd950c36466ed49cce2f3d8179edb284e5b18

packages/ui-react/src/containers/Layout/Layout.tsx
```diff
@@ -36,23 +39,23 @@ const Layout = () => {
...
   const favoritesEnabled = !!config.features?.favoritesList;
-  const { menu, assets, siteName, description, styling, features } = config;
+  const { menu, assets, siteName, description, features } = config;
   const metaDescription = description || t('default_description');
+  const footerText = unicodeToChar(env.APP_FOOTER_TEXT);
 
   const profileController = getModule(ProfileController, false);
 
   const { searchPlaylist } = features || {};
-  const { footerText } = styling || {};
```
budzworthy, line 45: Keep

packages/common/src/services/FavoriteService.ts
```diff
@@ -15,31 +25,60 @@ export default class FavoriteService {
 
   private readonly apiService;
   private readonly storageService;
+  private readonly accountService;
 
-  constructor(apiService: ApiService, storageService: StorageService) {
+  constructor(@inject(INTEGRATION_TYPE) integrationType: string, apiService: ApiService, storageService: StorageService) {
     this.apiService = apiService;
     this.storageService = storageService;
+    this.accountService = getNamedModule(AccountService, integrationType, false);
```
budzworthy, line 34: Does normal injection not work for this service? It's a shame having to do all the manual handling of integration type here.

packages/common/src/services/WatchHistoryService.ts
```diff
@@ -1,23 +1,36 @@
...
-  private readonly storageService: StorageService;
+  private readonly apiService;
+  private readonly storageService;
+  private readonly accountService;
 
-  constructor(apiService: ApiService, storageService: StorageService) {
+  constructor(@inject(INTEGRATION_TYPE) integrationType: string, apiService: ApiService, storageService: StorageService) {
     this.apiService = apiService;
     this.storageService = storageService;
+    this.accountService = getNamedModule(AccountService, integrationType);
```
budzworthy, line 34: Same question here, does normal injection not work for this service?

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: That would look something like this?

- Only rethrow known errors (wrong_email)
- Fallback to wrong_combination for all unexpected errors
- Use finally to always clear the loading state (this also works with the `return` in the `try` block

```ts
  login = async (email: string, password: string, referrer: string) => {
    useAccountStore.setState({ loading: true });

    try {
      const response = await this.accountService.login({ email, password, referrer });
      
      if (response) {
        await this.afterLogin(response.user, response.customerConsents);
        await this.favoritesController?.restoreFavorites().catch(logDev);
        await this.watchHistoryController?.restoreWatchHistory().catch(logDev);
        
        return;
      }
    } catch (error: unknown) {
      if (error instanceof Error && error.message.toLowerCase().includes('invalid param email')) {
        throw new FormValidationError({ email: [i18next.t('account:login.wrong_email')] });
      }
    } finally {
      useAccountStore.setState({ loading: false });
    }

    throw new FormValidationError({ form: [i18next.t('account:login.wrong_combination')] });
  };
```

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: See https://github.com/jwplayer/ott-web-app/pull/450/files/b700fbb7ea31cde67740bab99246a0b705058c55#r1484537050

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: I'm actually not sure why there even is a condition for `if (response)`. All service methods should return a response or throw (which is the case for login and register). I think we can remove this condition entirely.

ChristiaanScheermeijer: Good catch btw 😄

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: Yes, we've made the AccountStore the source of truth for consents, because the consents were stored in the AccountStore but also used as useQuery result.

But I agree that the loading state is now "misused" even more. We should add separate `loading` states for these cases in the AccountStore.

We'll look into this.

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: 👀

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: We had a single case where this threw an error (duplicate serie IDs). The controller in concert should indeed decide whether errors are thrown or not. I will update the `restoreFavorites` and `restoreWatchHistory` methods instead.

packages/common/src/controllers/CheckoutController.ts
ChristiaanScheermeijer: We only show the first error on all form fields and want to remove the `string[]` type. But if not, we could pass all errors indeed.

packages/common/src/controllers/CheckoutController.ts
ChristiaanScheermeijer: I will look into this

packages/common/src/services/integrations/cleeng/formatters/consents.ts
budzworthy, line 1: What's the reason to put this in its own file instead of just leaving it in CleengAccountService?

packages/common/src/controllers/CheckoutController.ts
ChristiaanScheermeijer: It's a bit weird, but yes: 

https://github.com/jwplayer/ott-web-app/blob/3e3e2b14c4926e596fad46c2ea7cc99dbced05f0/packages/common/src/services/integrations/jwp/JWPCheckoutService.ts#L200

Although, we don't catch Cleeng's "invalid coupon" errors anymore. This is fixed.

packages/common/src/controllers/EpgController.test.ts
ChristiaanScheermeijer: I do like the idea (for the @jwp/common package at least), but let's do that sometime later 😄

packages/common/src/services/FavoriteService.ts
ChristiaanScheermeijer: Unfortunately, all integration services injections require the integration type. Only within the integration we can do normal injection because the integration type is known there.

packages/common/src/services/integrations/cleeng/formatters/consents.ts
ChristiaanScheermeijer: Same as for utils, these are easier to test and validate. There could be 10~20 formatters in the near future which would clutter the service too much IMO

ChristiaanScheermeijer:
Thanks for reviewing this thoroughly @dbudzins, I've implemented most of the requested changes. Let me know what you think!

Have a good weekend 😄

docs/configuration.md
```diff
@@ -160,8 +160,7 @@ Use the `styling` object to define extra styles for your application.
   "styling": {
     "backgroundColor": null,
     "highlightColor": null,
-    "headerBackground": null,
-    "footerText": "Blender Foundation"
```
AntonLantukh, line 164: Do you think we should remove it?

packages/common/src/services/FavoriteService.ts
budzworthy: Maybe we should reconsider (in a future PR) how we're handling that then. I would like to minimize this boilerplate.

packages/common/src/paths.tsx
```diff
@@ -0,0 +1,26 @@
+export const PATH_HOME = '/';
+
+export const PATH_MEDIA = '/m/:id/*';
+export const PATH_PLAYLIST = '/p/:id/*';
+export const PATH_LEGACY_SERIES = '/s/:id/*';
+
+export const PATH_SEARCH = '/q/*';
+export const PATH_ABOUT = '/o/about';
+
+export const PATH_USER_BASE = '/u';
```
AntonLantukh, line 10: Maybe we could use only one constant for main routes?

```
const PATH = {
    HOME: '/',
    ....
}
```

packages/common/src/paths.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const PATH_USER = `${PATH_USER_BASE}/*`;
+
+export const RELATIVE_PATH_USER_ACCOUNT = 'my-account';
+export const RELATIVE_PATH_USER_MY_PROFILE = 'my-profile/:id';
+export const RELATIVE_PATH_USER_FAVORITES = 'favorites';
+export const RELATIVE_PATH_USER_PAYMENTS = 'payments';
+export const RELATIVE_PATH_USER_PROFILES = 'profiles';
```
AntonLantukh, line 17: Same for user:

```
const RELATIVE_PATH_USER = {
    ACCOUNT: 'my-account',
    ...
}
```

platforms/web/tsconfig.json
```diff
@@ -22,6 +22,7 @@
     "scripts",
     "test",
     "vite.config.ts",
-    "../../packages/common/types"
+    "../../packages/common/types",
```
AntonLantukh, line 25: I would prefer importing them from the package instead of using a relative import here. Do you think we can do it?

platforms/web/src/containers/Root/Root.tsx
```diff
@@ -44,9 +52,7 @@ const DemoContentLoader = ({ query }: { query: BootstrapData }) => {
   return (
     <>
       {/* Show the error page when error except in demo mode (the demo mode shows its own error) */}
-      {!IS_DEMO_OR_PREVIEW && error && (
-        <ErrorPage title={error?.payload?.title} message={error?.payload?.description} error={error} helpLink={error?.payload?.helpLink} />
-      )}
+      {!IS_DEMO_OR_PREVIEW && error && renderError(error)}
```
AntonLantukh: I would prefer to use a component here to make it consistent.

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
```diff
@@ -49,26 +61,31 @@ export default function AppRoutes() {
   }
 
   const shouldManageProfiles =
-    !!user && profilesEnabled && !profile && (accessModel === 'SVOD' || accessModel === 'AUTHVOD') && !userModal && !location.pathname.includes('/u/profiles');
+    !!user &&
+    profilesEnabled &&
+    !profile &&
+    (accessModel === 'SVOD' || accessModel === 'AUTHVOD') &&
+    !userModal &&
+    !location.pathname.includes(PATH_USER_PROFILES);
```
AntonLantukh, line 69: Wow, so many things to consider.

packages/common/src/services/FavoriteService.ts
ChristiaanScheermeijer: I tried this, but it does change the containers a bit:

```tsx
// Functions
container.bind(INTEGRATION_TYPE).toDynamicValue(getIntegrationType);

const getDynamicIntegration =
  <T>(identifier: interfaces.ServiceIdentifier<T>) =>
  (context: interfaces.Context) => {
    const integrationType = context.container.get(INTEGRATION_TYPE) as string;
    return context.container.getNamed<T>(identifier, integrationType);
  };

// Integration getters
container.bind(AccountService).toDynamicValue(getDynamicIntegration('AccountService'));
container.bind(CheckoutService).toDynamicValue(getDynamicIntegration('CheckoutService'));
container.bind(SubscriptionService).toDynamicValue(getDynamicIntegration('SubscriptionService'));
container.bind(ProfileService).toDynamicValue(getDynamicIntegration('ProfileService'));

// Cleeng integration implementation
container.bind(DETERMINE_INTEGRATION_TYPE).toConstantValue(isCleengIntegrationType);
container.bind(CleengService).toSelf();
container.bind<AccountService>('AccountService').to(CleengAccountService).whenTargetNamed(INTEGRATION.CLEENG);
container.bind<CheckoutService>('CheckoutService').to(CleengCheckoutService).whenTargetNamed(INTEGRATION.CLEENG);
container.bind<SubscriptionService>('SubscriptionService').to(CleengSubscriptionService).whenTargetNamed(INTEGRATION.CLEENG);

// JWP integration implementation
container.bind(DETERMINE_INTEGRATION_TYPE).toConstantValue(isJwpIntegrationType);
container.bind(JWPEntitlementService).toSelf();
container.bind<AccountService>('AccountService').to(JWPAccountService).whenTargetNamed(INTEGRATION.JWP);
container.bind<CheckoutService>('CheckoutService').to(JWPCheckoutService).whenTargetNamed(INTEGRATION.JWP);
container.bind<SubscriptionService>('SubscriptionService').to(JWPSubscriptionService).whenTargetNamed(INTEGRATION.JWP);
container.bind<ProfileService>('ProfileService').to(JWPProfileService).whenTargetNamed(INTEGRATION.JWP);
```

With the above, we can return back to using inversify injection:

```tsx
export default AccountController {
  constructor(
    accountService: AccountService,
    checkoutService: CheckoutService,
    subscriptionService: SubscriptionService,
    favoritesController: FavoritesController,
    watchHistoryController: WatchHistoryController,
    profileController: ProfileController,
  ) {
    this.accountService = accountService;
    this.checkoutService = checkoutService;
    this.subscriptionService = subscriptionService;
    this.favoritesController = favoritesController;
    this.watchHistoryController = watchHistoryController;
    this.profileController = profileController;

    this.features = accountService ? this.accountService.features : DEFAULT_FEATURES;
  }
}
```

How does this work?

We only register the integration services once on the abstract classes to dynamic values. These will be resolved once requested. In each dynamic value, we request the implementation based on the integration type returned by the `INTEGRATION_TYPE` value. 

The only downside is that you must register the implementations with a different service identifier (plain string in the example above), but it will greatly simplify the usage in controllers 😄.

It essentially moves the `getNamedModule` code to the service container.

platforms/web/index.html
```diff
@@ -2,16 +2,30 @@
...
+  <link rel="icon" href="/images/icons/favicon.ico" type="image/x-icon" sizes="any">
+  <link rel="icon" type="image/png" sizes="196x196" href="/images/icons/favicon-32x32.png">
+  <link rel="icon" type="image/png" sizes="160x160" href="/images/icons/favicon-64x64.png">
+  <link rel="icon" type="image/png" sizes="96x96" href="/images/icons/favicon-96x96.png">
+  <link rel="icon" type="image/png" sizes="64x64" href="/images/icons/favicon-64x64.png">
+  <link rel="icon" type="image/png" sizes="32x32" href="/images/icons/favicon-32x32.png">
+  <link rel="icon" type="image/png" sizes="16x16" href="/images/icons/favicon-16x16.png">
+  <link rel="apple-touch-icon" href="/images/icons/apple-touch-icon-180x180.png">
+  <link rel="apple-touch-icon" href="/images/icons/apple-touch-icon-152x152.png">
+  <link rel="apple-touch-icon" href="/images/icons/apple-touch-icon-144x144.png">
```
AntonLantukh, line 23: Do we need these icons in case we already generate them automatically using assets-generator?

AntonLantukh: Can they also be auto-injected?

packages/ui-react/src/modules/container.ts
```diff
@@ -0,0 +1,5 @@
+import { Container } from 'inversify';
+
+const container = new Container({ defaultScope: 'Singleton', skipBaseClassChecks: true });
```
AntonLantukh, line 3: Should it be part of react or maybe common part?

packages/ui-react/src/containers/StartWatchingButton/StartWatchingButton.tsx
```diff
@@ -70,6 +73,11 @@ const StartWatchingButton: React.VFC<Props> = ({ item, playUrl, disabled = false
     return () => setRequestedMediaOffers(null);
   }, [mediaOffers, setRequestedMediaOffers]);
 
+  // the user can't purchase access in an AVOD platform due to missing configuration, so we hide the button
+  if (accessModel === ACCESS_MODEL.AVOD && !isEntitled) {
```
AntonLantukh, line 77: Hm, is `AVOD` model itself not enough to hide the button? Looks like we don't have any account for AVOD to check the access in case it is TVOD.

platforms/web/index.html
ChristiaanScheermeijer: @langemike do you know if auto-injecting the icons is possible as well?

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
ChristiaanScheermeijer: Yes... 😅 But the condition hasn't changed; it's only formatted by Prettier due to the `PATH_USER_PROFILES` usage.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -223,6 +93,19 @@ const Checkout = () => {
     );
   }
 
+  const cancelUrl = createURL(window.location.href, { u: 'payment-cancelled' });
+  const waitingUrl = createURL(window.location.href, { u: 'waiting-for-payment' });
+  const errorUrl = createURL(window.location.href, { u: 'payment-error' });
```
AntonLantukh, line 98: Should we also move them to urls constants?

platforms/web/src/containers/Root/Root.tsx
ChristiaanScheermeijer: Do you suggest creating another error component for handling `Error | AppError`, or to do that inline (2x)?

Or change `renderError` to `<BootstrapError error={error} />` and define `BootstrapError` in `Root.tsx`?

packages/hooks-react/src/useCheckout.ts
```diff
@@ -0,0 +1,98 @@
...
+    mutationKey: ['submitPaymentStripe'],
+    mutationFn: checkoutController.directPostCardPayment,
+    onSuccess: onSubmitStripePaymentSuccess,
+  });
+
+  useEffect(() => {
+    if (offer && !order && !createOrder.isLoading) {
+      createOrder.mutate({ offer });
+    }
+  }, [offer, order, createOrder]);
```
AntonLantukh, line 75: Should we probably do it from the outside and just keep hooks + state here?

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -46,15 +46,17 @@ export default class CheckoutController {
     return this.accountService.svodOfferIds;
   };
 
-  createOrder = async (offer: Offer, paymentMethodId?: number): Promise<void> => {
+  createOrder = async (offer: Offer): Promise<void> => {
     const { getAccountInfo } = useAccountStore.getState();
     const { customer } = getAccountInfo();
 
+    const paymentMethods = await this.getPaymentMethods();
+    const paymentMethodId = paymentMethods[0]?.id;
```
AntonLantukh, line 54: Won't it be a problem in case several methods are available?

packages/common/src/controllers/AccountController.ts
```diff
@@ -103,26 +101,6 @@ export default class AccountController {
     return this.accountService.sandbox;
   }
 
-  updatePersonalShelves = async () => {
```
AntonLantukh, line 106: Hm, how do we update shelves now after the registration?

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -70,23 +72,31 @@ export default class CheckoutController {
...
-      couponCode,
-    };
+    let response;
+
+    try {
+      response = await this.checkoutService.updateOrder({ order, paymentMethodId, couponCode });
+    } catch (error: unknown) {
+      // TODO: we currently (falsely) assume that the only error caught is because the coupon is not valid, but there
+      //  could be a network failure as well (JWPCheckoutService)
+      throw new FormValidationError({ couponCode: [i18next.t('account:checkout.coupon_not_valid')] });
```
AntonLantukh, line 82: Should it be a form error in this case?

platforms/web/src/containers/Root/Root.tsx
AntonLantukh: I mean hanging it to `<BootstrapError error={error} /`, mostly just to have the same way conditions are built.

AntonLantukh:
@ChristiaanScheermeijer thank you! I still have some questions regarding continue watching / favorites logic. I will take one more look tomorrow.

packages/ui-react/src/modules/container.ts
ChristiaanScheermeijer: This is used for replaceable React components. It's not used yet and might change, but the common package shouldn't know about React components.

packages/ui-react/src/containers/StartWatchingButton/StartWatchingButton.tsx
ChristiaanScheermeijer: No, because we still want to show the button for AVOD (free) content. `isEntitled` is set to `true` when the user can watch the item for all access models.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
ChristiaanScheermeijer: Not sure, I think they depend on the current domain and pathname or do you mean just the `u` constants?

packages/hooks-react/src/useCheckout.ts
ChristiaanScheermeijer: @royschut what do you think? 

FYI we have this PR in review: https://github.com/Videodock/ott-web-app/pull/93

packages/common/src/controllers/CheckoutController.ts
ChristiaanScheermeijer: The logic hasn't changed and sets the initial payment method ID when initializing the order for the checkout form. When the user changes the payment method, the `updateOrder` method is called with the correct payment method id.

packages/common/src/controllers/CheckoutController.ts
ChristiaanScheermeijer: Although changed, it works exactly like before. The only error thrown by the JWP service is `Invalid coupon code`. We will have to improve that first before we can throw a more specific form error here.

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: That's a good question, I will test this tomorrow 😄

platforms/web/index.html
langemike: Unfortunately the asset-generator has a shortcomings to create the correct link to the icons, because of the monorepo structure. Though we can inject the tags based on our own defined settings. I created this PR for it https://github.com/Videodock/ott-web-app/pull/126

docs/configuration.md
ChristiaanScheermeijer: I think we shouldn't promote it if it isn't supported in the dashboard anymore. It is still working though

platforms/web/tsconfig.json
ChristiaanScheermeijer: That is possible, but not in `tsconfig#includes`. We will have to create a `types/index.d.ts`, import the global definition files, and update the package.json with `"types": "./types/index.d.ts",`. Only then can we add it to the `tsconfig#compilerOptions.types` property as `@jwp/ott-common` and `@jwp/ott-ui-react`.

However, I'm intending to remove this entirely. It's better to only import the required types from common where needed.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
AntonLantukh: Yeah, just `u` constant.

ChristiaanScheermeijer: Can we do that in a separate PR? Because these `u` constants are used a lot (~27 files and 82 occurrences 😅)

AntonLantukh: Of course!

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: I've moved the `restoreFavorites` and `restoreWatchHistory` calls to the `afterLogin` method. Before, the restore methods were called three times in a row when the session was restored. It still calls restore twice because the AppController and AccountController will call these methods. But I think this problem is already in the develop branch.

^ This also calls the restore methods after registration

platforms/web/src/containers/Root/Root.tsx
ChristiaanScheermeijer: Done!

packages/hooks-react/src/useCheckout.ts
royschut: The advantage of having this logic oriented effect here is that it can be used from other react platforms (such as react-native), keeping the Checkout container purely UI.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
royschut: In most cases we use this util, which is typed from `AccountModals`: 
```javascript 
const modalURLFromLocation = (location: Location, u: keyof AccountModals | null, queryParams?: QueryParamsArg) => string
```

However, in this case we need the full url (including hostname), because they're used as return url's, so I decided to use the `createURL` util.

We could create a wrapper util to make it more strict. What do you think: @AntonLantukh 

```javascript
const createAccountURL = (u: keyof AccountModals) => createURL(window.location.href, { u });
```

packages/common/src/controllers/AccountController.ts
AntonLantukh: Thank you. But do we need to call `restoreFavorites` and `restoreWatchHistory` separately in the `login` method (we also call `afterLogin`)?

AntonLantukh: It also looks like we lose watch history now after a customer completes the registration. The use case:

1. I watch some videos and add them to Favorites (I don't have any registration yet, it can be free content).
2. I decide to register and my watchhistory gets overwritten with the data from the api (empty list).

AntonLantukh: While registering, I also saw the the password field error was highlighted even though it matched the expected format. However I am not sure yet whether it has something to do with the changes in this PR.

platforms/web/tsconfig.json
AntonLantukh: Do you want to do it in this PR?

packages/hooks-react/src/useCheckout.ts
AntonLantukh: My concern was the mutation happens automatically when mounting, which can be kind of unexpected. But it is okay if you think you can reuse it in the future.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
AntonLantukh: Yeah, it could work. Thank you.

royschut: Perfect! I suggest we merge this first (as Christiaan suggested) while I'll implement it into our current working branch.

platforms/web/tsconfig.json
ChristiaanScheermeijer: I prefer to do it separately and finish this PR asap because there are many new PRs queued...

https://github.com/Videodock/ott-web-app/pulls

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: I fixed the restoration and personal shelves after registering.

I couldn't reproduce the password error 🤔 Do you have more information about this?

AntonLantukh: Okay, I think the problem is that the error for the password field is also shown when the account already exists (though the proposed password itself was okay).

Screenshot 2024-02-20 at 15 01 06

packages/common/src/paths.tsx
AntonLantukh: What do you think about using one constant for routes?

ChristiaanScheermeijer: We've discussed this and concluded that this is an improvement (although minor considering this PR). We have created a ticket to change this in the next sprint. Is that OK?

packages/common/src/controllers/AccountController.ts
ChristiaanScheermeijer: Ah yes, we've also found that (a11y) issue and fixed that here: https://github.com/Videodock/ott-web-app/pull/107

This will be submitted next week 😄

AntonLantukh:
@ChristiaanScheermeijer thank you for the changes!

AntonLantukh (approved):

ChristiaanScheermeijer:
Thanks for reviewing @dbudzins and @AntonLantukh 👍 🙌

ChristiaanScheermeijer:
@dbudzins @AntonLantukh, can we merge this when you're ready? (the next PRs are already on the way 😄)

Issue #121, opened 2022-07-20
Add a Proper Logging Framework


budzworthy:
We are mostly just using the `logDev` method everywhere. We should really add a logging framework with an adjustable log level.

ChristiaanScheermeijer:
@dbudzins @AntonLantukh @royschut @langemike @MelissaDTH 

In the next sprint, we will implement a logging application, and I think this might help with that.

We've used [Pino](https://github.com/pinojs/pino) before, but this is mainly created for NodeJS applications and might be overkill.

Now that we have services, I am considering creating a `LoggingService` with a simple interface. For example:

```ts
interface ILoggingService {
  debug(namespace: string, ...args: unknown[]): void;
  info(namespace: string, ...args: unknown[]): void;
  warn(namespace: string, ...args: unknown[]): void;
  error(namespace: string, error: Error, ...args: unknown[]): void;
  fatal(namespace: string, error: Error, ...args: unknown[]): void;
  
  // optionally implement support for (custom) transporters, for example:
  // consoleTransporter, sentryTransporter, cloudWatchTransporter, ... 
  registerTransporter(transporter: (level: string, args: unknown[], error?: Error) => void): void;
}
```

I see two potential ways of adding a logging implementation. 

**Allow for registering custom transporters (see above)**

- Pro: This allows for multiple transporters (easier)
- Pro: Less logic to re-implement (determine the log level and which logs to output)
- Con: More logic to implement initially

**Override the default LoggingService with a custom implementation** 

- Pro: Full control
- Pro: Easier to maintain and implement
- Con: Custom implementations need to implement more 

Although, a third solution is to override the existing LoggingService and only override a generic method that handles the actual logging. This might be even more simple:

```ts
class SentryLoggingService extends LoggingService {
  override log(level: string, namespace: string, args: unknown[], error?: Error) {
    // console logging
    super.log(level, namespace, args, error);

    // log to Sentry
    Sentry.captureMessage(`[${namespace}]: ${args.join(',')}`, level);
  }
}
```

Should we accept only one string and mark the rest as additional data (context) which may be useful debug data?

```ts
interface ILoggingService {
  debug(namespace: string, message: string, ...additionalData: unknown[]): void;
  // ...
}
```

What do you think?

budzworthy:
[...]

PR #117, opened 2022-07-18
fix(project): assets object not required anymore

AntonLantukh:
## Description

- empty logo url support

This PR solves #116 .

### Steps completed:

According to our definition of done, I have completed the following steps:

- [ ] User stories met
- [ ] Storybook stories
- [ ] Unit tests added
- [ ] Linting passing
- [ ] Unit tests passing
- [ ] Docs updated (including config and env variables)
- [ ] Translations added
- [ ] UX tested
- [ ] Browsers / platforms tested
- [ ] Rebased & ready to merge without conflicts
- [ ] Reviewed own code

Issue #116, opened 2022-07-18
Allow null banner image


budzworthy:
Do not validate the banner image. It should be allowed to be null.

PR #455, opened 2024-02-29
Feat / accessibility and bug fixes

ChristiaanScheermeijer:
This PR results from many features and fixes done on the [Videodock/ott-web-app](https://github.com/Videodock/ott-web-app) fork. 

We've mainly worked on accessibility, UI improvements and bug fixes.

From now on, we hope to submit smaller pull requests directly to the open-source repo. This should make reviewing more bearable for you guys 😅 

### Features

* **a11y:** add auto focus to play/pause button of players ([9a78103](https://github.com/Videodock/ott-web-app/commit/9a781038dd63ccc2d1d3c506fe96ade153ad953a))
* **a11y:** apply correct aria attributes and improve searchbar ([#117](https://github.com/Videodock/ott-web-app/issues/117)) ([52a620c](https://github.com/Videodock/ott-web-app/commit/52a620cd0d3fc30bbf510fe59eaca2f2d3378bc2))
* **a11y:** apply the autocomplete attribute to input fields([#115](https://github.com/Videodock/ott-web-app/issues/115)) ([ab4e6f5](https://github.com/Videodock/ott-web-app/commit/ab4e6f5aa8cbdcd18c347ee1b472e58c11dd443a))
* **a11y:** form validation improvements for screen readers ([4628f0b](https://github.com/Videodock/ott-web-app/commit/4628f0b9f6ae4b76bd6d9de279bfa7ecdce1753f))
* **a11y:** improve accessibility of sidebar ([c58b7cd](https://github.com/Videodock/ott-web-app/commit/c58b7cd807c07306c3cc0b8596ed464e0894f320))
* **a11y:** optimize profile menu ([1c5f78d](https://github.com/Videodock/ott-web-app/commit/1c5f78ddc8f88c5884203670c0790fc689b7780b))
* **a11y:** slide chevron buttons hitarea increased ([9494434](https://github.com/Videodock/ott-web-app/commit/9494434257cdbc79d3ce8425cc596f3f05ce19a1))
* **a11y:** slider navigation optimized ([c5ed8c0](https://github.com/Videodock/ott-web-app/commit/c5ed8c02de0277a2a0b91ef16e1e4c0ceed790b6))
* **a11y:** video meta data items separated ([#116](https://github.com/Videodock/ott-web-app/issues/116)) ([d72aa2b](https://github.com/Videodock/ott-web-app/commit/d72aa2b68908f7b5057eff05c2b21681b75fa3aa))
* align sidebar backdrop with modal ([1af94d1](https://github.com/Videodock/ott-web-app/commit/1af94d1d70ca14e361ceb34e3397d310fcbca321))
* Render favicons dynamically ([e9b4f4b](https://github.com/Videodock/ott-web-app/commit/e9b4f4ba81b7b1dc52fb7ef6dbfdc990b7eeb095))

### Bug Fixes

* **a11y:** add aria-pressed to password icon and underline to footer links ([e80a840](https://github.com/Videodock/ott-web-app/commit/e80a8400a1d3ce6aecdd914c7c57121d0c9fd8fe))
* **a11y:** add role to make lock icon readable ([549fb7f](https://github.com/Videodock/ott-web-app/commit/549fb7fcd6577eb0e21ffad8bb377631ae093602))
* **a11y:** add selected state to season filters ([1117332](https://github.com/Videodock/ott-web-app/commit/1117332fc804fa8383d5170110b3d2a85a82b327))
* **a11y:** apply aria-hidden to checkbox asterisk ([#121](https://github.com/Videodock/ott-web-app/issues/121)) ([5ada707](https://github.com/Videodock/ott-web-app/commit/5ada707fe292b2a00de615199388658a9c11e1a7))
* **a11y:** ensure correct focus on first sidebar menu item ([0f45a87](https://github.com/Videodock/ott-web-app/commit/0f45a87006a9afa50d4bf516dae40f7e8250f8df))
* **a11y:** ensure wcag compliance with 1.5 line-height for texts ([#109](https://github.com/Videodock/ott-web-app/issues/109)) ([f621ef7](https://github.com/Videodock/ott-web-app/commit/f621ef7e43f8e4f8c5fe4fec2206eb49aed54d60))
* **a11y:** improve screenreader focus on start watching button ([cdb20ee](https://github.com/Videodock/ott-web-app/commit/cdb20ee2df4e18f524c608d6a63062deaf313bd2))
* **a11y:** only set focus programmatically on keyboard navigation ([693656a](https://github.com/Videodock/ott-web-app/commit/693656a752c367018ad043da3d622fce45096c3c))
* **a11y:** optimize footer for accessibility ([060701d](https://github.com/Videodock/ott-web-app/commit/060701dc5407b1ac49b1b930e8c03b651b733e9b))
* **a11y:** roles and aria-attributes regarding modals ([#125](https://github.com/Videodock/ott-web-app/issues/125)) ([a37027a](https://github.com/Videodock/ott-web-app/commit/a37027ad45ae2c3e9f325bb92a9721dd5b166de7))
* **a11y:** usermenu item count & correct html structure ([244d68b](https://github.com/Videodock/ott-web-app/commit/244d68b55534e8a9e740b755174d81e83d183008))
* **a11y:** videoplayer focus (iOS) & restoration ([52aaf67](https://github.com/Videodock/ott-web-app/commit/52aaf67b5f2583ebe81edd72de07ddbd588c088a))
* active state of language menu and apply aria-current ([f866c9a](https://github.com/Videodock/ott-web-app/commit/f866c9a67d59a90fdd8fad1de2032e05c22bec17))
* broken checkbox in personal details form ([1adaae3](https://github.com/Videodock/ott-web-app/commit/1adaae3e01247c7a8ead0fe970d8e912ae840be3))
* cinema not covering full viewport height ([64a94a9](https://github.com/Videodock/ott-web-app/commit/64a94a985ab72f05a9f4546b384c371d9848a562))
* hide disabled questions in personal details screen ([8e3db3e](https://github.com/Videodock/ott-web-app/commit/8e3db3e01aca59c3f8c7356bfaa4bc04f854ab87))
* hide optional label in season select dropdown ([eac1809](https://github.com/Videodock/ott-web-app/commit/eac1809ce86ddb0bf679394281ed6f615ac8052b))
* improve consents required form error ([8289ba1](https://github.com/Videodock/ott-web-app/commit/8289ba182284c6c3307dc3fe0db3a025ec4c3b10))
* **menu:** ensure sidebar is scrollable ([39e0ae4](https://github.com/Videodock/ott-web-app/commit/39e0ae40fa57d0e4067507be353d8acd678e845a))
* **menu:** prevent text to go on two lines in buttons ([2ccc932](https://github.com/Videodock/ott-web-app/commit/2ccc93219c75dcfbd5c6996e2fb65f98d2b093aa))
* **payment:** fix new subscription not showing in UI ([525478f](https://github.com/Videodock/ott-web-app/commit/525478f01168bcf675669db2e9eb1648b6e892ee))
* **payment:** layout of invoices on mobile ([04b928f](https://github.com/Videodock/ott-web-app/commit/04b928f735292487738f32f4225f1f92809c71f5))
* **payment:** redirect to wrong media after payment ([bb45b52](https://github.com/Videodock/ott-web-app/commit/bb45b52f52b2dc180b36672cf0184959959ed31d))
* **payment:** ui not updating after paypal payment ([8b79cc1](https://github.com/Videodock/ott-web-app/commit/8b79cc1b43c93ed6afd92d9edfaea3db1e067af9))
* **project:** checkbox values for customer consents ([1b89dac](https://github.com/Videodock/ott-web-app/commit/1b89dac0142fc3a443b208f0060ace3f027f755b))
* **project:** favicons & preconnect jwapp.com ([bf90a73](https://github.com/Videodock/ott-web-app/commit/bf90a73dd1cbc8d7280fd0c43456fc5276e627ee))
* **project:** overall styling of the hub page ([eed5f24](https://github.com/Videodock/ott-web-app/commit/eed5f24f541e8f7f04a8f1f7bb23ecf6161bccf6))
* **search:** prevent keyboard from persisting in searchbar on submit ([9bc80de](https://github.com/Videodock/ott-web-app/commit/9bc80decce0e964a51ad28d1132d253f416b5259))
* shelf margin and grid loading state ([72f62a7](https://github.com/Videodock/ott-web-app/commit/72f62a7359b6eba90b3afe810c46f1333b5ae124))
* show error on email field when invalid ([ef6392b](https://github.com/Videodock/ott-web-app/commit/ef6392bcf973b1086fd2db720db0e44af92a7f89))
* show error when delete account fails ([92ddd9c](https://github.com/Videodock/ott-web-app/commit/92ddd9c3f404ef50e1e3bf019c14dd7a1ee9ef21))
* **user:** error when updating customer ([eba0e2a](https://github.com/Videodock/ott-web-app/commit/eba0e2a5a0093a2bee1971ddfa95a0a43d8e1178))
* **user:** update question value while typing ([0a1ed83](https://github.com/Videodock/ott-web-app/commit/0a1ed83a311e4671bb482f3d35f5d5ac019def58))
* **videodetail:** buttons not aligning properly ([ddcb214](https://github.com/Videodock/ott-web-app/commit/ddcb214116bd388ab0e5d477bd40a82ef54a688a))
* **videodetail:** don’t render empty duration metadata ([0f11a77](https://github.com/Videodock/ott-web-app/commit/0f11a778ede1b6dc68f9338c4677e6d902c15a9a))
* **videodetail:** page behind player too wide ([94129dc](https://github.com/Videodock/ott-web-app/commit/94129dcd428002ce2c8daafeead0a50e9cc3d2ed))

### Form validation improvements for screen readers

We revamped the form validation by a lot. Most notably, we now announce form validation errors for screen reader users (while not presenting them visually), and we have made every submit button always enabled. This last one is a requirement from an a11y standpoint. For a full overview of improvements, see: Videodock/ott-web-app#107

### Improved modals for accessibility

Even though we recently introduced and improved the use of ARIA modals and dialogs. We improved it even further. The most notable change is that the Cinema component (video player) is now wrapped in a Modal component. This has certain benefits:

- Since the video player lays on top of the page and behaves like a (fullscreen) modal, it becomes more semantically correct. 
- We can reuse a lot of logic (eg. `inert`, programmatically setting and resetting the focus consistently and listening to the Escape key to close)

The Cinema component uses a grow/shrink animation instead of a fade in/out as an extra benefit.

We also improved the initial focus and restored the focus into and from the modal. Some edge cases were on some screen readers (iOS and Android).

### Enhanced user menu & profile menu for accessibility

The user and profile menu contained a heading and horizontal rule within the `<li>` elements. This caused the screen reader to announce the amount of list items incorrectly. Also, the HTML structure needs to be corrected.

Because the Usermenu was previously merely used as a pass-through component, passing along a lot of props from `<UserMenu />` to `<ProfileMenu />`, it felt logical to split them into separate components.

We also applied some other ARIA attributes such as `role="button"` to MenuButtons (actions instead of links) and `aria-current="true" `to the active elements (eg. currently selected profile).

### Shelf optimised for screen readers

We optimized navigation through the shelves for screen reader users. We looked at other examples online and applied the best practices. We also added a page indicator for screen readers. It announces the current state of the carrousel as pages. It gets read when you use the chevron buttons.

We also increased the hit area of the chevron (next/prev) buttons according to the WCAG specs.

### Small accessibility improvements

These are small but could be of particular interest:
1. We increased the line height to 1.5 for running text, as is required by the WCAG specs.
2. We made improvements to the video metadata for screen readers so that they will be read individually instead of in one sentence
3. The search bar has been optimized for accessibility
4. When the pipe character is used within the footer text. The footer will be transformed into a list. This is a common practice

### Offer refactor

We've moved most of the logics behind the Choose Offer step from the UI to the CheckoutController, using a similar approach as we recently did for the Registration and Checkout steps. This should make the UI more simple and make the useOffer hook reusable between containers and potential other platforms.

We also fixed a bug where after a PayPal payment the incorrect offer was checked for entitlement, and raised the delay in useCheckAccess to compensate for a delay in the Cleeng backend that results into an out-of-sync UI after buying a subscription.

## Related WCAG 2.1 (AA) requirements

- WCAG 1.1.1 - Non-text Content
- WCAG 1.4.12 - Text Spacing
- WCAG 2.4.3 - Focus order
- WCAG 3.3.2 - Labels or instructions
- WCAG 3.3.3 - Error suggestion
- WCAG 4.1.2 - Name, Role value


.github/workflows/pr-deployment-previews.yml
budzworthy, line 1: How is this different from the firebase previews? Maybe it makes sense for this to stay in the videodock repo?

packages/common/src/utils/formatting.ts
```diff
@@ -15,17 +13,17 @@ export const formatDurationTag = (seconds: number): string | null => {
...
-  if (!duration) return null;
+export const formatDuration = (duration: number) => {
+  if (!duration) return '';
 
   const hours = Math.floor(duration / 3600);
   const minutes = Math.round((duration - hours * 3600) / 60);
 
-  const hoursString = hours ? `${hours}h ` : '';
-  const minutesString = minutes ? `${minutes}m ` : '';
+  const hoursString = hours ? `${hours}hrs ` : '';
```
budzworthy, line 25: Should these be translations?

.github/workflows/pr-deployment-previews.yml
ChristiaanScheermeijer: We use this workflow for our forked repository and thought it could serve as an example workflow by disabling it here.

The problem is that we can't keep the repositories synced when this file is in all our branches. We could set up a different way of deploy previews using a webhook and lambda, but this was easier for now.

If you think this file should be removed, we can do that.

packages/common/src/utils/formatting.ts
ChristiaanScheermeijer: Yes, but we found more similar cases that need to be updated. This quick fix is an improvement because instead of screenreaders reading 10 meters, it now reads 10 minutes in English and Spanish.

See this discussion: https://github.com/Videodock/ott-web-app/pull/116#discussion_r1490592227

We have made a ticket for this, but it's not a high-priority ticket.

packages/common/src/controllers/AccountController.ts
```diff
@@ -467,7 +469,17 @@ export default class AccountController {
     assertModuleMethod(this.accountService.deleteAccount, 'deleteAccount is not available in account service');
     assertFeature(canDeleteAccount, 'Delete account');
 
-    return this.accountService.deleteAccount({ password });
+    try {
+      await this.accountService.deleteAccount({ password });
+    } catch (error: unknown) {
+      const message = error instanceof Error ? error.message.toLowerCase() : '';
+
+      if (message.includes('invalid credentials')) {
```
AntonLantukh, line 477: Nice fix, thank you!

packages/common/src/services/integrations/jwp/JWPAccountService.ts
```diff
@@ -344,7 +321,6 @@ export default class JWPAccountService extends AccountService {
       });
 
       const user = this.formatAccount(data.account);
-      user.metadata = await this.getAccountMetadata(data.account);
```
AntonLantukh, line 347: Will we still get favorites / continue watching saved in the account? Do we expect it to be done separately (updateWatchHistory / updateFavorites) in case we remove it here?

packages/common/src/utils/media.ts
```diff
@@ -49,3 +51,39 @@ export const getLegacySeriesPlaylistIdFromEpisodeTags = (item: PlaylistItem | un
...
+  if (media.pubdate) metaData.push(String(new Date(media.pubdate * 1000).getFullYear()));
+  if (!episodesLabel && duration) metaData.push(duration);
+  if (episodesLabel) metaData.push(episodesLabel);
+  if (media.genre) metaData.push(media.genre);
+  if (media.rating) metaData.push(media.rating);
+
+  return metaData;
+};
+
+export const createPlaylistMetadata = (playlist: Playlist, episodesLabel?: string) => {
```
AntonLantukh: Hm, I wonder if playlist metadata should also be part of the media utiles. Should we use `metadata.ts` for metadata generation?

packages/ui-react/src/components/ErrorPage/ErrorPage.tsx
```diff
@@ -32,11 +32,12 @@ const ErrorPage = ({ title, message, learnMoreLabel, ...rest }: Props) => {
 
 export const ErrorPageWithoutTranslation = ({ title, children, message, learnMoreLabel, error, helpLink }: Props) => {
   const logo = useConfigStore((s) => s.config?.assets?.banner);
+  const alt = ''; // intentionally empty for a11y, because adjacent text alternative
```
AntonLantukh, line 35: Hm, not sure I understand why it can't be a 'Company logo'.

packages/ui-react/src/containers/Layout/Layout.tsx
```diff
@@ -199,22 +202,19 @@ const Layout = () => {
...
-        {!!footerText && (
-          <MarkdownComponent
-            // The extra style below is just to fix the footer on mobile when the dev selector is shown
-            className={classNames(styles.footer, { [styles.testFixMargin]: IS_DEVELOPMENT_BUILD })}
-            markdownString={footerText}
-            tag="footer"
-            inline
-          />
-        )}
+        {!!footerText && <Footer text={footerText} />}
```
AntonLantukh, line 205: Maybe we could retrieve it right in the component?

packages/ui-react/src/components/Modal/Modal.module.scss
```diff
@@ -8,7 +8,7 @@
   left: 0;
   z-index: variables.$modals-z-index;
   width: 100vw;
-  height: calc(100vh - calc(100vh - 100%));
+  height: 100dvh;
```
AntonLantukh, line 11: Nice one!

packages/ui-react/src/components/Player/Player.tsx
```diff
@@ -72,7 +72,16 @@ const Player: React.FC<Props> = ({
   const handleComplete = useEventCallback(onComplete);
   const handleUserActive = useEventCallback(onUserActive);
   const handleUserInactive = useEventCallback(onUserInActive);
-  const handleFirstFrame = useEventCallback(onFirstFrame);
+  const handleFirstFrame = useEventCallback(() => {
+    // a11y: auto focus to play/pause button
+    const playerContainer = playerRef.current?.getContainer();
+    const playPauseButton = playerContainer?.querySelector(
```
AntonLantukh, line 78: Can we do it in `handleFirstFrame` function?

packages/ui-react/src/components/Sidebar/Sidebar.tsx
```diff
@@ -18,16 +19,62 @@ const Sidebar: React.FC<SidebarProps> = ({ isOpen, onClose, children }) => {
...
+
   const htmlAttributes = { inert: !isOpen ? '' : undefined }; // inert is not yet officially supported in react. see: https://github.com/facebook/react/pull/24730
 
   useEffect(() => {
     if (isOpen) {
+      // Before inert on the body is applied in Layout, we need to set this ref
       lastFocusedElementRef.current = document.activeElement as HTMLElement;
+
+      // When opened, adjust the margin-right to accommodate for the scrollbar width to prevent UI shifts in background
+      document.body.style.marginRight = `${scrollbarSize()}px`;
```
AntonLantukh, line 32: Could you please elaborate on this one? Not sure I see any changes as it equals to 0.

packages/ui-react/src/components/Sidebar/Sidebar.tsx
```diff
@@ -18,16 +19,62 @@ const Sidebar: React.FC<SidebarProps> = ({ isOpen, onClose, children }) => {
...
+      }
+    } else {
+      document.body.style.removeProperty('margin-right');
+      document.body.style.removeProperty('overflow-y');
+    }
+  }, [isOpen]);
+
+  useEffect(() => {
+    const handleEscKey = (event: KeyboardEvent) => {
+      if (isOpen && event.key === 'Escape') {
```
AntonLantukh, line 47: Do we need it in case we show the sidebar on mobile and tablets?

packages/ui-react/src/components/Player/Player.tsx
```diff
@@ -72,7 +72,16 @@ const Player: React.FC<Props> = ({
...
+      '.jw-controlbar [aria-label="Play"], .jw-button-container [aria-label="Pause"]',
+    ) as HTMLDivElement | null;
+    playPauseButton?.focus();
```
AntonLantukh, line 81: Looks like in some cases it prevents us from clicking on a "Back" button.

packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaMovie/MediaMovie.tsx
```diff
@@ -70,18 +71,24 @@ const MediaMovie: ScreenComponent<PlaylistItem> = ({ data, isLoading }) => {
...
   const pageTitle = `${data.title} - ${siteName}`;
   const canonicalUrl = data ? `${window.location.origin}${mediaURL({ media: data })}` : window.location.href;
 
-  const primaryMetadata = formatVideoMetaString(data);
+  const primaryMetadata = <VideoMetaData attributes={createVideoMetadata(data)} />;
   const shareButton = <ShareButton title={data.title} description={data.description} url={canonicalUrl} />;
-  const startWatchingButton = <StartWatchingButton item={data} playUrl={mediaURL({ media: data, playlistId: feedId, play: true })} />;
+  const startWatchingButton = (
+    <StartWatchingButton
+      key={id} // necessary to fix autofocus on TalkBack
```
AntonLantukh, line 87: Should the button be focused when navigating to the series / movie page?

packages/hooks-react/src/useForm.ts
```diff
@@ -114,6 +120,12 @@ export default function useForm<T extends GenericFormValues>({
           }
         }
 
+        if (error.inner.every((error) => error.type === 'required')) {
+          newErrors.form = t('validation_form_error.required');
+        } else {
+          newErrors.form = t('validation_form_error.other');
+        }
```
AntonLantukh, line 127: Why do we need to differentiate here?

packages/hooks-react/src/useOffers.ts
```diff
@@ -1,45 +1,62 @@
...
+
+  const chooseOffer = useMutation({
+    mutationKey: ['chooseOffer'],
+    mutationFn: checkoutController.chooseOffer,
+  });
+
+  const switchSubscription = useMutation({
+    mutationKey: ['switchSubscription'],
+    mutationFn: checkoutController.switchSubscription,
+    onSuccess: () => accountController.reloadSubscriptions({ delay: 7500 }), // @todo: Is there a better way to wait?
```
AntonLantukh, line 37: Do we wait for Cleeng here?

packages/ui-react/src/components/EditPasswordForm/EditPasswordForm.tsx
```diff
@@ -36,13 +37,18 @@ const EditPasswordForm: React.FC<Props> = ({
...
+  validationError,
   email,
   onResendEmailClick,
 }: Props) => {
   const { t } = useTranslation(['account', 'user']);
   return (
     <form onSubmit={onSubmit} data-testid={testId('forgot-password-form')} noValidate className={styles.forgotPasswordForm}>
-      {errors.form && <FormFeedback variant="error">{errors.form}</FormFeedback>}
+      {errors.form && (
+        <FormFeedback variant="error" visible={!validationError}>
```
AntonLantukh, line 48: Hm, I wonder if we it makes sense have a <Form /> component which could include FormFeedback. That way we won't need to add it individually every time.

AntonLantukh (commented):
Thanks for the changes! Sorry it took me time to review it.

packages/common/src/services/integrations/jwp/JWPAccountService.ts
ChristiaanScheermeijer: The favorites and history were proxied through the `user.metadata` to `customer.externalData`. Now, the service does all the work and will use the InPlayer APIs or `customer.externalData` for Cleeng.

packages/common/src/utils/media.ts
ChristiaanScheermeijer: That's an easy change 🙂

packages/ui-react/src/components/ErrorPage/ErrorPage.tsx
ChristiaanScheermeijer: Because the logo doesn't add any value for screen readers, it's recommended to use an empty alt attribute: https://www.w3.org/WAI/tutorials/images/decorative/

We should use an alt attribute describing the image when dealing with other images (like a movie poster). But this is difficult because no good image description is available (descriptive !== "Sintel movie poster").

packages/ui-react/src/containers/Layout/Layout.tsx
ChristiaanScheermeijer: I prefer to keep the logic and data retrieval in containers/screens. The Footer component is UI only.

packages/ui-react/src/components/Player/Player.tsx
ChristiaanScheermeijer: This seems to be the most logical event because a) the player is ready for playback, and b) the player controls are visible.

The best alternative would be the ready event. However, this won't work well when using the InlinePlayer because it will disrupt the navigation by focusing the play button when loading the page.

Do you think of an alternative event?

packages/ui-react/src/components/Sidebar/Sidebar.tsx
ChristiaanScheermeijer: The `scrollbarSize` returns the actual scrollbar width of the current browser. When opening a modal (or sidebar), the scrollbar will temporarily hide to prevent background scrolling. But when the scrollbar is visible, hiding it will cause a layout shift.

This `marginRight` prevents the layout shift by reserving the previous scrollbar spacing.

This is the same method used in the Modal component. We have a ticket to look into this further in the next sprint. Possibly also introducing some modal manager because stacking multiple modals doesn't work very well at the moment.

packages/ui-react/src/components/Sidebar/Sidebar.tsx
ChristiaanScheermeijer: We aligned the sidebar behavior with the modal component because it essentially also presents a modal. When doing so, it is expected to support certain keyboard shortcuts. Although it might not be needed for mobile and tablets, it doesn't hurt to have it there just in case someone uses assertive technology on these specific devices or uses 200% zoom:

**1280px viewport with 200% zoom**
image

packages/ui-react/src/components/Player/Player.tsx
ChristiaanScheermeijer: We haven't run into this problem while testing. Can you elaborate?

packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaMovie/MediaMovie.tsx
ChristiaanScheermeijer: Because the OTT Web App is an SPA, it can be difficult for screen reader users to realize that the page has been navigated after clicking a card. 

To mitigate this issue, the focus can be reset to the first part that is changed after the route. We chose to focus the first interactive, but alternatively, we could focus the `<main />` or `<h1>` element instead.

@langemike @royschut can you elaborate on why we chose to focus the Start watching button?

packages/hooks-react/src/useForm.ts
ChristiaanScheermeijer: This is to make sure the form error is read correctly for screen reader users. Not all errors are visually shown but are read by the screen reader to notify users when the form contains errors. 

When all errors relate to required fields, we can make the message concrete by saying, "Please fill in all required fields". However, when there are multiple error types, we just say, "Please correct your mistakes before proceeding".

image

packages/hooks-react/src/useOffers.ts
ChristiaanScheermeijer: Both InPlayer and Cleeng need some time to process external payments. Even though the mutation succeeds, the subscription endpoint can yield an outdated response.

FYI @royschut has worked on improving this part by introducing a delay and retry mechanism: https://github.com/Videodock/ott-web-app/pull/147 (needs a rebase)

packages/ui-react/src/components/EditPasswordForm/EditPasswordForm.tsx
ChristiaanScheermeijer: For a later refactor 😄

ChristiaanScheermeijer:
Hi @dbudzins @AntonLantukh, thank you again for reviewing this PR 😄 

Correct me if I'm wrong, but after evaluating your feedback, these are the open points:

- Disable or remove AWS S3 workflow (waiting on @dbudzins)
- Move metadata utils to metadata.ts? (@dbudzins opinion?)
- Is there a better event for focusing the player UI? (alternatives by @AntonLantukh)
- Back button bug? (need more details @AntonLantukh)

I've responded to each conversation respectively, let's keep the discussions there for the overview.

packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaMovie/MediaMovie.tsx
langemike: It seemed like the most logical element above the fold, making it the quickest way for a user to instantly play a video.

The focus will only be set on this button when the video content on the page switches. This means it only happens when you choose a video from within the CardGrid. It doesn't happen when you navigate to the page.

The alternatives are also viable options.

ChristiaanScheermeijer: Thanks @langemike! 

We can look into solving this for all pages in a later sprint. Reach Router solves this automatically, we could take inspiration from that:

https://reach.tech/router/accessibility

Or: https://jshakespeare.com/accessible-route-change-react-router-autofocus-heading/

An alternative for SPA routing is announcing "Navigated to {page}" using an `aria-live` region. But I think the focus method is preferable because it is similar to routing a non-SPA site (e.g., hard page reload).

packages/hooks-react/src/useOffers.ts
royschut: The `7500` duration was copied from what [previously was the submit handler](https://github.com/jwplayer/ott-web-app/blob/develop/packages/ui-react/src/containers/AccountModal/forms/ChooseOffer.tsx#L68) handling a subscription switch, indeed waiting for processing time in the backend.

The [coming PR ](https://github.com/Videodock/ott-web-app/pull/147) that @ChristiaanScheermeijer mentions will wait for a shorter time and simply retry a couple of times if it isn't ready.

packages/ui-react/src/components/ErrorPage/ErrorPage.tsx
AntonLantukh: I agree that if it is a decorative image or a background image then it doesn't make much sense to add `alt` to it. I think I am also okay leaving it as an empty string here.

packages/ui-react/src/components/Player/Player.tsx
AntonLantukh: Try to open a video, hover on the back button and click on it.

ChristiaanScheermeijer: Hmm, I see what you mean. It's weird that we haven't noticed this before. 

It seems that there is a small race condition going on between the JW Player `userInactive` event when clicking the back button. I think we are only seeing this now because we have moved the focus to the controls and by clicking the back button, the player controls unfocus logic hides the UI immediately.

The `onMouseDown` event fires, but the `onClick` and `onMouseUp` events don't because the element is hidden.

I could solve it by delaying the `userInactive` event by 100ms, but relates directly to the time between the mouse down and mouse up events for this to work. Beceause the player fires the `userInactive` event on the onMouseDown event.

```tsx
  const handleUserInactive = useCallback(() => {
    setTimeout(() => setUserActive(false), 100);
  }, []);
```

Perhaps a better solution would be to prevent hiding the top bar when we detect a mouseDown on the back button.

packages/hooks-react/src/useOffers.ts
AntonLantukh: Thank you!

packages/ui-react/src/components/Player/Player.tsx
ChristiaanScheermeijer: I fixed it properly by keeping the player overlay visible when it has focus. This also solves the issue that the overlay hides when navigating with the keyboard.

[7bb3f20](https://github.com/jwplayer/ott-web-app/pull/455/commits/7bb3f207ef794614d09b33d248dfcf4c7b220520)

packages/ui-react/src/components/Sidebar/Sidebar.tsx
AntonLantukh: Makes sense to me. Though I didn't manage to get any value except 0. Can be also connected with the way Chrome / Safari  handle `overflow: scroll` on Mac now.

packages/ui-react/src/components/Player/Player.tsx
AntonLantukh: Maybe using `playerInstance` in the `handleFirstFrame` method (PlayerContainer)?

packages/ui-react/src/components/Player/Player.tsx
AntonLantukh: Works great! Thanks!

ChristiaanScheermeijer:
@AntonLantukh @dbudzins this PR is ready for merge 😄

budzworthy (approved):

PR #459, opened 2024-03-12
Chore / maintenance 2024-q1

ChristiaanScheermeijer:
## Description

@MelissaDTH has performed a maintenance update for the OTT Web App repo. In a maintenance update, we fix security issues found by dependabot or `yarn audit` and try upgrading major dependency upgrades.

The upgrades in this PR are:

- eslint: `^7.31.0` -> `^8.57.0`
- typescript: `^4.3.4` -> `5.2.3` (@typescript/eslint-parser doesn't support higher)
- vitest: `^0.34.6` -> `^1.3.1`
- vite: `^4.4.8` -> `^5.1.5`
- vite-plugin-stylelint: `^4.3.0` -> `^5.3.1`


tsconfig.base.json
```diff
@@ -13,7 +13,7 @@
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "allowSyntheticDefaultImports": true,
-    "importsNotUsedAsValues": "error",
+    "verbatimModuleSyntax": false,
```
AntonLantukh: We can't use it now because we use types as values for services / controllers injection, right?

ChristiaanScheermeijer: We actually can! A few files needed an update 😄

AntonLantukh (approved):

ChristiaanScheermeijer:
One small update. The lighthouse run failed due to the lack of support for ESM syntax. I've renamed `lighthouserc.js` to `lighthouserc.cjs`, reverted the ESM update, and updated the workflow to point to the `*.cjs` file.

Hopefully next update we can clean some of the CJS files 😄

PR #462, opened 2024-03-15
chore: replace depcheck with knip

budzworthy:
## Description

Depcheck was getting annoying with monorepo tools and refactoring. It doesn't work with any local imports besides relative paths, so tsconfig baseurl and paths all have to be manually added to the depcheck ignore file. See https://github.com/depcheck/depcheck/issues/564.

Knip does the same thing, but does take tsconfig and other environments into consideration. Additionally, knip uses a single, combined config file, which seems easier to manage than all of the fragmented depcheck files.

### Steps completed:

~- [ ] Acceptance criteria met~
~- [ ] Unit tests added~
~- [ ] Docs updated (including config and env variables)~
~- [ ] Translations added~
~- [ ] UX tested~
~- [ ] Browsers / platforms tested~
- [x] Reviewed own code


ChristiaanScheermeijer (approved):
Nice cleanup @dbudzins! 

I do have a slight concern about having the config in the workspace root, because when adding a platform, it needs to be updated as well. This could be annoying when syncing, but I guess it doesn't update that often either...

PR #469, opened 2024-03-21
feat(project): OWA-56 add DRM check for CDN analytics

olga-jwp:
## Description

[OWA-52](https://jwplayer.atlassian.net/browse/OWA-52)

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

[OWA-52]: https://jwplayer.atlassian.net/browse/OWA-52?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/common/src/utils/analytics.ts
+++ b/packages/common/src/utils/analytics.ts
@@ -5,9 +5,11 @@ import type { PlaylistItem, Source } from '../../types/playlist';
 
 export const attachAnalyticsParams = (item: PlaylistItem) => {
   // @todo pass these as params instead of reading the stores
-  const { config } = useConfigStore.getState();
+  const { config, settings } = useConfigStore.getState();
   const { user } = useAccountStore.getState();
   const { profile } = useProfileStore.getState();
+
+  const vendorId = settings.vendorId;
 
   const { sources, mediaid } = item;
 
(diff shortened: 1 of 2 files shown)


packages/common/src/utils/analytics.ts
```diff
@@ -17,22 +19,31 @@ export const attachAnalyticsParams = (item: PlaylistItem) => {
...
+    const urlSearchParams = url.searchParams;
 
     const mediaId = mediaid.toLowerCase();
     const sourceUrl = url.href.toLowerCase();
 
-    // Attach user_id and profile_id only for VOD and BCL SaaS Live Streams
-    const isVOD = sourceUrl === `https://cdn.jwplayer.com/manifests/${mediaId}.m3u8`;
-    const isBCL = sourceUrl === `https://content.jwplatform.com/live/broadcast/${mediaId}.m3u8`;
+    const cdnUrlVOD = `https://cdn.jwplayer.com/manifests/${mediaId}.m3u8`;
+    const cdnUrlBCL = `https://content.jwplatform.com/live/broadcast/${mediaId}.m3u8`;
```
ChristiaanScheermeijer, line 28: Do you think we should remove the extension check? The sources could also contain MPEG-DASH or MP4 files that won't have the analytic parameters.

packages/common/src/utils/analytics.ts
```diff
@@ -5,10 +5,12 @@ import type { PlaylistItem, Source } from '../../types/playlist';
 
 export const attachAnalyticsParams = (item: PlaylistItem) => {
```
ChristiaanScheermeijer, line 6: Not related to your change, but this function mutates the given item, should we create a new object instead?

```ts
export const attachAnalyticsParams = (item: PlaylistItem) => {
  const updatedSources = item.sources.map(source => {
    // logic
    return { ...source, file: url.toString() };
  });

  return { ...item, sources: updatedSources };
};
```

packages/common/src/utils/analytics.ts
olga-jwp: Hey Christiaan, that is a good catch. I was referring to Marco's [previous notes](https://jwplayer.atlassian.net/browse/OWA-52) while working on this task. 
Even though it is still blocked by Media Core team I would suggest discussing this in our monthly sync.
CC: @dbudzins @AntonLantukh

PR #474, opened 2024-03-26
Fix: redirect when no integration is configured

ChristiaanScheermeijer:
## Description

When navigating directly to a user page (`/u/payments`) when no integration is configured, an infinite spinner shows, this is caused by the `AccountStore#loading` state being set to `true` initially, but this is never reset to `false` when no integration is initialized.

CC: @langemike (can't assign as a reviewer)

### DIFF
--- a/packages/common/src/stores/AccountStore.ts
+++ b/packages/common/src/stores/AccountStore.ts
@@ -19,7 +19,7 @@ type AccountStore = {
 };
 
 export const useAccountStore = createStore<AccountStore>('AccountStore', (set, get) => ({
-  loading: true,
+  loading: false,
   user: null,
   subscription: null,
   transactions: null,


langemike (commented):
LGTM!

One note: in your PR title and commit message mentions a redirect. Is there maybe missing something? Because this only fixes showing an infinite spinner

ChristiaanScheermeijer:
Thanks @langemike, the redirect logic is already in place in the `User.tsx` page:

https://github.com/jwplayer/ott-web-app/blob/71433abfca80431243effc246afdbe8c6901521e/packages/ui-react/src/pages/User/User.tsx#L70-L74

But the condition is never met because of the loading state permanently being set to `true`.

AntonLantukh (approved):

ChristiaanScheermeijer:
I also have an alternative solution by inversing the logic.

1. We set the `AccountStore#loading` property to `false`
2. When we call `UserService#initialize` we first set the loading state to `true`

Because the app defers rendering the route until the AccountStore is initialized, the result should be the same...

AntonLantukh:
@ChristiaanScheermeijer I think I would prefer having `false` by default. It is just intuitively clear.

ChristiaanScheermeijer:
@AntonLantukh updated the PR. We already set the loading state to `true` in the `AccountController#initialize`, so only the default state had to be updated 😄

AntonLantukh (approved):

AntonLantukh:
@ChristiaanScheermeijer thank you!

royschut (approved):

PR #475, opened 2024-03-26
Feat / reload site on update

ChristiaanScheermeijer:
## Description

This small configuration change will ensure users will always use the latest version of the web app. Before, you need to visit the site, close it or clear the cache to ensure the latest version is used.

This is because the PWA will wait until a suitable moment to activate the "waiting" next version of the deployment. On many occasions, publishers (and we) want to be able to deploy fixes immediately.

Read more about this here:

https://vite-pwa-org.netlify.app/guide/auto-update.html

 CC: @langemike

### DIFF
--- a/platforms/web/src/index.tsx
+++ b/platforms/web/src/index.tsx
@@ -36,4 +36,7 @@ if (rootElement) {
   console.info('Application - rootElement not found');
 }
 
-registerSW();
+const refresh = registerSW({
+  immediate: true,
+  onNeedRefresh: () => refresh(true),
+});
--- a/platforms/web/vite.config.ts
+++ b/platforms/web/vite.config.ts
@@ -59,6 +59,7 @@ export default ({ mode, command }: ConfigEnv): UserConfigExport => {
       StylelintPlugin(),
       svgr(),
       VitePWA({
+        registerType: 'autoUpdate',
         manifestFilename: 'manifest.json',
         manifest: {
           name: app.name,


langemike:
The SW caused a quite aggressive cache, which could only be circumvented by hard-refreshing. When a new version has been released we want our users to immediately see these changes. Therefore I think auto-reload is a good approach.

The auto-reload will only happen when users visit the website in a new window/tab and not during their visit in the active tab. I approve.

royschut (approved):

AntonLantukh:
@langemike hm, are you sure the current active tab won't be refreshed if we call `registerSW` again? Do we also need to 'reregister' sw or can we just leave the body function empty, like `onNeedRefresh: () => {}`?

```
The disadvantage of using this behavior is that the user can lose data in any browser windows/tabs in which the application is open and is filling in a form.
```

Btw, pretty nice implementation with react hooks (custom prompt is used):
https://vite-pwa-org.netlify.app/frameworks/react.html#prompt-for-update

langemike:
> hm, are you sure the current active tab won't be refreshed if we call registerSW again? Do we also need to 'reregister' sw or can we just leave the body function empty, like onNeedRefresh: () => {}?

I could not find an easy answer to that while reading the docs and source code. I'm not too familiar with ServiceWorkers and I reasonably assumed that in-memory JS code will not cause `onNeedRefresh` to be called within a active session. Of course we need to verify this.

budzworthy (approved):

ChristiaanScheermeijer:
@AntonLantukh the SW is only refreshed when opening the page, we don't periodically poll for new versions in active sessions. This means that losing data is limited to the following scenario:

1. Open a new tab and navigate to ott-web-app
2. Click the sign-up button and fill in the form
3. **new version is released**
4. Open a second tab and navigate to ott-web-app
5. The second tab sees the new SW version and reloads all pages
6. The first tab is refreshed, and the form data is lost

So I think that:

1. Users can ignore the reload page modal and still be using an outdated version (which may contain bugs)
2. The sign-up form isn't massive, and we assume that people don't leave it stale
3. When the sign-up form is emptied due to a refresh, the user can fill it in pretty quickly

royschut (approved):

PR #476, opened 2024-03-27
Fix / missing feedback when submitting coupon

ChristiaanScheermeijer:
## Description

This PR addresses an issue we found during testing. For some reason, there was no feedback when submitting a coupon code. I tried to reproduce this, but the only way I could was to make the `updateOrder` fetch fail. For example, Google Chrome can be put in "Offline" mode using the network tab.

Because there was no feedback, I traced the code and remembered (and spotted some TOODs :-)) that the `updateOrder` method was a bit funky.

This PR addresses some problems and solves the TODOs:

1. Prevent `responseData.errors` to contain Error objects (only strings are expected)
2. Handle errors in try/catch only in the CheckoutController
3. Throw errors with "fixed" messages in integrations
4. Add a generic form error when the request fails due to (service outage/network problems)
5. Prevent mutating the order in the JWP updateOrder method

Although the generic error shows "Unknown error", we know the reason. But we should solve this problem structurally, not just for the `updateOrder` method.

CC: @langemike

### DIFF
--- a/packages/common/src/controllers/CheckoutController.ts
+++ b/packages/common/src/controllers/CheckoutController.ts
@@ -91,32 +91,22 @@ export default class CheckoutController {
   };
 
   updateOrder = async (order: Order, paymentMethodId?: number, couponCode?: string | null): Promise<void> => {
-    let response;
-
     try {
-      response = await this.checkoutService.updateOrder({ order, paymentMethodId, couponCode });
+      const response = await this.checkoutService.updateOrder({ order, paymentMethodId, couponCode });
+
+      if (response.responseData.order) {
+        useCheckoutStore.getState().setOrder(response.responseData?.order);
+      }
     } catch (error: unknown) {
-      // TODO: we currently (falsely) assume that the only error caught is because the coupon is not valid, but there
-      //  could be a network failure as well (JWPCheckoutService)
-      throw new FormValidationError({ couponCode: [i18next.t('account:checkout.coupon_not_valid')] });
-    }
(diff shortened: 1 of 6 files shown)


packages/common/src/controllers/CheckoutController.ts
```diff
@@ -91,32 +91,22 @@ export default class CheckoutController {
...
-      throw new FormValidationError({ couponCode: [i18next.t('account:checkout.coupon_not_valid')] });
-    }
+      const response = await this.checkoutService.updateOrder({ order, paymentMethodId, couponCode });
 
-    if (response.errors.length > 0) {
-      // clear the order when the order doesn't exist on the server
-      if (response.errors[0].includes(`Order with ${order.id} not found`)) {
-        useCheckoutStore.getState().setOrder(null);
+      if (response.responseData.order) {
+        useCheckoutStore.getState().setOrder(response.responseData?.order);
```
AntonLantukh, line 98: Should we also set form error here in case `responseData.order` is not defined?

AntonLantukh (approved):

packages/common/src/controllers/CheckoutController.ts
AntonLantukh: @ChristiaanScheermeijer not a big thing, so I can also merge it as is.

ChristiaanScheermeijer: Fine by me. It should work the same as before

PR #478, opened 2024-03-27
Fix / Retain password input after form validations

MelissaDTH:


MelissaDTH:
Has been accidentally opened in the wrong repo, therefore closed.

PR #456, opened 2024-03-01
Docs / update docs for workspaces

ChristiaanScheermeijer:
## Description

In the last sprint, I already changed most of the documentation. During this sprint, @langemike reviewed this and made some changes.

> Note: this branch is rebased onto the `feat/accessibility-and-bug-fixes` branch

PR #482, opened 2024-04-03
Feat / improve generate translations script

ChristiaanScheermeijer:
## Description

This PR updates the i18next script that extracts all translations from the source files. I've added that this happens for each platform in the platforms folder. The workspace dependencies are resolved from the package.json and used in the i18next script.

The settings can be changed by a simple `.i18n.json` file in the platform directory (not used in the public repo). For instance, the `bundle` option creates TypeScript files, exporting all JSON files to be bundled with the platform code.

I also added a pass that checks and fills missing translations between the platforms. This is a QOL feature when syncing new changes from the public repo when translations are added.


scripts/i18next/generate.ts
```diff
@@ -0,0 +1,280 @@
...
+   */
+  resourcesPath: string;
+};
+
+function getWorkspacePackages(packageJSON: SimplePackageJSON) {
+  const deps = Object.keys(packageJSON.dependencies);
+  const devDeps = Object.keys(packageJSON.devDependencies);
+  const peerDeps = Object.keys(packageJSON.peerDependencies);
+
+  const packages = [...deps, ...devDeps, ...peerDeps].filter((name) => name.startsWith('@jwp/'));
```
budzworthy: Would it be safer to use `yarn workspaces info` to figure out which packages are local, so we don't have to rely on the `@jwp` prefix?

scripts/i18next/generate.ts
```diff
@@ -0,0 +1,280 @@
...
+   * The path where the resources are bundled
+   */
+  bundlePath: string;
+  /**
+   * The path where the resources file is created
+   */
+  resourcesPath: string;
+};
+
+function getWorkspacePackages(packageJSON: SimplePackageJSON) {
```
budzworthy: Does this function need to be recursive? I don't want to have to rely on the peer dependency model everywhere to surface 2nd+ level workspace dependencies.

scripts/i18next/generate.ts
```diff
@@ -0,0 +1,280 @@
...
+  const packages = [...deps, ...devDeps, ...peerDeps].filter((name) => name.startsWith('@jwp/'));
+
+  return packages.map((name) => {
+    return realpathSync(join(NODE_MODULES_DIRECTORY, name));
+  });
+}
+
+function extractTranslations(directories: string[], platformName: string, config: Config) {
+  const searchPaths = directories.map((path) => {
+    return join(path.replace(process.cwd(), ''), 'src/**/*.{ts,tsx}').slice(1);
```
budzworthy: This seems to rely on convention and may be error prone.

budzworthy:
@ChristiaanScheermeijer I got a little way though this PR, and I started to question the approach a bit. Some of my concerns:

1. Hybrid approach where translation tags are baked into `packages` but translation values are in `platforms`
2. Complicates translation reuse since values and namespaces may span multiple packages, but can't be shared
3. Unsure whether other platforms will be able to use i18next. I assume react native can, so hooks and ui-react are fine, but usage in common may not be
4. Current solution relies on several assumptions / conventions, such as only parsing `@jwp` packages and  `/src` directory for translation keys

Have you considered alternatives such as each packages maintaining its own translations and/or having platforms pass translations into dependent code?

ChristiaanScheermeijer:
Thanks for reviewing and raising this @dbudzins 

I agree that the translations are a bit weird in the current setup. We tried to make it as easy as possible because managing translations seems one of the hardest things to do right 😅. Perhaps we can use this PR to think and discuss better alternatives. 

I like the idea of moving the translation parser responsibility to each platform. This removes any logic from the root and each platform gains full control.

The only part missing is the automatic syncing between platforms. This already caught us a few times after syncing the open-source repo with translation changes.

I see the following options:

**Install and configure i18next-parser per platform**

This is your suggestion and seems valid to me. The parser needs to traverse to the packages folder, but we also do this for TypeScript and Vite.

- Remove `i18next-parser` from root package.json
- Install `i18next-parser` in platforms/web
- Move the ./scripts/i18next folder to ./platforms/web/scripts/i18next
  - Although this script is reusable, we might make it configure and accept parameters instead
- Move i18next-parser.config.js to platforms/web
- Move i18next scripts from root package.json to platforms/web/package.json
- Update documentation

**Platforms & packages translations**

We can take the above one step further by making each package responsible for their translations as well. Each package will have an i18n folder and scripts to generate the translations. Then, we can add an i18next script in the root to run each workspace i18next script separately.

Platforms can consume the translations by copying or bundling them with full control. Customising translations may be challenging because we want to change specific translations used in the packages.

I'm not sure yet how to manage translations when passing them to the packages. Will this not make it a lot harder to keep track of which translations are added/missing?

Unless we leverage TS for this, I'm afraid this can become quite cumbersome and difficult to keep in sync.

Let me know what you think or have more options to explore 😄

scripts/i18next/generate.ts
ChristiaanScheermeijer: Hmm interesting, running `yarn workspaces info` returns the following for the web platform:

```json
{
  "@jwp/ott-web": {
    "location": "platforms/web",
    "workspaceDependencies": [
      "stylelint-config-jwp"
    ],
    "mismatchedWorkspaceDependencies": []
  }
```

The stylelint-config-jwp package is the only package installed in the `devDependencies`. The rest is still listed in the `peerDependencies`. I do remember we talked about this before. This indeed seems to suggest we should not be using `peerDependencies` for the workspace packages...

budzworthy:
@ChristiaanScheermeijer I would lean towards the per-platform approach over the per-package approach for the reasons you've identified. It seems safer to me to rely on crawling imports to find all the potential translations instead of crawling the workspaces manually (also will tree shake the unused translations per platform.)

It might be handy to make a package to package sync script that can just be run from a standalone yarn command.

Is it ok to have i18next.t baked into common or should we try to avoid this so that common can be used everywhere even non-react platforms?

ChristiaanScheermeijer:
Thanks @dbudzins

> I would lean towards the per-platform approach over the per-package approach for the reasons you've identified. It seems safer to me to rely on crawling imports to find all the potential translations instead of crawling the workspaces manually (also will tree shake the unused translations per platform.)

Agree, I will look into that!

> It might be handy to make a package to package sync script that can just be run from a standalone yarn command.

Do you think this can live in the workspace root, or should we also make this platform-specific?

> Is it ok to have i18next.t baked into common or should we try to avoid this so that common can be used everywhere even non-react platforms?

Luckily, [i18next](https://www.npmjs.com/package/i18next) and [react-i18next](https://www.npmjs.com/package/react-i18next) are separate packages. The i18next package is framework- and even platform-agnostic, so we are safe to use that in the common package. I'm not closed to alternatives or not using i18next (it is quite a performance killer), but replacing it will not be an easy task 😅

ChristiaanScheermeijer:
@dbudzins I gave it a shot, let me know what you think. I will update the documentation if we want to proceed with this change.

Usage:

```shell
# parses all translations based on the installed workspace dependencies (with a `./src` folder)
$ cd platforms/web
$ yarn web i18next
```

```shell
# fills missing translations in the target platform (using the source platform as base)
# E.g. web --> app when translation in app is empty
$ yarn i18next:sync
Usage:
  yarn i18next:sync <sourcePlatform> <targetPlatform>

Use this script to fill missing translations in the target platform
using the translations of the source platform as base.

Example:
  yarn i18next:sync web app

Missing arguments!
```

Let me know what you think!

.github/workflows/release-create-release-candidate-branch.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           git config --global user.email 'ott-release-script@jwplayer.com'
           git fetch origin ${{ github.ref_name }}
           git merge origin/${{ github.ref_name }}
-          yarn && yarn i18next
+          yarn && yarn web i18next
```
budzworthy: The `yarn web` trick is kind of neat, but would it be cleaner if this was run via `yarn workspace @jwp/ott-web run i18next`? Does that mess up the working directory and paths?

platforms/web/package.json
```diff
@@ -14,6 +14,7 @@
     "codecept:desktop": "cd test-e2e && rm -rf \"./output/desktop\" && codeceptjs run-workers --suites ${WORKER_COUNT:=8} --config ./codecept.desktop.cjs",
     "codecept:mobile": "cd test-e2e && rm -rf \"./output/mobile\" && codeceptjs run-workers --suites ${WORKER_COUNT:=8} --config ./codecept.mobile.cjs",
     "generate-pwa-assets": "pwa-assets-generator",
+    "i18next": "npx ts-node ./scripts/i18next/parse-translations.ts",
```
budzworthy: Can you pass `./` into `parse-translations` to avoid having to use cwd?

platforms/web/scripts/i18next/parse-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+function syncTranslations() {
+  const result = execSync('yarn workspaces --json info').toString('utf-8');
+  const info = JSON.parse(result);
+  const data = JSON.parse(info.data) as WorkspaceInfo;
+
+  const dependencies = data['@jwp/ott-web']?.workspaceDependencies || [];
+
+  const globs = dependencies
+    .map((dependency) => data[dependency])
+    .filter((entry) => entry && existsSync(join('../..', entry.location, 'src')))
```
budzworthy: If we are going to use the assumption that dependencies should have a /src directory, I would prefer we throw an error when one doesn't exist instead of just filtering them out. Otherwise it may be hard to debug later down the road.

platforms/web/scripts/i18next/parse-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+type WorkspaceInfo = Record<string, WorkspaceEntry>;
+
+function syncTranslations() {
+  const result = execSync('yarn workspaces --json info').toString('utf-8');
+  const info = JSON.parse(result);
+  const data = JSON.parse(info.data) as WorkspaceInfo;
+
+  const dependencies = data['@jwp/ott-web']?.workspaceDependencies || [];
```
budzworthy: Do we need to inspect dependencies recursively?

platforms/web/scripts/i18next/parse-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+    .filter((entry) => entry && existsSync(join('../..', entry.location, 'src')))
+    .map((entry) => `'${join(entry.location, 'src/**/*.{ts,tsx}')}'`)
+    .join(' ');
+
+  const command = `npx i18next --config ${i18nextParserConfig} '${platformGlob}' ${globs}`;
+
+  console.info('Running command:');
+  console.info(` ${command}`);
+
+  execSync(command, { env: process.env, stdio: 'pipe', cwd: '../../' });
```
budzworthy: This is half nitpicking and half future proofing, but maybe it would be prudent to calculate the workspace root? If we assume that we have the root of the jwp/web package, we can simply remove the location that we got from workspace info. Or we can use something like this: https://www.npmjs.com/package/find-yarn-workspace-root

platforms/web/scripts/i18next/parse-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+const platformGlob = join(process.cwd(), 'src/**/*.{ts,tsx}');
+
+type WorkspaceEntry = {
+  location: string;
+  workspaceDependencies: string[];
+};
+
+type WorkspaceInfo = Record<string, WorkspaceEntry>;
+
+function syncTranslations() {
```
budzworthy: Can you rename this so it doesn't get confused with the `sync-translations` file?

platforms/web/scripts/i18next/parse-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+  const info = JSON.parse(result);
+  const data = JSON.parse(info.data) as WorkspaceInfo;
+
+  const dependencies = data['@jwp/ott-web']?.workspaceDependencies || [];
+
+  const globs = dependencies
+    .map((dependency) => data[dependency])
+    .filter((entry) => entry && existsSync(join('../..', entry.location, 'src')))
+    .map((entry) => `'${join(entry.location, 'src/**/*.{ts,tsx}')}'`)
+    .join(' ');
```
budzworthy: I guess it's kind of a downside that we will end up including all of the translations for any dependant packages, even if only a single function from a single file is really imported. It'd be kind of cool to use a depedency tree tool or something to get all of the individual files that the platform depends on, but that might be inefficient.

budzworthy (commented):
I think this is an ok way to start, pending some robustness improvements discussed below. Upon further digging, it seems like we might just be able to build 1 large global list of translations for the whole workspace and then, if we can figure out how (I haven't yet) tree shake the used translations at build time. That'd be a really slick solution.

.github/workflows/release-create-release-candidate-branch.yml
ChristiaanScheermeijer: `yarn web` is a shortcut for `yarn --cwd platforms/web`, which I believe is the same as running `yarn workspace @jwp/ott-web`.

The explicit `yarn workspace` command might be clearer for users familiar with yarn workspaces...

platforms/web/package.json
ChristiaanScheermeijer: I believe we still need to resolve `./` using the `cwd` because the `process.argv[2]` will result in the string `./`.

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: Hmmm interesting point. The workspaces info command returns all workspace packages including the config packages. So, throwing an error would be a bit too strict IMO.

We could log which workspace packages are skipped, but this will clutter the output because we know which ones should be ignored.

In theory, we could check if the `i18next` dependency is listed in the workspace package and only include it then...

PS: This will be solved when we parse the exact files needed for extraction (see below).

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: Good point. The `@jwp/ott-ui-react` package may depend on a `@jwp/ott-ui-web` package which is not listed in the platform dependencies.

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: Yes, good point

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: It will be less sufficient and probably need some extra dependencies, but it sounds more stable than "just" including all files...

I've tried [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser) real quick and using the API I can quickly list all files used from an entry file.

```ts
import { type IReporterOutput } from 'dependency-cruiser';

async function run() {
  const { cruise } = await import('dependency-cruiser'); // TS says no
  const ARRAY_OF_FILES_AND_DIRS_TO_CRUISE: string[] = ['platforms/web/src'];
  try {
    const cruiseResult: IReporterOutput = await cruise(ARRAY_OF_FILES_AND_DIRS_TO_CRUISE);

    console.log(cruiseResult.output.modules.slice(0, 5));
  } catch (error) {
    console.log(error);
  }
}

run();
```
This will yield:

```js
[
  {
    source: 'platforms/web/src/App.tsx',
    dependencies: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object]
    ],
    dependents: [ 'platforms/web/src/index.tsx' ],
    orphan: false,
    valid: true
  },
  {
    source: 'platforms/web/src/styles/main.scss',
    followable: false,
    coreModule: false,
    couldNotResolve: false,
    matchesDoNotFollow: false,
    dependencyTypes: [ 'local', 'import' ],
    dependencies: [],
    dependents: [ 'platforms/web/src/App.tsx' ],
    orphan: false,
    valid: true
  },
  {
    source: 'platforms/web/src/containers/Root/Root.tsx',
    dependencies: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object]
    ],
    dependents: [ 'platforms/web/src/App.tsx' ],
    orphan: false,
    valid: true
  },
  ...
}
```

The arguments will become really big, though; I could quickly try to format the arguments based on the output.

I also see some "unresolved" entries caused by aliasing that need to be solved:

```js
  {
    source: '#src/hooks/useTrackConfigKeyChange',
    followable: false,
    coreModule: false,
    couldNotResolve: true,
    matchesDoNotFollow: false,
    dependencyTypes: [ 'unknown' ]
  }
```

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: Yes, good point. CodeQL is also warning about this uncontrolled path.

.github/workflows/release-create-release-candidate-branch.yml
budzworthy: Yes, that was my thinking. I got excited when I first read this, because I though yarn had built some sort of intelligent workspace picker, but then I dug down to find the `yarn web` implementation.

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: I've played with this idea a bit more and now use `tsc --listFilesOnly` to resolve all files.

Next, I created a script that parses the result and builds a glob with nested globs to reduce the size.

For example: 

```
packages/common/src/{utils/{analytics,common}.ts,stores/AccountStore.ts}
```

However, the glob string length is currently 13336 😅

budzworthy: What would happen if we just called i18next parse on each file individually?

.github/workflows/release-create-release-candidate-branch.yml
ChristiaanScheermeijer: See the discussion here: https://github.com/jwplayer/ott-web-app/pull/456#discussion_r1517766999

Do you suggest keeping it out of the documentation and only having it available for convenience?

platforms/web/scripts/i18next/parse-translations.ts
ChristiaanScheermeijer: OK, I've tried running the i18next script using every file in the `tsc --listFilesOnly` output. This seems to fail due to unknown reasons inevitably.

I also tried mimicking some of the i18next-parser CLI and use vinyl-fs directly with the i18next-parser transform:

```ts
import { execSync } from 'child_process';
import { pathToFileURL } from 'node:url';

import vfs from 'vinyl-fs';

async function esConfigLoader(filepath: string) {
  return (await import(pathToFileURL(filepath).toString())).default;
}

async function run() {
  const cwd = process.cwd();
  const output = execSync('tsc --listFilesOnly --project ./platforms/web/tsconfig.json');
  let files = output.toString('utf-8').split('\n');

  // @ts-ignore
  const { transform: Transform } = await import('i18next-parser');

  files = files.map((file) => file.replace(cwd, '').slice(1));
  files = files.filter((file) => !file.includes('node_modules') && file !== '');

  const config = (await import(process.cwd() + '/platforms/web/i18next-parser.config.js')).default;

  let count = 0;

  vfs
    .src(files)
    .pipe(
      new Transform(config)
        .on('reading', function (file) {
          count++;
        })
        .on('error', function (message, region) {
          if (typeof region === 'string') message += ': ' + region.trim();
          console.log('  [error]   ' + message);
        })
        .on('warning', function (message) {
          console.log('  [warning] ' + message);
        })
        .on('finish', function () {
          console.log();
          console.log('  Stats:  ' + count + ' files were parsed');
        }),
    )
    .pipe(vfs.dest(process.cwd()));
}

run();
```

This does work! We do lose some flexibility like having a different config, but that shouldn't matter because this script is owned by the platform. 

So, in turn, this script does the following:

- Dry compiles the platform using the tsconfig.json
- Extract all absolute file paths of every require/import
- Makes all paths relative (not needed?) and removes all paths containing node_modules (workspace deps are symlink resolved)
- Pass all paths to vinyl-fs, which reads the files and streams them through i18next-parser
- Write the output like the i18next-parser CLI does

What do you think? 😄

ChristiaanScheermeijer: @dbudzins @AntonLantukh do you want me to submit a PR with this proposal?

budzworthy: Absolutely, if I understand you right, that sounds great! Workspace support for i18n parsing with tree shaking? :heart_eyes: 

This seems like the kind of thing that could be it's own stand-alone open source project contributed to the community if you so wanted. Awesome work!

ChristiaanScheermeijer: Do you still think the `parse-translations.ts` script belongs to the web platform, or should I make it generic now like the `sync-translations` script in the workspace root?

```
Usage:
  yarn i18next:parse <platform> [resourcePath] [--debug]

  platform           The platform to parse the translations from

  resourcePath       The path to the resources file with generated namespaces
                     Defaults to `src/i18n/resources.ts`

  --debug            Print additional debug information

Use this script to parse all translations from source files
by reading the tsconfig.json.

Example:
  yarn i18next:parse web
```

ChristiaanScheermeijer: I actually made a different repo for this and published it on NPM! 

- https://www.npmjs.com/package/i18next-parser-workspaces
- https://github.com/ChristiaanScheermeijer/i18next-parser-workspaces

The latest commit installs this package and changes the 18next package script to: 

```shell
$ i18next-workspaces ./ --resources src/i18n/resources.ts
```

The name is not definitive. Let me know if you have suggestions 😄 

Perhaps I can move the sync script as well and use subcommands instead of the confusing CLI options.

```shell
$ i18next-workspaces parse ./
$ i18next-workspaces resources ./src/i18n/resources.ts --locale en --prettier ./.prettierrc
$ i18next-workspaces sync ./ ../app --dry
```

ChristiaanScheermeijer: Latest update:

- I added subcommands to the i18next-workspaces CLI
- Add the sync (fill-missing) script to i18next-workspaces

If we want to sync two platforms, we can use:

```shell
npx i18next-workspaces ./platforms/web ./platforms/app
```

ChristiaanScheermeijer: @dbudzins @AntonLantukh any update? 😄

budzworthy (approved):
lgtm

.github/workflows/release-create-release-candidate-branch.yml
AntonLantukh: The point I had was about helping people interested in the web app to launch it. In this case you won't need to dig into docs file packages and platforms. Just use these commands and see the app. I don't mind it being `yarn workspace @jwp/ott-web` or `yarn start:web` in the docs if it could improve readability.

ChristiaanScheermeijer:
@dbudzins @AntonLantukh I think this PR is ready to be merged. I rebased the branch with develop and validated it by running `yarn workspace @jwp/ott-web run i18next`, which removed all profile-related translations (as expected since this feature was removed).

PR #485, opened 2024-04-05
Feat / modal manager

ChristiaanScheermeijer:
## Description

This PR implements a `ModalProvider` and `useModal` hook that manages the visible modals when registered via the useModals hook.

By doing so, the following features are added:

- Close specific or all open modals using the context (Native/remote back button)
- Open multiple modals without interfering with each other
- Keep the initial focus of the first modal opened in case of open (1) -> close (1) and open(2) -> close(2)
- Handle HTMLElement#inert and body scrolling from a single place
- Escape key only closes the last opened modal

This change removes lots of duplicate code from the Modal and Sidebar components.

See the whole discussion and review here: https://github.com/Videodock/ott-web-app/pull/167

### DIFF
--- a/packages/ui-react/src/containers/ModalProvider/ModalProvider.tsx
+++ b/packages/ui-react/src/containers/ModalProvider/ModalProvider.tsx
@@ -1,0 +1,225 @@
+import { createContext, type PropsWithChildren, type RefObject, useCallback, useEffect, useMemo, useState } from 'react';
+import useEventCallback from '@jwp/ott-hooks-react/src/useEventCallback';
+import { logDev } from '@jwp/ott-common/src/utils/common';
+
+import scrollbarSize from '../../utils/dom';
+
+type Modal = {
+  modalId: string;
+  modalRef: RefObject<HTMLElement>;
+  onClose?: () => void;
+  containerRef?: RefObject<HTMLElement>;
+  lastFocus: Element | null;
+};
+
+type ModalContextValue = {
+  openModal: (modalId: string, modalRef: RefObject<HTMLElement>, onClose?: () => void, containerRef?: RefObject<HTMLElement>) => void;
+  closeModal: (modalId: string, notify?: boolean) => void;
+  closeAllModals: (notify?: boolean) => void;
+  focusActiveModal: () => void;
+  modals: Modal[];
+};
+
+export const ModalContext = createContext<ModalContextValue>({
+  openModal() {
+    throw new Error('Not implemented');
+  },
+  closeModal() {
+    throw new Error('Not implemented');
+  },
+  closeAllModals() {
+    throw new Error('Not implemented');
+  },
+  focusActiveModal() {
+    throw new Error('Not implemented');
+  },
+  modals: [],
+});
+
+const byId = (modalId: string) => (modal: Modal) => modal.modalId === modalId;
+const notById = (modalId: string) => (modal: Modal) => modal.modalId !== modalId;
+
+const focusModal = (openedModal: Modal, targetElement?: Element | null) => {
+  const { modalRef } = openedModal;
+
+  requestAnimationFrame(() => {
+    if (!modalRef.current) return;
+
+    // prefer focusing the targetElement
+    if (targetElement && targetElement instanceof HTMLElement && modalRef.current.contains(targetElement)) {
+      return targetElement.focus();
+    }
+
+    // find the first interactive element
+    const interactiveElement = modalRef.current.querySelectorAll('input, a, button, [tabindex]')[0] as HTMLElement | null;
+
+    if (!interactiveElement) {
+      logDev('Failed to focus modal contents', { openedModal, targetElement });
+    }
+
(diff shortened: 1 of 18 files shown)


packages/ui-react/src/containers/ModalProvider/useModal.ts
```diff
@@ -0,0 +1,49 @@
...
+  const closeModalEvent = useEventCallback(() => {
+    if (isOpen) {
+      closeModal(modalId, false);
+    }
+  });
+
+  useEffect(() => {
+    // react when the `open` prop changes
+    openModalEvent();
+  }, [open, openModalEvent]);
```
AntonLantukh, line 35: Do you think it could be valuable if we use `open` or `show` just for the initial render (and assume the default value is true). If it is `false` by default, then we can call `openModal` method from the context later with some readable id. In this case there is no limitation that we can open / close modal window just from the modal itself:

```
<Sidebar containerRef={containerRef} id="sidebar" />
```

Inner hook for Modal / Dialog / Alert:

```
export const useInnerModal = ({ modalId, show = true, onClose, modalRef, containerRef }: Params) => {
  const { openModal, closeModal, modals, registerModal, removeModal } = useContext(ModalContext);
  const isOpen = modals.some((modal) => modal.modalId === modalId);
    
  useEffect(() => {
    registerModal(modalId, modalRef, onClose, containerRef); // can be separate state of registered entities. When we register a modal, we also check that id is unique.
  
    if (show) {
        openModal(modalId);
    }
    
    return () => removeModal(modalId); 
  }, []);
  
  return {
    isOpen,
    closeModal,
    openModal,
  };
}
```

Additionally we can have one more hook just to be able to use modal actions from the context (we can also address the context directly):

```
export const useModalActions = () => {
  const { openModal, closeModal } = useContext(ModalContext);

   return {openModal, closeModal}
  });

```

packages/ui-react/src/containers/ModalProvider/ModalProvider.tsx
```diff
@@ -0,0 +1,225 @@
...
+      originFocus.focus({ preventScroll: true });
+    } else if (document.activeElement instanceof HTMLElement) {
+      logDev('Failed to restore focus', { closedModal, originFocus });
+      document.activeElement.blur();
+      window.focus();
+    }
+  });
+};
+
+let originFocusElement: HTMLElement | null = null;
```
AntonLantukh, line 96: Should this one be part of the internal context state?

packages/ui-react/src/containers/ModalProvider/useModal.ts
ChristiaanScheermeijer: While creating the ModalManager, at one point, I had it exactly like that. But after a while, I felt that most logic in the ModalProvider became much more complex due to the following:

- The order of opening matters. To close the last opened modal, the stack needs to be sorted or time-stamped
- Currently, it's a bit funky that both states control the `open` state, I'd prefer to have a single state responsible for the open state, but this would mean a big refactor of the internal state and click actions (it's doable though)
- When we add the possibility to open modals using the context, we need to pass an `onOpen` callback so the state can be updated

Since we currently don't really need the benefit of opening modals using the context, I would like to keep this PR simple.

packages/ui-react/src/containers/ModalProvider/ModalProvider.tsx
ChristiaanScheermeijer: It's not needed in the context, and I want to prevent unnecessary state updates (especially in the `useEffect` that sets this variable).

packages/ui-react/src/containers/ModalProvider/useModal.ts
AntonLantukh: I was thinking about having two stacks: 1 for registered windows with ids (or simply dictionary), 1 for opened windows. This way second one will behave the same way it does now. I am not sure about `onOpen` callback. In case we use `openModal`, then `isOpen` will be updated automatically.

My main concern is actually this part where we use several open states:

```
  const isOpen = modals.some((modal) => modal.modalId === modalId);

  // replace with React `useEffectEvent` when stable: https://react.dev/reference/react/experimental_useEffectEvent
  const openModalEvent = useEventCallback(() => {
    if (open) {
      openModal(modalId, modalRef, onClose, containerRef);
    } else if (isOpen) {
      closeModal(modalId, false);
    }
  });
```

and this one where we allow control via incoming props:

```
  useEffect(() => {
    // react when the `open` prop changes
    openModalEvent();
  }, [open, openModalEvent]);
```

So we have different ways now of how we close and open modals:

1. We open them using `open` prop.
2. We close them using callbacks (both for open and close). And we can't do it by settings `open` to false.

Should we just use open / show prop to control the state? This way no methods will be exposed from the hook.

ChristiaanScheermeijer:
Superseded by #505

AntonLantukh:
@ChristiaanScheermeijer I will close this one.

PR #487, opened 2024-04-08
Fix / Read title while searching (accessibility)

langemike:
This change causes the screen reader to announce the outcome of your search query while your type, to provide convenience for screen reader users. You could consider this a pragmatic _typeahead_ implementation.

Because `aria-live="polite" ` is used it is announced _after_ you performed a keystroke.

Good to know:
- I also applied `aria-live`to the `<ErrorPage>` so that title is also announced. Adding this also made me realize it can also be beneficial in other cases where ErrorPage get rendered.
- I needed to prevent the announcement of placeholder data while you type. I did that by applying two `isFetching`-conditions. Though, I think it's more logical to render the placeholder data based on the `isFetching`-state itself instead of `!!query`. Something to improve in the future.

This change is based on feedback we got from our internal accessibility auditor. He noted the following:
> Visually Impaired	Home	Screen reader does not automatically announce the typeahead search results. Screen reader should automatically announce the typeahead search results.

### DIFF
--- a/packages/ui-react/src/pages/Search/Search.tsx
+++ b/packages/ui-react/src/pages/Search/Search.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import { useParams } from 'react-router';
 import { Helmet } from 'react-helmet';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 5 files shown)


langemike:
I forgot to look at the e2e tests. They should be fixed now

royschut (approved):

ChristiaanScheermeijer (approved):

packages/ui-react/src/pages/Search/Search.tsx
```diff
@@ -56,6 +56,14 @@ const Search = () => {
     };
   }, []);
 
+  const title = isFetching
+    ? t('heading')
+    : !query
+    ? t('start_typing')
+    : playlist?.playlist.length
+    ? t('title', { count: playlist.playlist.length, query })
+    : t('no_results_heading', { query });
```
AntonLantukh: @langemike would be great to have a small function to calculate it with simple `if` conditions. I am not against ternary operators, would just prefer to limit it to 1 level to improve readability.

ChristiaanScheermeijer: I've replaced it with an `useMemo`, so we won't need an IIFE or pass `t` to a util.

AntonLantukh (approved):

PR #493, opened 2024-04-10
Fix / Cleeng PayPal invalid url error (clean)

ChristiaanScheermeijer:
## Description

See #492 for more information. This PR simply fixes the Cleeng PayPal error but keeps the flow intact.

### DIFF
--- a/packages/ui-react/src/utils/location.ts
+++ b/packages/ui-react/src/utils/location.ts
@@ -12,6 +12,6 @@ export const modalURLFromLocation = (location: Location, u: keyof AccountModals 
 };
 
 // Create a full modal url including hostname, mostly for external use (e.g paypal successUrl)
-export const modalURLFromWindowLocation = (u: keyof AccountModals, queryParams?: QueryParamsArg) => {
+export const modalURLFromWindowLocation = (u: keyof AccountModals | null, queryParams?: QueryParamsArg) => {
   return createURL(window.location.href, { u, ...queryParams });
 };
--- a/packages/ui-react/src/utils/location.ts
+++ b/packages/ui-react/src/utils/location.ts
@@ -11,7 +11,12 @@ export const modalURLFromLocation = (location: Location, u: keyof AccountModals 
   return createURL(`${location.pathname}${location.search || ''}`, { u, ...queryParams });
 };
 
-// Create a full modal url including hostname, mostly for external use (e.g paypal successUrl)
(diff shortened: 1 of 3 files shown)


packages/ui-react/src/utils/location.ts
```diff
@@ -12,6 +12,6 @@ export const modalURLFromLocation = (location: Location, u: keyof AccountModals
 };
 
 // Create a full modal url including hostname, mostly for external use (e.g paypal successUrl)
-export const modalURLFromWindowLocation = (u: keyof AccountModals, queryParams?: QueryParamsArg) => {
+export const modalURLFromWindowLocation = (u: keyof AccountModals | null, queryParams?: QueryParamsArg) => {
```
AntonLantukh: Is it still modal URL when the `u` param is not defined? Or should we have a separate function for this purpose?

ChristiaanScheermeijer: The goal of explicitly passing `null` to this function is to clear the modal parameter from the URL. We could create a separate util for this, but that would essentially do the same:

```tsx
export const closeModalURLFromWindowLocation = (queryParams?: QueryParamsArg) => {
  return createURL(window.location.href, { ...queryParams, u: null });
}
```

Alternatively, we could create a `createURLFromWindowLocation` util (following the convention) and empty the `u` param from the callee.

```tsx
const successUrlPaypal = offerType === 'svod' ? waitingUrl : createURLFromWindowLocation({ u: null });
```

langemike (approved):
Seems like a solid approach to me

packages/ui-react/src/utils/location.ts
royschut: Or you can use the `createURLFromLocation` util (we do the same in the AccountModal useEffect):
```javascript
const successUrlPaypal = offerType === 'svod' ? waitingUrl : createURLFromLocation(location, { u: null });
```

royschut: Or you could use `createURL(window.location.href, { u: null });` from Checkout.

AntonLantukh: `createURL(window.location.href, { u: null })` looks good. We can have a small function for this. It is just a bit confusing to me now that we build a modal url without any modal.

ChristiaanScheermeijer: Updated

AntonLantukh (approved):

ChristiaanScheermeijer:
@AntonLantukh @dbudzins can we merge this PR? 😄

PR #494, opened 2024-04-11
Test: add already subscribed/canceled check

ChristiaanScheermeijer:
## Description

Sometimes, the e2e tests fail due to a timeout on the waiting for subscription step. When this test retries, it often happens that the subscription is active.

I've added a simple check to skip the subscribe steps and only validate when the subscription is already active.

The same applies to the cancel subscription step.

My only worry is what if something didn't work as expected. Like a failure in the subscribe flow, but the subscription still ends up being active.

**Edit**

I forgot to mention that I split the subscription and payment tests to make the test more efficient running with workers because workers run a single test file.


platforms/web/test-e2e/tests/payments/payments_test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+const cleengProps: ProviderProps = {
+  config: testConfigs.svod,
+  monthlyOffer: constants.offers.monthlyOffer.cleeng,
+  yearlyOffer: constants.offers.yearlyOffer.cleeng,
+  paymentFields: constants.paymentFields.cleeng,
+  creditCard: constants.creditCard.cleeng,
+  applicableTax: 21,
+  canRenewSubscription: true,
+  canOpenReceipts: false, // Cleeng returns an error on Sandbox making this test flaky
```
royschut: `constants.canShowReceipts` is now `false` by default for Cleeng (not supported), so I think we can remove this comment.

royschut (approved):
Nice @ChristiaanScheermeijer!
This PR also fixes #483 which I have now re-opened and based on this one.

platforms/web/test-e2e/tests/payments/subscription_test.ts
```diff
@@ -41,111 +40,13 @@ function runTestSuite(props: ProviderProps, providerName: string) {
...
-
-    I.waitForText("Clicking 'continue' will bring you to the PayPal site.");
-    I.dontSee('Card number');
-    I.dontSee('Expiry date');
-    I.dontSee('Security code');
-    I.waitForText('Continue');
+    I.useConfig(props.config);
   });
 
   Scenario(`I can open the PayPal site - ${providerName}`, async ({ I }) => {
```
AntonLantukh: Looks like it is about payment now right? We check open in the "payments" suite.

platforms/web/test-e2e/tests/payments/payments_test.ts
```diff
@@ -0,0 +1,228 @@
...
+    // Still see payment info
+    cardInfo.forEach(([label, value]) => I.seeInField(label, value));
+  });
+
+  Scenario(`I can renew my subscription - ${providerName}`, async ({ I }) => {
+    if (props.canRenewSubscription) {
+      paidLoginContext = await I.registerOrLogin(paidLoginContext);
+      renewPlan(I, addYear(today), props.yearlyOffer.price);
+    }
+  });
```
AntonLantukh, line 104: Should we duplicate them in two suites?

platforms/web/test-e2e/tests/payments/subscription_test.ts
AntonLantukh: @ChristiaanScheermeijer do you think we can clean up some scenarios?

ChristiaanScheermeijer: Yes, I forgot to remove this one! I think this is the only duplicate scenario

ChristiaanScheermeijer: Oh, that's not right there are more. I intended to split the scenarios to make the test run faster parallel. Will update this

ChristiaanScheermeijer: PR is updated. I removed the same scenarios in both files after splitting it into two files 😅

AntonLantukh: "I can see my payments data", "I can choose an offer", "I can see offered subscriptions", and "I can see payment types" are still duplicated. Is it expected?

ChristiaanScheermeijer: The duplicated scenarios are removed now... this time for real!

platforms/web/test-e2e/tests/payments/payments_test.ts
ChristiaanScheermeijer: I removed the setting and test

AntonLantukh:
@ChristiaanScheermeijer I see you wanted to split them to increase the efficiency, but it still looks to me that semantically they are part of the same file. The difference I see is that one is about "actions" with subscriptions, the other one is about seeing the list of options. Any naming ideas? Like "offers" and "management" while changing the folder name to "subscriptions"?

AntonLantukh (approved):

PR #495, opened 2024-04-11
Test / fix e2e overlapping reports

ChristiaanScheermeijer:
## Description

Because we use a function to wrap the JWP and Cleeng scenarios, the CodeceptJS run workers' reporting fails. The reporter instance exists only once for each worker and it creates two Allure runs within the same instance because of the duplicate ``Feature(`subscription - ${providerName}`)`` calls.

I've played with [data-driven tests](https://codecept.io/advanced/#data-driven-tests), but this will run all steps sequentially.

The easiest fix was to either remove the `Feature` and `Before` calls from the `runTestSuite` functions so it only creates 1 test run.

After this change, the Allure reporting will show all the scenarios before this was the last one completed.

image

https://github.com/codeceptjs/allure-legacy/blob/8c31bd2d0736487edaa7d92a813562b2571cef49/index.js#L180-L182

Calls:

```js
Allure.prototype.startSuite = function(suiteName, timestamp) {
    this.suites.unshift(new Suite(suiteName, timestamp));
};
```

```js
Allure.prototype.endSuite = function(timestamp) {
    var suite = this.getCurrentSuite();

    suite.end(timestamp);
    if(suite.hasTests()) {
        writer.writeSuite(this.options.targetDir, suite);
    }
    this.suites.shift();
};
```

The culprit... This function only returns the first suite:

```js
Allure.prototype.getCurrentSuite = function() {
    return this.suites[0];
};
```

### DIFF
--- a/platforms/web/test-e2e/tests/account_test.ts
+++ b/platforms/web/test-e2e/tests/account_test.ts
@@ -14,15 +14,16 @@ const consentCheckbox = 'Yes, I want to receive Blender updates by email';
 const firstName = 'John Q.';
 const lastName = 'Tester';
 
+Feature(`account`).retry(Number(process.env.TEST_RETRY_COUNT) || 0);
+
 runTestSuite(testConfigs.jwpSvod, 'JW Player', 'direct', false);
 runTestSuite(testConfigs.svod, 'Cleeng', 'resetLink', true);
 
 function runTestSuite(config: typeof testConfigs.svod, providerName: string, resetPasswordType: string, canEditEmail: boolean) {
   let loginContext: LoginContext;
 
-  Feature(`account - ${providerName}`).retry(Number(process.env.TEST_RETRY_COUNT) || 0);
-
-  Before(async ({ I }) => {
+  async function beforeScenario(I: CodeceptJS.I) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
     I.useConfig(config);
 
     loginContext = await I.registerOrLogin(loginContext, () => {
@@ -34,9 +35,11 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string, res
 
       I.clickCloseButton();
     });
-  });
+  }
 
   Scenario(`I can see my account data - ${providerName}`, async ({ I }) => {
+    await beforeScenario(I);
+
     I.seeInCurrentUrl(constants.baseUrl);
     await I.openMainMenu();
 
@@ -73,6 +76,8 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string, res
   Scenario(`I can cancel Edit account - ${providerName}`, async ({ I }) => {
     if (!canEditEmail) return;
 
+    await beforeScenario(I);
+
     editAndCancel(I, editAccount, [
       { name: emailField, startingValue: loginContext.email, newValue: 'user@email.nl' },
       { name: passwordField, startingValue: '', newValue: 'pass123!?' },
@@ -81,6 +86,8 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string, res
 
   Scenario(`I get a duplicate email warning - ${providerName}`, async ({ I }) => {
     if (!canEditEmail) return;
+
+    await beforeScenario(I);
 
     editAndCancel(I, editAccount, [
       {
@@ -100,6 +107,8 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string, res
   Scenario(`I get a wrong password warning - ${providerName}`, async ({ I }) => {
     if (!canEditEmail) return;
 
+    await beforeScenario(I);
+
     editAndCancel(I, editAccount, [
       {
         name: emailField,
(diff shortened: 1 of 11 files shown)


ChristiaanScheermeijer:
I was playing with a different idea. 

It is possible to [customise the parallel execution](https://codecept.io/parallel/#custom-parallel-execution) of tests in CodeceptJS. Creating a folder for all integration tests and running these with different configurations would be a nice and scalable solution.

+ platforms/web/test-e2e/tests/integrations
+ platforms/web/test-e2e/tests/integrations/login_test.ts
+ platforms/web/test-e2e/tests/integrations/register_test.ts
+ platforms/web/test-e2e/tests/integrations/payment_test.ts
+ platforms/web/test-e2e/tests/integrations/subscription_test.ts

These suites will have to be split for two integration options:

```ts
const integrations = ['jpw', 'cleeng'];

// Create two configs and inject the integration (name for now)
const configs = integrations.map((integrationName) => ({
  inject: {
    integrationName,
  },
}));

// for each integration, spawn a worker and register the test group containing all integration tests 
// can also be split over multiple workers...
for (const config of configs) {
  const worker = workers.spawn();
  worker.addTests(testGroups[0]);
  worker.addConfig(config);
}
```

The rest of the files are spread over multiple workers like normal and don't need any special configuration.

platforms/web/test-e2e/tests/account_test.ts
```diff
@@ -14,15 +14,16 @@ const consentCheckbox = 'Yes, I want to receive Blender updates by email';
...
 runTestSuite(testConfigs.jwpSvod, 'JW Player', 'direct', false);
 runTestSuite(testConfigs.svod, 'Cleeng', 'resetLink', true);
 
 function runTestSuite(config: typeof testConfigs.svod, providerName: string, resetPasswordType: string, canEditEmail: boolean) {
   let loginContext: LoginContext;
 
-  Feature(`account - ${providerName}`).retry(Number(process.env.TEST_RETRY_COUNT) || 0);
-
-  Before(async ({ I }) => {
+  async function beforeScenario(I: CodeceptJS.I) {
```
budzworthy, line 25: Does it not work to just move the `Before` to the top level where you moved `Feature`? This seems like an otherwise nice, simple enough fix, but I wish we didn't need to copy all the `await beforeScenario` calls. Seems too easy to forget to add when you create a new Scenario.

budzworthy: Oh I see, Before depends on the runTestSuite args...

budzworthy:
> I was playing with a different idea.

@ChristiaanScheermeijer This looks promising and more maintainable to me. Any downsides?

platforms/web/test-e2e/tests/account_test.ts
ChristiaanScheermeijer: I don't like the `beforeScenario` much either, we could also make this a custom step: `I.amLoggedIn(config);`

We also might be able to refactor the `accountContext` to a [shared value](https://codecept.io/parallel/#sharing-data-between-workers). We can then remove that from each test as well.

ChristiaanScheermeijer:
> @ChristiaanScheermeijer This looks promising and more maintainable to me. Any downsides?

I was playing with this yesterday and I got it (almost) working. The run workers script does feel a bit too custom because I have to trick codecept into duplicating the integration tests with a different config.

The biggest downside is that the e2e tests will only test 1 integration when you run them via the UI or with steps (easier to debug). This is because the run-workers script changes the codecept config before spawning them.

This is what I got so far:

I created two integration configurations because injection works via `require` and no inline objects :')

image

The codecept config can either have a predefined `include`, or we need to ignore the tests by default (when running steps or UI):

```ts
    },
  },
  include: {
    I: './utils/steps_file.ts',
    integration: `./configs/jwp`,
  },
  bootstrap: null,
```

The `run-workers` script creates two config groups with overrides:

```ts
// Create two configs and inject the integration (name for now)
const configs = integrations.map((integrationName) => ({
  include: {
    integration: `./configs/${integrationName}`,
  },
}));
```

Currently, 4 workers are spawned, but this needs to be worked out a bit more:

```ts
for (const config of configs) {
  for (const group of testGroups) {
    const worker = workers.spawn();
    worker.addTests(group);
    worker.addConfig(config);
    worker.addOptions({ config: __dirname + '/codecept.desktop.ts', verbose: true });
  }
}
```

The challenge here is to balance everything out. If we only create two workers for the integrations, the rest of the tests are distributed over the remainder of the workers.

However, we could also run the non-integration tests on the integration-configured worker...

ChristiaanScheermeijer:
@dbudzins do you think this idea still has potential? We are looking into optimizing the e2e tests (duration), but this might change based on this outcome.

budzworthy:
> @dbudzins do you think this idea still has potential? We are looking into optimizing the e2e tests (duration), but this might change based on this outcome.

Hm, it's getting a bit fiddly now.... it looks like just splitting the files fixes the reporting issue. It's kinda annoying, but seems like a quick easy fix: https://github.com/jwplayer/ott-web-app/pull/503

For whatever reason it also looks like they still won't run in parallel, but since there are a lot of tests to run, I think this will shake out in the wash.

What do you think?

ChristiaanScheermeijer:
@dbudzins As of now, I'm doubting between keeping it "as is" or proceeding with #503. I can pick up on #503, submit a PR and close this one?

AntonLantukh (approved):
@ChristiaanScheermeijer both approaches ("as is" AND "separate files in 503") are fine for us. You can choose the one which is easier to maintain. Having custom workers logic looks too complex for our case.

AntonLantukh:
@ChristiaanScheermeijer any preference here?

PR #498, opened 2024-04-15
fix: Plans prompt after social register

mirovladimitrovski:
## Description

This PR solves #OWA-73.
Also [SERV-15345](https://jwplayer.atlassian.net/browse/SERV-15345)

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code

[SERV-15345]: https://jwplayer.atlassian.net/browse/SERV-15345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
+++ b/packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
@@ -6,7 +6,6 @@ import type { RegistrationFormData } from '@jwp/ott-common/types/account';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import AccountController from '@jwp/ott-common/src/controllers/AccountController';
 import { checkConsentsFromValues, extractConsentValues, formatConsentsFromValues } from '@jwp/ott-common/src/utils/collection';
-import useSocialLoginUrls from '@jwp/ott-hooks-react/src/useSocialLoginUrls';
 import useForm from '@jwp/ott-hooks-react/src/useForm';
 import { modalURLFromLocation } from '@jwp/ott-ui-react/src/utils/location';
 import { useAccountStore } from '@jwp/ott-common/src/stores/AccountStore';
@@ -55,8 +54,6 @@ const Registration = () => {
     setConsentValues(extractConsentValues(publisherConsents));
   }, [accountController, publisherConsents]);
 
-  const socialLoginURLs = useSocialLoginUrls(window.location.href.split('?')[0]);
-
   const { handleSubmit, handleChange, handleBlur, values, errors, validationSchemaError, submitting } = useForm<RegistrationFormData>({
     initialValues: { email: '', password: '' },
     validationSchema: object().shape({
@@ -98,7 +95,6 @@ const Registration = () => {
       consentValues={consentValues}
       publisherConsents={publisherConsents}
       loading={loading || publisherConsentsLoading}
-      socialLoginURLs={socialLoginURLs}
     />
   );
 };
--- a/packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
+++ b/packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
@@ -6,6 +6,7 @@ import type { RegistrationFormData } from '@jwp/ott-common/types/account';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import AccountController from '@jwp/ott-common/src/controllers/AccountController';
 import { checkConsentsFromValues, extractConsentValues, formatConsentsFromValues } from '@jwp/ott-common/src/utils/collection';
+import useSocialLoginUrls from '@jwp/ott-hooks-react/src/useSocialLoginUrls';
 import useForm from '@jwp/ott-hooks-react/src/useForm';
 import { modalURLFromLocation } from '@jwp/ott-ui-react/src/utils/location';
 import { useAccountStore } from '@jwp/ott-common/src/stores/AccountStore';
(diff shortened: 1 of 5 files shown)


darkoatanasovski (dismissed):

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
```diff
@@ -53,7 +53,7 @@ export default function AppRoutes() {
   useNotifications();
 
   if (userData.user && !userData.loading && window.location.href.includes('#token')) {
-    return <Navigate to="/" />; // component instead of hook to prevent extra re-renders
+    return <Navigate to={`/${location.search}`} />; // component instead of hook to prevent extra re-renders
```
AntonLantukh: Do we also need this change?

packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
```diff
@@ -55,7 +55,9 @@ const Registration = () => {
     setConsentValues(extractConsentValues(publisherConsents));
   }, [accountController, publisherConsents]);
 
-  const socialLoginURLs = useSocialLoginUrls(window.location.href.split('?')[0]);
+  const socialLoginReturnURL = useMemo(() => `${window.location.href.split('/?')[0]}${modalURLFromLocation(location, 'personal-details')}`, [location]);
```
AntonLantukh: I think it should also work:

```

const socialLoginReturnURL = useMemo(() => modalURLFromLocation(location, 'personal-details')
, [location]);

```

mirovladimitrovski: I don't want the rest of the query parameters to be kept (whatever they may be), I want to make sure only `u=personal-details` is kept.

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
mirovladimitrovski: Yes, otherwise being redirected to `u=personal-details#token=AAABBBCCC` wouldn't make a difference if we keep this navigating `to="/"`. Our goal is to open the `personal-details` modal when we're redirected from social media login, meaning we must keep the search params as-is and only remove the `#token` hash from the URL.

packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
AntonLantukh: Then it is still better to use `location.pathname` to avoid exposing window in the component.

mirovladimitrovski: `window` was already there before me making this change, I assumed it's there for a reason (it's also used the same way in the `Login.tsx` component). I assumed the reason was that we're dealing with absolute URL navigation, so we need to be sure we're using the full path, including origin, pathname, and the search params we want included. There are a couple of reasons I didn't want to get into re-evaluating its usage: `location.pathname` doesn't include the protocol prefix `https:` and that may be fine for relative navigation with React Router, but the API may need a full literal URL and I wouldn't risk breaking the functionality if it already works the way it is, and finally keep the focus on what needs to be done incrementally. The other reason is that `modalURLFromLocation` doesn't remove the search params from `location`.
image

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
royschut: Maybe we can use `PATH_HOME` here:
```javascript
createURL(`${PATH_HOME}${location.search}`, {});
```

packages/ui-react/src/containers/AccountModal/forms/Registration.tsx
royschut: You can use `modalURLFromWindowLocation` to include the `https://`.
```javascript
modalURLFromWindowLocation('personal-details'); // https://app-preview.jwplayer.com/?app-config=2basfsic&u=personal-details
```

However, the query params are still kept, but also in your approach when appending `modalURLFromLocation`. Is that really an issue? I can imagine situations where we might still need them.

mirovladimitrovski: @royschut You're right, good catch. I'm simply trying to replicate what occurs right after normal signup/register and do the same after registering with social media, which is:

```
navigate(modalURLFromLocation(location, 'personal-details'));
```

And you're right, it does keep the query params. So I'm changing my mind, `modalURLFromWindowLocation` would suffice.

AntonLantukh (approved):

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
ChristiaanScheermeijer: I'm reading this just now and think if the only reason is to remove the token, we don't need a navigate at all. I also feel this breaks the expectations when a user logs in on a media page, for example.

`/m/1234/movie-i-want-to-watch` -> `social login` -> `back on home page :-(`

budzworthy: @mirovladimitrovski can you test the flow above

mirovladimitrovski: @dbudzins  @ChristiaanScheermeijer I didn't really question the navigate at all because it was already there and I just added the search params on top of it, but you're right about what you're describing and we need to create a ticket for this.

mirovladimitrovski: @dbudzins @ChristiaanScheermeijer Here it is: https://github.com/jwplayer/ott-web-app/pull/499

PR #499, opened 2024-04-16
fix: only remove token from hash after login with social media

mirovladimitrovski:
## Description

This PR solves [OWA-74](https://jwplayer.atlassian.net/browse/OWA-74).

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

[OWA-74]: https://jwplayer.atlassian.net/browse/OWA-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


platforms/web/src/containers/AppRoutes/AppRoutes.tsx
```diff
@@ -54,7 +53,8 @@ export default function AppRoutes() {
   useNotifications();
 
   if (userData.user && !userData.loading && window.location.href.includes('#token')) {
-    return <Navigate to={`${PATH_HOME}${location.search}`} />; // component instead of hook to prevent extra re-renders
+    // leave the location intact and only remove the token from the hash
+    return <Navigate to={{ ...location, hash: '' }} replace />; // component instead of hook to prevent extra re-renders
```
ChristiaanScheermeijer, line 57: 1. You can use the router's location instead of `window.location`
2. No need to pass in the previous location, only set want you want to change
3. Remove the entry from the history by adding `replace` 

```suggestion
  if (userData.user && !userData.loading && location.hash.includes('#token')) {
    // leave the location intact and only remove the token from the hash
    return <Navigate to={{ hash: '' }} replace />; // component instead of hook to prevent extra re-renders
  }
```

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
```diff
@@ -44,13 +65,25 @@ const PersonalDetails = () => {
...
+  const closeModal = useCallback(() => {
+    navigate(modalURLFromLocation(location, null), { replace: true });
+  }, [location, navigate]);
+
   useEffect(() => {
-    if (data && (!data.isCaptureEnabled || !data.shouldCaptureBeDisplayed)) nextStep();
+    if (!data) return;
+
+    if (isPreviouslyRegistered) {
+      closeModal();
```
ChristiaanScheermeijer: This might be tricky. `closeModal` depends on the `location` which changes when the modal closes. This re-reruns the effect again. If you need the `closeModal` in an effect, you might want to use `useEventCallback` to make the reference stable.

```tsx
const closeModal = useEventCallback(() => {
  navigate(modalURLFromLocation(location, null), { replace: true });
})l
```

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
```diff
@@ -33,6 +33,27 @@ const PersonalDetails = () => {
...
+    };
+
+    const enabledFields = data.settings.filter((item) => item.enabled);
+
+    const hasSomeFilled = enabledFields.some((item) => !isEmpty(item));
+
+    const hasNoneRequiredEmpty = !enabledFields.some((item) => item.required && isEmpty(item));
+
+    return hasSomeFilled && hasNoneRequiredEmpty;
+  }, [data]);
```
ChristiaanScheermeijer: This logic is way too specific for this container. Can't we determine this using the `getCaptureStatus` method?

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
ChristiaanScheermeijer: Although `nextStep` is also using this pattern...

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
mirovladimitrovski: 1. It is router's `location`. https://github.com/jwplayer/ott-web-app/pull/499/files/ff89ff6e43f86fda891ca0e89e5b3264852aca62#diff-f5da7eac53f6186c62433bfc665c384dac9bd3ff37eac17a58260b0fa874f806R40
2. I did that at first, but unfortunately I have to include the search params explicitly, otherwise they get excluded. So I have to either go with `{ ...location,` or with `, search: location.search }`. 
3. `replace` is already there, look again. :)

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
mirovladimitrovski: Once the modal gets closed there's no effect to be re-run, the modal component is completely unmounted.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
mirovladimitrovski: The `data` variable used here is exactly that.
https://github.com/jwplayer/ott-web-app/pull/499/files/ff89ff6e43f86fda891ca0e89e5b3264852aca62#diff-ac96434ea99a92eb0ab9c31a8d8887e831830f61510f4cd65351cd302931380aR29

All I'm doing is determining if there's any evidence the user has filled out the `Personal Details` form before, which indirectly is an indicator that they aren't a newly registered user. Additionally I'm checking if there's any field that might have been added in the meantime that's `required` and hasn't been filled out, in which case we don't want this modal to be skipped.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
mirovladimitrovski: My point is that the fact that `closeModal` depends on `location` doesn't make any real difference, because whenever `location` gets changed, this component will no longer be rendered.

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
ChristiaanScheermeijer: 1. I mean the condition: `window.location.href.includes('#token')`
2. Hmm, that's unexpected... You can leave it as is 😄 
3. Whoops, you're right!

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
ChristiaanScheermeijer: Hmmm, then I'm afraid the social login/register user flow isn't quite stable.

The `shouldCaptureBeDisplayed` value reflects whether the personal details form should be shown due to an update in the capture fields or the user hasn't filled in anything yet. The Cleeng API determines this for us.

The JWP service will always return true in this case:

https://github.com/jwplayer/ott-web-app/blob/d6c720271040ee67606e5e678c0179112474bd3a/packages/common/src/services/integrations/jwp/JWPAccountService.ts#L396-L399

Does it make more sense to do the check here and leverage the existing flow?

The reason I bring this up is because we are trying to extract most business logic to service/controllers, and this change re-adds some integration-specific logic into the UI.

mirovladimitrovski: That's a good suggestion, I'll look into it right away.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
mirovladimitrovski: I'm resolving this discussion as obsolete, as I removed `closeModal` in my last commit.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
mirovladimitrovski: @ChristiaanScheermeijer please take a look at my last commit:
https://github.com/jwplayer/ott-web-app/pull/499/commits/a9d3630ed790d482bb038f1eecfd74b3ee68fa9e

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
```diff
@@ -142,6 +147,7 @@ const PersonalDetails = () => {
   const { setValue, handleSubmit, handleChange, values, errors, validationSchemaError, submitting } = useForm<PersonalDetailsFormData>({
     initialValues,
     onSubmit: PersonalDetailSubmitHandler,
+    isReadyForInit: !isLoading,
```
AntonLantukh: Maybe the form init should just happen when the whole data is available? Somewhere in the child component which is rendered after everything is loaded.

mirovladimitrovski: Good idea.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
```diff
@@ -45,16 +46,20 @@ const PersonalDetails = () => {
...
+    if (questions) {
       setQuestionValues(Object.fromEntries(questions.map((question) => [question.key, ''])));
     }
   }, [data, nextStep, questions]);
 
   const initialValues: PersonalDetailsFormData = {
-    firstName: '',
-    lastName: '',
+    firstName: (fields.firstNameLastName?.answer as Record<string, string | null>)?.firstName ?? '',
+    lastName: (fields.firstNameLastName?.answer as Record<string, string | null>)?.lastName ?? '',
```
ChristiaanScheermeijer: Very nice changes @mirovladimitrovski! This looks clean

I'm curious (not for now) if we can move all `useForm` hooks to the component like this. This ensures that the `initialValues` are defined and makes a lot of props redundant. Initially, I thought the form logic belonged to the container level, but this seems fine. 

The only feedback remaining is the casting in the above code. It might be minor because we should be able to assume that `fields.firstNameLastName.answer` is an object with multiple values, I prefer to ensure that's the case.

In a refactor PR we did that has not landed yet, I only use `CustomFormField` types, which makes this a bit simpler, but perhaps we can add a simple util that resolves/extracts the value from a union type.

```ts
  const getCaptureValue = (field?: CleengCaptureField | CleengCaptureQuestionField, answerKey?: string) => {
    if (typeof field?.answer === 'string') return field.answer;
    if (typeof field?.answer === 'object' && field.answer !== null && !!answerKey) return field.answer[answerKey] ?? '';
    return '';
  };
```

```suggestion
    firstName: getCaptureValue(fields.firstNameLastName, 'firstName'),
    lastName: getCaptureValue(fields.firstNameLastName, 'lastName'),
```

ChristiaanScheermeijer (approved):

ChristiaanScheermeijer:
Don't forget to squash and give the commit(s) a nice name for the changelogs 😄 I think the changelog parser skips the `OWA-<num>: message` commits currently in the develop branch.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
mirovladimitrovski: Yeah, I don't like the casting either, especially that it's the same casting written twice. I reduced it to one statement in my latest commit. I think we're good to go for now.

AntonLantukh (approved):
Nice changes, thank you!

AntonLantukh (approved):

PR #500, opened 2024-04-16
Fix / accessibility bugs and form validation

MelissaDTH:
## Description

This pull request includes recent accessibility improvements implemented in the [Videodock/ott-web-app](https://github.com/Videodock/ott-web-app) fork. It includes validation improvements and an update to the `IconButton` component, see below:

### Implemented fixes
- Enhances form validation on the Account page
- Fixes an issue with the old password form where errors persisted after validation
- Updates the `IconButton` component to utilise a `<button>` element instead of a `<div>` element, this improves its accessibility by providing a more semantically correct representation of a clickable button, ensuring that aria attributes function as intended
    - Also ensures adherence to button styling and prevents them from acting as a submit button after the above fix    

### Commits
- fix(project): improve form validation on account page
- fix(project): old password form error after validation
- fix(a11y): icon button to use button element
- chore(a11y): adhere icon button styling & don’t act as a submit button
- fix(a11y): focus not returned when closing the main menu using VoiceOver
### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


packages/ui-react/src/components/Header/Header.tsx
```diff
@@ -226,14 +226,7 @@ const Header: React.FC<Props> = ({
           {t('skip_to_content')}
         </a>
         <div className={styles.menu}>
-          <IconButton
-            className={styles.iconButton}
-            aria-label={sideBarOpen ? t('close_menu') : t('open_menu')}
```
AntonLantukh, line 231: Do you think we can remove this condition?

packages/ui-react/src/components/TextField/TextField.tsx
```diff
@@ -69,7 +69,7 @@ const TextField: React.FC<Props> = ({
     const inputType = 'type' in otherInputProps ? otherInputProps.type : 'text';
 
     const ariaAttributes = {
-      'aria-invalid': Boolean(error && !value),
+      'aria-invalid': Boolean(error && !!value),
```
AntonLantukh, line 72: Looks like this change led to some snapshot updates where the default error state is set to `true` now. Is it expected?

platforms/web/test-e2e/tests/video_detail_test.ts
```diff
@@ -46,20 +46,20 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string) {
       I.seeCssPropertiesOnElements(`text="${constants.agent327Description}"`, { 'max-height': height });
     }
 
-    I.seeElement('div[aria-label="Expand"]');
-    I.dontSeeElement('div[aria-label="Collapse"]');
+    I.seeElement('button[aria-label="Expand"]');
+    I.dontSeeElement('button[aria-label="Collapse"]');
```
AntonLantukh, line 50: It should probably also work this way:

```
    I.seeElement('[aria-label="Expand"]');
    I.dontSeeElement('[aria-label="Collapse"]');
```

AntonLantukh:
@MelissaDTH could you please take a look? The PR looks good, just some minor things I wanted to clarify.

packages/ui-react/src/components/TextField/TextField.tsx
MelissaDTH: Yes, that is to be expected! The `DemoConfigDialog` and `PersonalDetailsForm` snapshot files are updated in the instances where errors are intentionally being passed. As a result, `aria-invalid` is now correctly set to `true` for these affected fields, where they previously incorrectly were set to `false`.

packages/ui-react/src/components/Header/Header.tsx
MelissaDTH: Yes, we think so: The `aria-expanded` toggle already indicates whether or not it is “Open menu”, just like we currently do with the “View password” label on the view/hide password toggle.

platforms/web/test-e2e/tests/video_detail_test.ts
MelissaDTH: I've noted this for another moment!

PR #501, opened 2024-04-16
Refactor / Layout and AppRoutes cleanup

ChristiaanScheermeijer:
## Description

We noticed that when adding platforms, we hoisted the AppRoutes and Layout containers to change the routing and main layout with platform-specific changes. However, these currently contain lots of logic and conditionals, which feel redundant and error-prone when syncing upstream changes.

I might have gone too far with this PR, but the idea was to split more logic into smaller containers. The `Header` component was way too big and received way too many props. I could simply wrap the Header component in a container, but then we still need to hoist all this logic when making a small change to the header for a platform. For example, when we don't want the user menu or actions.

The size of the AppRoutes and Layout containers is an essential part of this PR:

https://github.com/jwplayer/ott-web-app/blob/78ec79bc2ed0a28fd28b5050faf799a1364e4714/platforms/web/src/containers/AppRoutes/AppRoutes.tsx

https://github.com/jwplayer/ott-web-app/blob/78ec79bc2ed0a28fd28b5050faf799a1364e4714/packages/ui-react/src/containers/Layout/Layout.tsx

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [ ] Reviewed own code


packages/ui-react/src/containers/Layout/Layout.tsx
```diff
@@ -1,224 +1,82 @@
...
   const banner = assets.banner;
-  const canLogin = accessModel !== ACCESS_MODEL.AVOD;
 
-  useEffect(() => {
-    if (isLoggedIn && profilesEnabled && !profiles?.length) {
-      profileController?.unpersistProfile();
-    }
-    // Trigger once on the initial page load
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
```
ChristiaanScheermeijer, line 90: @AntonLantukh do you remember why this was needed? Can we move this to the controller instead?

AntonLantukh: I don't really remember...It was right in the component, all I did was wrapping it into the `useEffect` hook. I think it may be connected with the case when you have a profile saved in the local storage but no profile on the back-end side. So probably this way we can solve the inconsistency problem.

langemike:
Moving forward, a change like this is necessary. I see some minor improvements we could apply in the CSS selectors, but that can always be improved upon in a later phase. I hope we can merge this ASAP.

langemike (approved):

packages/ui-react/src/containers/Layout/Layout.tsx
AntonLantukh: @kiremitrov123 maybe you know why this part was needed?

ChristiaanScheermeijer: My guess is that this is only needed when switching configs (between integration with and without profiles).

packages/ui-react/src/components/Header/HeaderActionButton.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import IconButton, { type Props as IconButtonProps } from '../IconButton/IconButton';
+
+import styles from './Header.module.scss';
+
+const HeaderActionButton = ({ children, className, ...rest }: IconButtonProps) => {
```
AntonLantukh, line 8: Do you think we need this component? Looks like the only difference is in the additional class.

packages/ui-react/src/containers/Layout/Layout.tsx
```diff
@@ -1,224 +1,82 @@
...
+          <HeaderSkipLink />
+          <HeaderMenu sideBarOpen={sideBarOpen} onClick={openSideBar} />
+          <HeaderBrand siteName={siteName} logoSrc={banner} setLogoLoaded={() => undefined} />
+          <HeaderNavigation navItems={navItems} />
+          <HeaderActions>
+            <HeaderSearch />
+            <HeaderLanguageSwitcher />
+            <HeaderUserMenu />
+          </HeaderActions>
+        </Header>
```
AntonLantukh, line 73: Should we probably do it right in the Header.tsx component?

platforms/web/package.json
```diff
@@ -35,8 +35,7 @@
     "react-i18next": "^12.3.1",
     "react-router": "6.14.2",
     "react-router-dom": "6.14.2",
-    "wicg-inert": "^3.1.1",
-    "zustand": "^3.6.9"
```
AntonLantukh, line 39: Do we remove it here because we already have it in `common`?

packages/ui-react/src/containers/SiteMetadata/SiteMetadata.tsx
```diff
@@ -0,0 +1,24 @@
+import { Helmet } from 'react-helmet';
+import React from 'react';
+import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
+import { shallow } from '@jwp/ott-common/src/utils/compare';
+import { useTranslation } from 'react-i18next';
+
+const SiteMetadata = () => {
```
AntonLantukh, line 7: Should we move it to components? Doesn't look that this one is used as container.

packages/ui-react/src/containers/HeaderLanguageSwitcher/HeaderLanguageSwitcher.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useCallback, useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
+import { useUIStore } from '@jwp/ott-common/src/stores/UIStore';
+
+import LanguageMenu from '../../components/LanguageMenu/LanguageMenu';
+
+const HeaderLanguageSwitcher = () => {
```
AntonLantukh, line 8: Should we make it to Header folder in components?

packages/ui-react/src/containers/HeaderSearch/HeaderSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+import CloseIcon from '@jwp/ott-theme/assets/icons/close.svg?react';
+
+import Icon from '../../components/Icon/Icon';
+import SearchBar from '../../components/SearchBar/SearchBar';
+import useSearchQueryUpdater from '../../hooks/useSearchQueryUpdater';
+import HeaderActionButton from '../../components/Header/HeaderActionButton';
+
+import styles from './HeaderSearch.module.scss';
+
+const HeaderSearch = () => {
```
AntonLantukh, line 17: Should we also make it part of the Header component?

packages/ui-react/src/components/UserMenu/UserMenu.tsx
```diff
@@ -1,104 +1,89 @@
...
+              <ProfilesMenu
+                onButtonClick={onClose}
+                profiles={profiles ?? []}
+                currentProfile={profile}
+                selectingProfile={profileLoading}
+                selectProfile={onSelectProfile}
+                small
+              />
+            )}
+            <UserMenuNav focusable={open} onButtonClick={onClose} showPaymentItems={true} currentProfile={profile} favoritesEnabled={favoritesEnabled} small />
```
AntonLantukh, line 82: Should we use a dynamic value for `showPaymentItems`?

packages/ui-react/src/components/Header/HeaderActionButton.tsx
ChristiaanScheermeijer: This is needed for component composition. Because this uses the `styles.iconButton` from the `Header.module.scss` file, it would be weird to import the header styles instead.

It's not much, but it does make it clear

packages/ui-react/src/containers/Layout/Layout.tsx
ChristiaanScheermeijer: That would defeat the purpose. We will have separate copies of the Layout component for each platform. With this, we can change the composition by adding/removing JSX instead of copying all files that need to be changed.

platforms/web/package.json
ChristiaanScheermeijer: `zustand` was only used in the AppRoutes for the profiles logic. After moving that logic to `ui-react` knip marked that dependency unused

packages/ui-react/src/containers/SiteMetadata/SiteMetadata.tsx
ChristiaanScheermeijer: `useConfigStore` is used 😄

packages/ui-react/src/containers/HeaderLanguageSwitcher/HeaderLanguageSwitcher.tsx
ChristiaanScheermeijer: This is a container (wrapping the `LanguageMenu` component)

packages/ui-react/src/containers/HeaderSearch/HeaderSearch.tsx
ChristiaanScheermeijer: This also is a container (although this one also contains UI...). The UI can be split into a component in the Header folder, but the container remains. We probably want to make a better/stricter convention for this

packages/ui-react/src/components/UserMenu/UserMenu.tsx
ChristiaanScheermeijer: It was already there 😅

packages/ui-react/src/containers/Layout/Layout.tsx
AntonLantukh: Thanks, makes sense now. Missed the part where you mentioned customisations for different platforms.

packages/ui-react/src/containers/HeaderSearch/HeaderSearch.tsx
AntonLantukh: Okay I see your point now, using components for all stateless UI and containers for stateful. My concern with the header is that it is split into several folders now across components and containers.

Maybe `features` folder could help us in the future where we would group some connected containers/components (Header (or Layout for Cards, Header and Footer), Form, EPG, Payments, Subscriptions ...)? Components could be just for something we reuse across features (maybe also stateful). Just thinking out loud now.

AntonLantukh (approved):

AntonLantukh:
@ChristiaanScheermeijer please resolve the conflicts and I will merge it.

packages/ui-react/src/containers/HeaderSearch/HeaderSearch.tsx
ChristiaanScheermeijer: Thanks, Anton! Yes, I remember @dbudzins also suggesting something similar. Let's discuss this in the near future! 😄

ChristiaanScheermeijer:
@AntonLantukh this branch is rebased and squashed (making it easier to rebase)

PR #502, opened 2024-04-16
Fix / lose focus after activating share button

ChristiaanScheermeijer:
## Description

When activating the share button, the focus is lost because of the select and copy action. This change stores the document activeElement and restoring focus after the copy command.

> Should we be using Clipboard APIs? 🤔 

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/packages/ui-react/src/utils/clipboard.test.ts
+++ b/packages/ui-react/src/utils/clipboard.test.ts
@@ -1,0 +1,28 @@
+import { copyToClipboard } from './clipboard';
+
+document.execCommand = vi.fn();
+const execCommandMocked = vi.mocked(document.execCommand);
+
+describe('clipboard', () => {
+  test('copies text to clipboard', () => {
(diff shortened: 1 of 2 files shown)


langemike (approved):
LGTM!

royschut (approved):

AntonLantukh:
@ChristiaanScheermeijer `clipboard.writeText` looks pretty safe, any concerns?

ChristiaanScheermeijer:
> @ChristiaanScheermeijer `clipboard.writeText` looks pretty safe, any concerns?

It should be safe and use [transient permissions](https://developer.mozilla.org/en-US/docs/Web/Security/User_activation), which means the user has to interact with the page.

But changing the implementation is not in the scope of this fix 😄

PR #505, opened 2024-04-18
Refactor / replace modal with HTML Dialog

ChristiaanScheermeijer:
## Description

This PR is an alternative to #485. In the modal manager PR, we try to manage everything ourselves, which gives us ultimate control but is risky and needs to be maintained in the future.

The [HTML Dialog](https://caniuse.com/dialog) element has been out for some time now, but I actually never used it because of the browser support. But looking at the support table, I don't see why we shouldn't use it. Over 96% of the global browsers support the HTML Dialog element.

Using the native dialog element removes the need for the ModalProvider, because we can use the DOM to close open modals.

Only the body scrolling was still possible so I modified the `useModal` hook to hide the scrollbars when a modal is open.

Although we don't have to manage focus anymore, I did find one edge case when opening the account modal from the sidebar. Because the last focused element doesn't exist anymore (I fixed this in the ModalProvider), the body receives focus when closing the last modal.

This is a proposal PR, so let me know if you have any feedback or criticism.

### DIFF
--- a/packages/hooks-react/src/useOffers.ts
+++ b/packages/hooks-react/src/useOffers.ts
@@ -1,5 +1,4 @@
-import { useMutation } from 'react-query';
-import { useEffect } from 'react';
+import { useMutation, useQuery } from 'react-query';
 import { shallow } from '@jwp/ott-common/src/utils/compare';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import { useCheckoutStore } from '@jwp/ott-common/src/stores/CheckoutStore';
@@ -21,9 +20,9 @@ const useOffers = () => {
     shallow,
   );
 
-  const { mutate: initialise, isLoading: isInitialisationLoading } = useMutation<void>({
-    mutationKey: ['initialiseOffers', requestedMediaOffers],
-    mutationFn: checkoutController.initialiseOffers,
+  const { isLoading: isInitialisationLoading } = useQuery<void>({
+    queryKey: ['initialiseOffers', requestedMediaOffers],
+    queryFn: checkoutController.initialiseOffers,
   });
 
   const chooseOffer = useMutation({
@@ -36,10 +35,6 @@ const useOffers = () => {
     mutationFn: checkoutController.switchSubscription,
     onSuccess: () => accountController.reloadSubscriptions({ delay: 7500 }), // @todo: Is there a better way to wait?
   });
-
-  useEffect(() => {
-    initialise();
-  }, [requestedMediaOffers, initialise]);
 
   const hasMediaOffers = mediaOffers.length > 0;
   const hasSubscriptionOffers = subscriptionOffers.length > 0;
--- a/packages/ui-react/src/components/Sidebar/__snapshots__/Sidebar.test.tsx.snap
+++ b/packages/ui-react/src/components/Sidebar/__snapshots__/Sidebar.test.tsx.snap
@@ -1,102 +1,71 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`<SideBar /> > renders sideBar closed 1`] = `
-<div>
-  <div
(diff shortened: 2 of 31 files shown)


ChristiaanScheermeijer:
^ not sure what happened here 🤔

packages/hooks-react/src/useOffers.ts
```diff
@@ -37,10 +36,6 @@ const useOffers = () => {
     onSuccess: () => accountController.reloadSubscriptions({ delay: 7500 }), // @todo: Is there a better way to wait?
   });
 
-  useEffect(() => {
-    initialise();
-  }, [requestedMediaOffers, initialise]);
-
```
ChristiaanScheermeijer, line 43: @royschut this `useEffect` caused an extra re-render (unit tests found this), switching the `useMutation` to `useQuery` should have the same effect because the key depends on the `requestedMediaOffers` as well. 

What do you think?

royschut: Sure, that should work, and simplifies as well. 👍

ChristiaanScheermeijer:
@AntonLantukh @dbudzins, what do you think of this PR as an alternative to #485?

I do have to test this (and the polyfill) for older browsers (TV browsers), but if you have a strong preference over this or #485, please let me know 😄

packages/ui-react/src/components/Sidebar/__snapshots__/Sidebar.test.tsx.snap
```diff
@@ -1,104 +1,5 @@
...
-      >
-        <span>
-          Home
-        </span>
-      </a>
-    </nav>
-  </div>
-</div>
-`;
+exports[`<SideBar /> > renders sideBar closed 1`] = `<div />`;
```
AntonLantukh: Hm, all the content got removed. Is it expected?

AntonLantukh (dismissed):
@ChristiaanScheermeijer I like this approach. Do you think it will also work with TV devices? If it does, then I would proceed with this one.

ChristiaanScheermeijer:
> @ChristiaanScheermeijer I like this approach. Do you think it will also work with TV devices? If it does, then I would proceed with this one.

We're testing this, but currently, the web app isn't working at all on (older) TV devices. PR #519 fixes this problem :-)

packages/ui-react/src/components/Sidebar/__snapshots__/Sidebar.test.tsx.snap
ChristiaanScheermeijer: Yes, the Sidebar is now rendered in a modal/dialog which doesn't render when it is hidden.

ChristiaanScheermeijer: But... I found that the Sidebar open snapshot was also empty. This is because there is nothing rendered in the container due to the usage of portals in the Modal component. This is fixed now.

ChristiaanScheermeijer:
I've tested this on Chrome 68 (which we will support after #519), and it seems to work fine. The HTML dialog element is supported from Chrome 37 and up.

royschut:
We have performed our manual testing workflow, including this PR and found 2 bugs that I have fixed just now:
- When the player unmounts, an infinite render loop was initiated. The useCallback wrapper fixes this.
- A click on the backdrop didn't close the modals anymore. Fixed by a simple click handler that checks if the modal itself (which can only be the backdrop) is clicked.

AntonLantukh:
@royschut @ChristiaanScheermeijer looks good to me! I want to merge this PR if all the tests pass. Any objections?

UPD: tests seem to fail, could you take a look?

ChristiaanScheermeijer:
Hi @AntonLantukh, thanks! I've fixed the e2e issues. This PR is ready to be (squashed) merged 😄

PR #508, opened 2024-04-22
Feat / subscription retry logic

royschut:
(Copy of #483, which was automatically closed when its base PR (#494) was merged, and I couldn't reopen)

## Access check retry mechanism
This PR adds a retry mechanism to the access check (`reloadActiveSubscription()`), to make sure that the UI notices granted access, even if the backend takes longer to process a purchase. This also makes it possible to reduce the initial delay. Note: I've left the initial delay after a subscription switch on `3000`ms, because a switch presumably always takes long (it was initially set to `7500`).

### DIFF
--- a/packages/hooks-react/src/useOffers.ts
+++ b/packages/hooks-react/src/useOffers.ts
@@ -34,7 +34,7 @@ const useOffers = () => {
   const switchSubscription = useMutation({
     mutationKey: ['switchSubscription'],
     mutationFn: checkoutController.switchSubscription,
-    onSuccess: () => accountController.reloadSubscriptions({ delay: 7500 }), // @todo: Is there a better way to wait?
+    onSuccess: () => accountController.reloadSubscriptions({ delay: 3000, retry: 10 }), // A subscription switch usually takes at least 3 secs
   });
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


ChristiaanScheermeijer (approved):

packages/hooks-react/src/useOffers.ts
```diff
@@ -34,7 +34,7 @@ const useOffers = () => {
   const switchSubscription = useMutation({
     mutationKey: ['switchSubscription'],
     mutationFn: checkoutController.switchSubscription,
-    onSuccess: () => accountController.reloadSubscriptions({ delay: 7500 }), // @todo: Is there a better way to wait?
+    onSuccess: () => accountController.reloadSubscriptions({ delay: 3000, retry: 10 }), // A subscription switch usually takes at least 3 secs
```
AntonLantukh, line 37: @royschut  Will it work properly if we already have an "activeSubscription"? I mean the case here when we switch a subscription. So it looks like this condition `!activeSubscription && !!retry && retry > 0` won't work.

royschut: Good point, probably not. We could fix it using the approach proposed by @dbudzins in #483, but that's too large of a refactor for this moment. I have noted this for another moment.

But until that, for now I see two options:
- We set the initial delay for this case back to 7500, and merge this PR as it is still a step forward
- We set this PR back into Draft for now

I'm fine with both options. :)

AntonLantukh: I am good with the first option. Just wanted to make sure this is something you are aware of.

AntonLantukh (approved):

PR #519, opened 2024-04-25
Feat / Extend minimum browser support

langemike:
We want to support a broader user base by extending our browser support to `chrome68` and up.

Vite internally uses `['es2020', 'edge88', 'firefox78', 'chrome87', 'safari14']`. So the only difference is the supported chrome browsers. See https://vitejs.dev/guide/build#browser-compatibility for reference.

I only needed to write a CSS fallback for the `dvh` unit because it was the only unsupported CSS feature we have ran into.
This change also contains a variable injection fix for `index.html` which has been introduced (with the upgrade to vite 5, I assume).

We also did some internal experiments with [@vitejs/plugin-legacy](https://www.npmjs.com/package/@vitejs/plugin-legacy), but this change has a bigger impact on our bundle/build. If we even want to broaden our browser support further than `chrome68` then `@vitejs/plugin-legacy` would definitely be a viable option.

I've tested this on Chrome 68 using browserstack.

Our original PR: https://github.com/Videodock/ott-web-app/pull/186
Our `@vitejs/plugin-legacy` experiment PR (for reference): https://github.com/Videodock/ott-web-app/pull/182

### DIFF
--- a/packages/ui-react/src/components/Sidebar/Sidebar.module.scss
+++ b/packages/ui-react/src/components/Sidebar/Sidebar.module.scss
@@ -13,6 +13,7 @@
   display: inline-block;
   width: 270px;
   max-width: 90vw;
+  height: 100vh; // Fallback
   height: 100dvh;
   overflow-y: auto;
   background-color: var(--body-background-color);
--- a/platforms/web/src/index.tsx
+++ b/platforms/web/src/index.tsx
@@ -1,10 +1,10 @@
+import 'core-js/es/array/flat-map';
+import 'core-js/es/object/from-entries';
 import React from 'react';
 import { createRoot } from 'react-dom/client';
 import 'wicg-inert';
 import { registerSW } from 'virtual:pwa-register';
 import { configureEnv } from '@jwp/ott-common/src/env';
-import 'core-js/es/array/flat-map';
-import 'core-js/es/object/from-entries';
 
 import './modules/register';
 
--- a/platforms/web/src/index.tsx
+++ b/platforms/web/src/index.tsx
@@ -1,5 +1,4 @@
-import 'core-js/es/array/flat-map';
-import 'core-js/es/object/from-entries';
+import 'virtual:polyfills';
 import React from 'react';
 import { createRoot } from 'react-dom/client';
 import 'wicg-inert';
--- a/platforms/web/src/index.tsx
+++ b/platforms/web/src/index.tsx
@@ -3,6 +3,8 @@ import { createRoot } from 'react-dom/client';
 import 'wicg-inert';
 import { registerSW } from 'virtual:pwa-register';
 import { configureEnv } from '@jwp/ott-common/src/env';
+import 'core-js/es/array/flat-map';
+import 'core-js/es/object/from-entries.js';
 
 import './modules/register';
 
--- a/platforms/web/src/index.tsx
+++ b/platforms/web/src/index.tsx
@@ -4,7 +4,7 @@ import 'wicg-inert';
 import { registerSW } from 'virtual:pwa-register';
 import { configureEnv } from '@jwp/ott-common/src/env';
 import 'core-js/es/array/flat-map';
-import 'core-js/es/object/from-entries.js';
+import 'core-js/es/object/from-entries';
 
 import './modules/register';
 
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -37,12 +37,13 @@ const config: KnipConfig = {
         '@babel/core', // Required peer dependency for babel plugins
         '@types/luxon', // Used in tests
         'babel-plugin-transform-typescript-metadata', // Used to build with decorators for ioc resolution
+        'core-js', // Conditionally imported at build time
         'eslint-plugin-codeceptjs', // Used by apps/web/test-e2e/.eslintrc.cjs
         'luxon', // Used in tests
         'playwright', // Used in test configs
         'sharp', // Requirement for @vite-pwa/assets-generator
         'tsconfig-paths', // Used for e2e test setup
-        'virtual:pwa-register', // Service Worker code is injected at build time,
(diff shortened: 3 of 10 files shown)


AntonLantukh:
@langemike is "Chrome 68" something you need for Smart TV devices?

packages/ui-react/src/components/Sidebar/Sidebar.module.scss
```diff
@@ -29,6 +29,7 @@
   display: none;
   width: 270px;
   max-width: 90vw;
+  height: 100vh; // Fallback
```
AntonLantukh, line 16: Maybe this way?

```
.sidebar{
    height: 100vh;
}

@supports (height: 100dvh) {
  .sidebar{
    height: 100dvh;
  }   
}
```

langemike: The support of this syntax is limited. https://caniuse.com/mdn-css_at-rules_supports_selector eg. chrome68 doesn't support this syntax.

If we going to extend the browser support further, I suggest to let `postcss-preset-env` to take care of the fallbacks. I experimented with this but this `dvh` unit needed a specific plugin and didn't get transformed automatically, unfortunately. 

Since this is the only CSS browser compatibility issue, I vouch for keeping it simple for now without adding a lot of overhead.

langemike:
> @langemike is "Chrome 68" something you need for Smart TV devices?

Correct!

platforms/web/src/index.tsx
```diff
@@ -1,3 +1,5 @@
+import 'core-js/es/array/flat-map';
+import 'core-js/es/object/from-entries';
```
AntonLantukh: @langemike do you think we can import it conditionally based on the environment variable (APP_LEGACY_BROWSER)?

langemike: I would not limit this env-var only to the conditional imports, which makes the approach more close like I originally intended, but I think we should keep it simple for now (at least) and make it more flexible if use-cases of demand for this increases.

AntonLantukh: My main concern here is the "polyfills" file we additionally get when we add Chrome legacy support. Even if we don't need it.

langemike: I revised the approach. Legacy support is now opt-in by providing `APP_LEGACY_BUILD=1` during `yarn build`.

A `polyfills.js` is still created because some dependencies have imported `core-js` internally.

Without legacy support (smaller files):
```
build/public/manifest.json                           0.49 kB │ gzip:   0.29 kB
build/public/index.html                              2.97 kB │ gzip:   0.81 kB
build/public/assets/default_avatar-D3fd0lZl.png     26.96 kB
build/public/assets/style-xyq8hHtg.css             164.70 kB │ gzip:  29.50 kB
build/public/assets/polyfills-BTIFuIxg.js          108.89 kB │ gzip:  43.26 kB │ map:   676.28 kB
build/public/assets/react-Dl1qj1q7.js              143.24 kB │ gzip:  46.11 kB │ map:   351.28 kB
build/public/assets/inplayer-Caq7sMti.js           313.25 kB │ gzip:  92.39 kB │ map:   708.10 kB
build/public/assets/index-BJFB1YLe.js              342.80 kB │ gzip: 103.81 kB │ map: 1,260.17 kB
build/public/assets/vendor-oOs1bjTA.js           1,615.64 kB │ gzip: 431.98 kB │ map: 4,926.55 kB
```

With `APP_LEGACY_BUILD=1` (bigger files):
```
build/public/manifest.json                           0.49 kB │ gzip:   0.29 kB
build/public/index.html                              2.97 kB │ gzip:   0.81 kB
build/public/assets/default_avatar-D3fd0lZl.png     26.96 kB
build/public/assets/style-xyq8hHtg.css             164.70 kB │ gzip:  29.50 kB
build/public/assets/polyfills-DtCB2TJL.js          90.71 kB │ gzip:  37.81 kB │ map:   567.16 kB
build/public/assets/react-Ba7Lyyxn.js              143.28 kB │ gzip:  46.12 kB │ map:   351.34 kB
build/public/assets/inplayer-CHHsr38C.js           313.25 kB │ gzip:  92.39 kB │ map:   708.10 kB
build/public/assets/index-CuUwpo7s.js              343.22 kB │ gzip: 104.00 kB │ map: 1,260.34 kB
build/public/assets/vendor-DQoT3ipD.js           1,615.76 kB │ gzip: 432.01 kB │ map: 4,926.75 kB
```

AntonLantukh: For some reason they are smaller with APP_LEGACY_BUILD, but the approach itself looks good 👍 Thank you.

AntonLantukh (approved):

platforms/web/src/index.tsx
ChristiaanScheermeijer: I think the file listings are switched in the comment above. The added polyfills make the polyfills file larger when `APP_LEGACY_BUILD` is enabled.

ChristiaanScheermeijer:
We discussed the knip ignore and virtual + relative path import via Slack and are planning to make the following (hopefully last) changes:

- Update import to `import 'virtual:polyfills';`
- Restore knip config and add `virtual:polyfills` to the ignoreDependencies
- Update the Vite plugin with:

```ts
    resolveId(id) {
      if (id.includes('virtual:polyfills')) {
        return '\0' + id;
      }
    },
    load(id) {
      if (id.includes('\0virtual:polyfills')) {
        return enabled ? `import './polyfills';` : 'export default {};';
      }
    },
```

langemike:
@ChristiaanScheermeijer I ammeded my previous commit and performed the approach you suggested.

I removed the ignore knip-config and put `virtual:polyfills` as a `peerDependency` to fix Knip's `Unlisted dependencies` warning.

ChristiaanScheermeijer (approved):
🙌

royschut (approved):

PR #530, opened 2024-05-18
Fix / live channels custom screen

ChristiaanScheermeijer:
## Description

This PR introduces two changes and a fix:

**Use `contentType` matching for live events custom screen instead of custom params**

- Add util function for contentType matching
- Remove custom mediaEvent screen from screenMapping.ts file

**Add support for live channels without TV Guide page**

When creating a 24x7 live channel via JWP dashboard the liveChannel contentType is added. This contentType was originally used only for live streams with a TV Guide (EPG). Now, this is only the case when the `liveChannelsId` param is present on the media item.

**Remove live event start date for 24x7 live channels**

The card shows the live event start date for 24x7 live channels because live channels contain the same parameters. Now the card will only show the event start date for live events.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/packages/common/src/utils/media.ts
+++ b/packages/common/src/utils/media.ts
@@ -1,5 +1,7 @@
-import type { Playlist, PlaylistItem } from '../../types/playlist';
+import type { PlaylistItem } from '../../types/playlist';
 import { CONTENT_TYPE } from '../constants';
+
+import { isContentType } from './common';
 
 type RequiredProperties<T, P extends keyof T> = T & Required<Pick<T, P>>;
 
@@ -7,9 +9,6 @@ type DeprecatedPlaylistItem = {
   seriesPlayListId?: string;
   seriesPlaylistId?: string;
 };
-
-export const isPlaylist = (item: unknown): item is Playlist => !!item && typeof item === 'object' && 'feedid' in item;
-export const isPlaylistItem = (item: unknown): item is PlaylistItem => !!item && typeof item === 'object' && 'mediaid' in item;
 
 // For the deprecated series flow we store seriesId in custom params
 export const getSeriesPlaylistIdFromCustomParams = (item: (PlaylistItem & DeprecatedPlaylistItem) | undefined) =>
@@ -22,12 +21,12 @@ export const isLegacySeriesFlow = (item: PlaylistItem) => {
 
 // For the new series flow we use contentType custom param to define media item to be series
 // In this case media item and series item have the same id
-export const isSeriesContentType = (item: PlaylistItem) => item.contentType?.toLowerCase() === CONTENT_TYPE.series;
+export const isSeriesContentType = (item: PlaylistItem) => isContentType(item, CONTENT_TYPE.series);
 
 export const isSeries = (item: PlaylistItem) => isLegacySeriesFlow(item) || isSeriesContentType(item);
 
 export const isEpisode = (item: PlaylistItem) => {
-  return typeof item?.episodeNumber !== 'undefined' || item?.contentType?.toLowerCase() === CONTENT_TYPE.episode;
+  return typeof item?.episodeNumber !== 'undefined' || isContentType(item, CONTENT_TYPE.episode);
 };
 
 export const getLegacySeriesPlaylistIdFromEpisodeTags = (item: PlaylistItem | undefined) => {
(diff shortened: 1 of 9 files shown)


AntonLantukh:
@ChristiaanScheermeijer `VCH.ScheduledStart` can be also set to Live Channels (BCL SaaS).

packages/common/src/utils/media.ts
```diff
@@ -47,5 +46,4 @@ export const getLegacySeriesPlaylistIdFromEpisodeTags = (item: PlaylistItem | un
   return;
 };
 
-export const isLiveChannel = (item: PlaylistItem): item is RequiredProperties<PlaylistItem, 'contentType' | 'liveChannelsId'> =>
-  item.contentType?.toLowerCase() === CONTENT_TYPE.liveChannel && !!item.liveChannelsId;
+export const isLiveChannel = (item: PlaylistItem): item is RequiredProperties<PlaylistItem, 'contentType'> => isContentType(item, CONTENT_TYPE.liveChannel);
```
AntonLantukh, line 49: Should it be part of "liveEvent.ts" file? We could probably also rename `liveEvent.ts` to `live.ts`

packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaLiveChannel/MediaLiveChannel.tsx
```diff
@@ -1,24 +1,25 @@
...
-  if (!liveChannelsId) {
-    return <ErrorPage title="Live channel not found" />;
+  // this live channel is part of a live channels (TV Guide) page
+  if (liveChannelsId) {
+    return <Navigate to={liveChannelsURL(liveChannelsId, data.mediaid)} replace={true}></Navigate>;
   }
 
-  return <Loading />;
+  // this live channel is "just" a media item
+  return <MediaMovie data={data} isLoading={isLoading} />;
```
AntonLantukh, line 22: Maybe "Live" tag can also be added + "Start Watching" button can be checked? Looks close to what we have in `MediaEvent.tsx`.

ChristiaanScheermeijer:
> @ChristiaanScheermeijer `VCH.ScheduledStart` can be also set to Live Channels (BCL SaaS).

That's correct; that's the reason why the live channel start date was shown on the cards. We should assume that when the contentType is set to `liveChannel`, the live stream is 24x7 and always live. When using contentType `liveEvent`, we calculate the live status and date using the BCL params.

packages/common/src/utils/media.ts
ChristiaanScheermeijer: Good feedback, but I prefer to do that separately and focus this PR on the fix (it already has some overhead due to the `isContentType` util)

packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaLiveChannel/MediaLiveChannel.tsx
ChristiaanScheermeijer: This PR fixes the "Live channel not found" when creating a live stream via JWP dashboard. 

We could improve it, but we might need to agree to some content-type standards first. Because `liveChannel` was intended to be used for live streams with EPG and live channels page. Since this will take some time, I suggest that we first ensure live streams are working again 😄

AntonLantukh:
> That's correct; that's the reason why the live channel start date was shown on the cards. We should assume that when the contentType is set to `liveChannel`, the live stream is 24x7 and always live. When using contentType `liveEvent`, we calculate the live status and date using the BCL params.

@ChristiaanScheermeijer Live Channels get same custom params as Live Events (except "VCH.ScheduledEnd"). So it is also possible that it is created ("PRE_LIVE") but not live yet.

ChristiaanScheermeijer:
Thanks for the feedback @AntonLantukh, but that sounds like a new feature to me 😄

AntonLantukh:
We can start with this solution, but I would propose the following as next steps:

1. Have one screen, which is "Live" (MediaEvent now) for all live events except ones being part of EPG
2. In case we have start time / end time -> we show it. If not -> it will look like a MediaMovie.
3. Tags will be just rendered conditionally depending on the presence

AntonLantukh (approved):

royschut (approved):

PR #535, opened 2024-05-28
Fix / live tag overlap issues

ChristiaanScheermeijer:
## Description

A small UI fix for the live tag in the EPG component. This only happens when setting the language to Spanish which uses more text for the live label (en vivo).

**Before:**

image

image
 
**After:**

image

image


MelissaDTH (approved):

langemike (approved):
lgtm

PR #538, opened 2024-05-29
Fix / Restrict text length to avoid scrolling issue

MelissaDTH:
## Description

**Issue:** The video detail page starts halfway down the viewport when a media item has a very long description, because the "Start Watching" button receives focus.

#### Solution
By limiting the number of lines in the video description, we not only ensure that the page does not start halfway down, but it also aligns better with the designs and overall visuals. 

- I've introduced a reusable `TruncatedText` component and modified `CollapsibleText` to remove the unnecessary `maxHeight` prop. Added a `maxHeight` constant with the original value for readability in calculations.
- Verified the fix with video descriptions of various lengths and tested with 175% zoom on smaller screens to ensure proper line limit and prevent page scrolling.

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/packages/ui-react/src/components/CollapsibleText/CollapsibleText.tsx
+++ b/packages/ui-react/src/components/CollapsibleText/CollapsibleText.tsx
@@ -4,6 +4,7 @@ import ChevronRight from '@jwp/ott-theme/assets/icons/chevron_right.svg?react';
 
 import IconButton from '../IconButton/IconButton';
 import Icon from '../Icon/Icon';
+import useBreakpoint from '../../hooks/useBreakpoint';
 
 import styles from './CollapsibleText.module.scss';
 
@@ -14,6 +15,7 @@ type Props = {
 
 const CollapsibleText: React.FC<Props> = ({ text, className }: Props) => {
   const divRef = useRef<HTMLDivElement>() as React.MutableRefObject<HTMLDivElement>;
+  const breakpoint = useBreakpoint();
   const [doesFlowOver, setDoesFlowOver] = useState(false);
   const [expanded, setExpanded] = useState(false);
 
@@ -24,7 +26,7 @@ const CollapsibleText: React.FC<Props> = ({ text, className }: Props) => {
 
   useEffect(() => {
     divRef.current && setDoesFlowOver(divRef.current.scrollHeight > divRef.current.offsetHeight + clippablePixels || maxHeight < divRef.current.offsetHeight);
-  }, [maxHeight, text]);
+  }, [maxHeight, text, breakpoint]);
 
   return (
     <div className={classNames(styles.collapsibleText)}>
--- a/packages/ui-react/src/components/CollapsibleText/CollapsibleText.tsx
+++ b/packages/ui-react/src/components/CollapsibleText/CollapsibleText.tsx
@@ -1,7 +1,6 @@
 import classNames from 'classnames';
 import React, { useEffect, useRef, useState } from 'react';
 import ChevronRight from '@jwp/ott-theme/assets/icons/chevron_right.svg?react';
-import useBreakpoint from '@jwp/ott-ui-react/src/hooks/useBreakpoint';
 
 import IconButton from '../IconButton/IconButton';
 import Icon from '../Icon/Icon';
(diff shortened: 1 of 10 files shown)


packages/ui-react/src/components/TruncatedText/TruncatedText.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+  test('renders and matches snapshot', () => {
+    const { container } = render(<TruncatedText text="Test..." maximumLines={8} />);
+
+    expect(container).toMatchSnapshot();
+  });
+
+  test('WCAG 2.2 (AA) compliant', async () => {
+    const { container } = render(<TruncatedText text="Test..." maximumLines={8} />);
+
+    expect(await axe(container, { runOnly: ['wcag21a', 'wcag21aa', 'wcag22aa'] })).toHaveNoViolations();
```
langemike: Even though the intention is great, this test doesn't add a lot of value because it's only a `div` with text in it. This cannot go wrong. Do you agree?

MelissaDTH: Hmm yes it might be a bit redundant, I agree, I'll remove this one

langemike (approved):
Nice!

packages/ui-react/src/components/CollapsibleText/CollapsibleText.tsx
```diff
@@ -11,25 +10,21 @@ import styles from './CollapsibleText.module.scss';
 type Props = {
   text: string;
   className?: string;
-  maxHeight?: 'none' | number;
 };
 
-const CollapsibleText: React.FC<Props> = ({ text, className, maxHeight = 'none' }: Props) => {
+const CollapsibleText: React.FC<Props> = ({ text, className }: Props) => {
   const divRef = useRef<HTMLDivElement>() as React.MutableRefObject<HTMLDivElement>;
-  const breakpoint = useBreakpoint();
```
ChristiaanScheermeijer, line 19: We might need the `useBreakpoint` here to update the `doesFlowOver` state when the browser resizes. It probably works now because the `TruncatedText` also uses the `useBreakpoint` hook, but not when this component is used directly.

packages/ui-react/src/components/VideoDetails/VideoDetails.tsx
```diff
@@ -49,7 +46,7 @@ const VideoDetails: React.VFC<Props> = ({
               <div className={styles.primaryMetadata}>{primaryMetadata}</div>
               {secondaryMetadata && <div className={styles.secondaryMetadata}>{secondaryMetadata}</div>}
             </div>
-            <CollapsibleText text={description} className={styles.description} maxHeight={isMobile ? 60 : 'none'} />
+            <TruncatedText text={description} maximumLines={120} className={styles.description} />
```
ChristiaanScheermeijer: Are you sure the maximum lines here is correct? 😅

packages/ui-react/src/components/CollapsibleText/CollapsibleText.tsx
MelissaDTH: Sounds good, I've reintroduced it!

ChristiaanScheermeijer:
Thanks for the updates @MelissaDTH! There is still one conceptual thing I didn't mention, and I'm curious if you and others agree.

I think we should simplify the `TruncatedText` component like the `CollapsibleText` component. Both have separate functions, but now the `TruncatedText` component is also responsible for choosing to truncate or collapse the text.

So I think we should:

1) Update the `TruncatedText` component to just truncate the given text
2) Either conditionally render either in the updated screens/components OR create a specific component that implements this logic

The advantage is that both components can be used agnostic and testing these components becomes more simple.

What do you think?

langemike:
> * Update the `TruncatedText` component to just truncate the given text

I agree. I would go for option 1!

MelissaDTH:
> I think we should simplify the `TruncatedText` component like the `CollapsibleText` component. Both have separate functions, but now the `TruncatedText` component is also responsible for choosing to truncate or collapse the text.
> ...

Thanks for the feedback, @ChristiaanScheermeijer. I spoke with Mike about this issue and during that conversation, I advocated for the idea that since `TruncatedText` doesn't have much functionality, it would simplify matters to assign it the responsibility of determining how to render the text based on whether it's being viewed on mobile or not. However, now that both of you suggest going in this direction, I've implemented it!

ChristiaanScheermeijer (approved):
Nice! 🎉

PR #539, opened 2024-05-31
Feat / log service

ChristiaanScheermeijer:
## Description

This PR implements a log service according to the discussion here #121 which can be extended or replaced with custom solutions. 

The next step is to add more debug logging (not at once).

### DIFF
--- a/packages/common/src/services/SettingsService.ts
+++ b/packages/common/src/services/SettingsService.ts
@@ -3,10 +3,10 @@ import ini from 'ini';
 import { getI18n } from 'react-i18next';
 
 import { CONFIG_FILE_STORAGE_KEY, CONFIG_QUERY_KEY, OTT_GLOBAL_PLAYER_ID } from '../constants';
-import { logDev } from '../utils/common';
 import { AppError } from '../utils/error';
 import type { Settings } from '../../types/settings';
 import env from '../env';
+import { logDebug, logWarn } from '../logger';
 
 import StorageService from './StorageService';
 
@@ -44,7 +44,7 @@ export default class SettingsService {
         return configKey;
       }
 
-      logDev(`Invalid app-config query param: ${configKey}`);
+      logWarn('SettingsService', `Invalid app-config query param: ${configKey}`);
     }
     // Yes this falls through from above to look up the stored value if the query string is invalid and that's OK
 
@@ -56,7 +56,7 @@ export default class SettingsService {
         return storedSource;
       }
 
-      logDev('Invalid stored config: ' + storedSource);
+      logWarn('SettingsService', 'Invalid stored config: ' + storedSource);
       await this.storageService.removeItem(CONFIG_FILE_STORAGE_KEY);
     }
 
@@ -78,8 +78,8 @@ export default class SettingsService {
     const settings = await fetch('/.webapp.ini')
       .then((result) => result.text())
       .then((iniString) => ini.parse(iniString) as Settings)
-      .catch((e) => {
-        logDev(e);
+      .catch((error) => {
+        logDebug('SettingsService', 'Failed to fetch or parse the ini settings', { error });
         // It's possible to not use the ini settings files, so an error doesn't have to be fatal
         return {} as Settings;
       });
@@ -106,7 +106,7 @@ export default class SettingsService {
 
     // The player key should be set if using the global ott player
     if (settings.playerId === OTT_GLOBAL_PLAYER_ID && !settings.playerLicenseKey) {
-      console.warn('Using Global OTT Player without setting player key. Some features, such as analytics, may not work correctly.');
+      logWarn('SettingsService', 'Using Global OTT Player without setting player key. Some features, such as analytics, may not work correctly.');
     }
 
     // This will result in an unusable app
--- a/packages/common/src/services/integrations/cleeng/CleengService.ts
+++ b/packages/common/src/services/integrations/cleeng/CleengService.ts
@@ -3,12 +3,13 @@ import { object, string } from 'yup';
 import { inject, injectable } from 'inversify';
 import { BroadcastChannel } from 'broadcast-channel';
 
-import { IS_DEVELOPMENT_BUILD, logDev } from '../../../utils/common';
+import { IS_DEVELOPMENT_BUILD } from '../../../utils/common';
 import { PromiseQueue } from '../../../utils/promiseQueue';
 import type { AuthData } from '../../../../types/account';
 import StorageService from '../../StorageService';
 import { GET_CUSTOMER_IP } from '../../../modules/types';
 import type { GetCustomerIP } from '../../../../types/get-customer-ip';
+import { logDebug, logError } from '../../../logger';
 
 import type { GetLocalesResponse } from './types/account';
 import type { Response } from './types/api';
@@ -130,12 +131,14 @@ export default class CleengService {
       };
     } catch (error: unknown) {
       if (error instanceof Error) {
-        logDev('Failed to refresh accessToken', error);
+        logDebug('CleengService', 'Failed to refresh accessToken', { error });
 
         // only logout when the token is expired or invalid, this prevents logging out users when the request failed due to a
         // network error or for aborted requests
         if (error.message.includes('Refresh token is expired or does not exist') || error.message.includes('Missing or invalid parameter')) {
-          if (!this.logoutCallback) logDev('logoutCallback is not set');
+          if (!this.logoutCallback) {
+            logDebug('CleengService', 'logoutCallback is not set');
+          }
           await this.logoutCallback?.();
         }
       }
@@ -300,7 +303,7 @@ export default class CleengService {
         return;
       }
     } catch (error: unknown) {
-      logDev('Failed to refresh tokens', error);
+      logDebug('CleengService', 'Failed to refresh tokens', { error });
     }
 
     // if we are here, we didn't receive new tokens
@@ -319,7 +322,7 @@ export default class CleengService {
     try {
       // token is already refreshing, let's wait for it
       if (this.isRefreshing) {
-        logDev('Token is already refreshing, waiting in queue...');
+        logDebug('CleengService', 'Token is already refreshing, waiting in queue...');
         return await this.queue.enqueue();
       }
 
@@ -330,7 +333,7 @@ export default class CleengService {
 
       await this.refreshTokens(this.tokens);
     } catch (error: unknown) {
-      logDev('Error caught while refreshing the access token', error);
+      logError('CleengService', 'Error caught while refreshing the access token', { error });
     }
   };
 
--- a/packages/common/src/utils/common.ts
+++ b/packages/common/src/utils/common.ts
@@ -70,7 +70,7 @@ export function hexToRgb(color: string) {
  * @link {https://stackoverflow.com/a/35970186/1790728}
  * @return {string}
  */
-export function calculateContrastColor(color: string) {
+export function calculateContrastColor(color: string): string {
   const rgb = hexToRgb(color);
 
   if (!rgb) {
@@ -93,16 +93,6 @@ export const IS_TEST_MODE = __mode__ === 'test';
 export const IS_PREVIEW_MODE = __mode__ === 'preview';
 // Production mode
 export const IS_PROD_MODE = __mode__ === 'prod';
-
-export function logDev(message: unknown, ...optionalParams: unknown[]) {
-  if ((IS_DEVELOPMENT_BUILD || IS_PREVIEW_MODE) && !IS_TEST_MODE) {
-    if (optionalParams.length > 0) {
-      console.info(message, optionalParams);
-    } else {
-      console.info(message);
-    }
-  }
-}
 
 export const isContentType = (item: PlaylistItem | Playlist, contentType: string) => item.contentType?.toLowerCase() === contentType.toLowerCase();
 
--- a/packages/common/vitest.setup.ts
+++ b/packages/common/vitest.setup.ts
@@ -1,5 +1,20 @@
 import 'vi-fetch/setup';
 import 'reflect-metadata';
+import { mockService } from './test/mockService';
+import LogTransporter from './src/services/logging/LogTransporter';
+import ConsoleTransporter from './src/services/logging/ConsoleTransporter';
+import { LogLevel } from './src/services/logging/LogLevel';
+
+beforeEach(() => {
+  mockService(
+    LogTransporter,
+    __debug__
+      ? new ConsoleTransporter(LogLevel.DEBUG)
+      : {
+          log() {},
+        },
+  );
+});
 
 // a really simple BroadcastChannel stub. Normally, a Broadcast channel would not call event listeners on the same
 // instance. But for testing purposes, that really doesn't matter...
--- a/platforms/web/src/modules/register.ts
+++ b/platforms/web/src/modules/register.ts
@@ -3,9 +3,12 @@ import { container } from '@jwp/ott-common/src/modules/container';
 import StorageService from '@jwp/ott-common/src/services/StorageService';
 import { GET_CUSTOMER_IP } from '@jwp/ott-common/src/modules/types';
 import type { GetCustomerIP } from '@jwp/ott-common/types/get-customer-ip';
+import LogTransporter from '@jwp/ott-common/src/services/logging/LogTransporter';
+import ConsoleTransporter from '@jwp/ott-common/src/services/logging/ConsoleTransporter';
+import { LogLevel } from '@jwp/ott-common/src/services/logging/LogLevel';
 
-import { LocalStorageService } from '#src/services/LocalStorageService';
 import { getOverrideIP } from '#src/utils/ip';
+import { LocalStorageService } from '#src/services/LocalStorageService';
 
 /**
  * Custom integration override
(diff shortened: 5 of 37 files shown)


royschut (approved):
Great work! I've kept a sharp eye on the debug levels you've chosen for the existing logs, but I'd agree with all of them.

langemike (approved):
This is a nice improvement!

packages/common/src/services/SettingsService.ts
```diff
@@ -78,8 +78,8 @@ export default class SettingsService {
     const settings = await fetch('/.webapp.ini')
       .then((result) => result.text())
       .then((iniString) => ini.parse(iniString) as Settings)
-      .catch((e) => {
-        logDev(e);
+      .catch((error) => {
+        getLogger().error('SettingsService', 'Failed to fetch or parse the ini settings', error);
```
ChristiaanScheermeijer: I'm not sure if this should be an `error`. The ini file is not required...

packages/common/src/utils/common.ts
```diff
@@ -92,16 +92,6 @@ export const IS_PREVIEW_MODE = __mode__ === 'preview';
 // Production mode
 export const IS_PROD_MODE = __mode__ === 'prod';
 
-export function logDev(message: unknown, ...optionalParams: unknown[]) {
```
budzworthy, line 97: The one thing I really liked about this implementation of logging was that it was compile-safe. If a developer accidentally left in a debug logging statement like "This program is so stupid why won't this work!?" then at compile time for the prod build, that would get stripped out. Probably not a huge deal, but keeps prod a bit more 'professional'. Thoughts?

packages/common/vitest.setup.ts
```diff
@@ -1,5 +1,17 @@
 import 'vi-fetch/setup';
 import 'reflect-metadata';
+import { mockService } from './test/mockService';
+import LogService from './src/services/logging/LogService';
+
+beforeEach(() => {
+  mockService(LogService, {
```
budzworthy: Can we use the console transporter to help with debugging tests?

platforms/web/src/modules/register.ts
```diff
@@ -31,6 +34,19 @@ container.bind(StorageService).to(LocalStorageService);
...
+ *
+ * Add custom log transporters by registering more modules to the LOG_TRANSPORTER type. The custom transporter must
+ * implement the LogTransporter interface.
+ *
+ * @example
+ * ```ts
+ * container.bind<LogTransporter>(LOG_TRANSPORTER).toDynamicValue(() => new GoogleGTMTransporter(import.meta.env.DEV ? LogLevel.SILENT : LogLevel.WARN));
+ * ```
+ */
+container.bind<LogTransporter>(LOG_TRANSPORTER).toDynamicValue(() => new ConsoleTransporter(import.meta.env.DEV ? LogLevel.DEBUG : LogLevel.ERROR));
```
budzworthy: What did you have in mind for setting up the log settings via configuration?

packages/common/src/log.ts
```diff
@@ -0,0 +1,6 @@
+import { getModule } from './modules/container';
+import LogService from './services/logging/LogService';
+
+export const getLogger = () => {
```
budzworthy: Is there a plan to extend getLogger in any way? If not, maybe it's less boilerplate if we export the individual methods (i.e. `logDebug`, `logInfo`, etc.) Then callers can just import and use a single method (`logDebug(...)` instead of `getLogger().debug(...)`.

It also gives us a potential mechanism to compile away minimum log levels in production builds. I think we could even figure out the configuration to expose these as 'globals' so anytime we add a log statement we don't have to muck with the imports.

packages/common/src/log.ts
budzworthy, line 1: Can we rename this file to Logger.ts?

packages/common/src/modules/types.ts
```diff
@@ -3,3 +3,5 @@ export const INTEGRATION_TYPE = Symbol('INTEGRATION_TYPE');
 export const DETERMINE_INTEGRATION_TYPE = Symbol('DETERMINE_INTEGRATION_TYPE');
 
 export const GET_CUSTOMER_IP = Symbol('GET_CUSTOMER_IP');
+
+export const LOG_TRANSPORTER = Symbol('LOG_TRANSPORTER');
```
budzworthy: Do we need a symbol def for this or can we just multiinject anything with the LogTransporter interface?

packages/common/src/services/SettingsService.ts
budzworthy: Ya, that makes sense. logDev was info level, right?

packages/common/src/services/integrations/cleeng/CleengService.ts
```diff
@@ -130,12 +131,12 @@ export default class CleengService {
...
     } catch (error: unknown) {
       if (error instanceof Error) {
-        logDev('Failed to refresh accessToken', error);
+        getLogger().debug('CleengService', 'Failed to refresh accessToken', { error });
 
         // only logout when the token is expired or invalid, this prevents logging out users when the request failed due to a
         // network error or for aborted requests
         if (error.message.includes('Refresh token is expired or does not exist') || error.message.includes('Missing or invalid parameter')) {
-          if (!this.logoutCallback) logDev('logoutCallback is not set');
+          if (!this.logoutCallback) getLogger().debug('CleengService', 'logoutCallback is not set');
```
budzworthy: I generally recommend to always use braces. Maybe we should update linting to enforce.
```suggestion
          if (!this.logoutCallback) {
            getLogger().debug('CleengService', 'logoutCallback is not set');
          }
```

packages/common/src/services/logging/LogService.ts
```diff
@@ -0,0 +1,38 @@
+import { injectable, multiInject, optional } from 'inversify';
+
+import { LOG_TRANSPORTER } from '../../modules/types';
+
+import type LogTransporter from './transporters/LogTransporter';
+import { LogLevel } from './LogLevel';
+
+@injectable()
+export default class LogService {
+  constructor(@optional() @multiInject(LOG_TRANSPORTER) private readonly transporters: LogTransporter[]) {}
```
budzworthy: Does this work without the symbol?
```suggestion
  constructor(@optional() @multiInject(LogTransporter) private readonly transporters: LogTransporter[]) {}
```

packages/common/src/services/logging/transporters/ConsoleTransporter.ts
```diff
@@ -0,0 +1,29 @@
...
+export default class ConsoleTransporter implements LogTransporter {
+  constructor(readonly logLevel: LogLevel) {}
+
+  log(level: LogLevel, scope: string, message: string, extra?: Record<string, unknown>, error?: Error) {
+    if (level < this.logLevel) return;
+
+    switch (level) {
+      case LogLevel.ERROR:
+      case LogLevel.FATAL:
+        console.error(`[${scope}] ${message}`, error);
```
budzworthy: Should we prepend the log level to give more insight?
```suggestion
        console.error(`${level}: [${scope}] ${message}`, error);
```

packages/common/src/services/logging/LogService.ts
```diff
@@ -0,0 +1,38 @@
...
+      transporter.log(logLevel, scope, message, extra, error);
+    });
+  }
+
+  private wrapError(error: unknown) {
+    return error instanceof Error ? error : new Error(String(error));
+  }
+
+  // Logs detailed debugging information, useful during development
+  debug = (scope: string, message: string, extra?: Record<string, unknown>) => this.log(LogLevel.DEBUG, scope, message, extra);
```
budzworthy: If you change the interface slightly, it will make some of the questions elsewhere easier, like not using log error for the ini file, but still being able to log the error.

```suggestion
  debug = (scope: string, message: string, {error?: unknown, extra?: Record<string, unknown>}) => this.log(LogLevel.DEBUG, scope, message, extra);
```

packages/common/src/log.ts
budzworthy: An alternative would be to add the 'scope' to the getLogger and store a single instance per use case instead of copying and pasting scope and getLogger.

packages/common/src/services/logging/LogService.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import type LogTransporter from './transporters/LogTransporter';
+import { LogLevel } from './LogLevel';
+
+@injectable()
+export default class LogService {
+  constructor(@optional() @multiInject(LOG_TRANSPORTER) private readonly transporters: LogTransporter[]) {}
+
+  private log(logLevel: LogLevel, scope: string, message: string, extra?: Record<string, unknown>, error?: Error) {
+    this.transporters.forEach((transporter) => {
```
budzworthy: I feel like this logic can be safely moved up a level from where it's at now in the transporters.
```suggestion
    this.transporters.filter(t => t.logLevel <= logLevel).forEach((transporter) => {
```

packages/common/src/log.ts
ChristiaanScheermeijer: Hmm, interesting. It is not meant to be extended but acts as a shortcut to the LogService. If we add all methods to the log.ts (Logger.ts) file, we don't need the LogService anymore. The Logger.ts can then just call the method on transporters directly. What do you think?

packages/common/src/log.ts
ChristiaanScheermeijer: Are you sure? We use that convention only for classes and components currently.

packages/common/src/modules/types.ts
ChristiaanScheermeijer: LogTransporter is an Interface which doesn't allow for usage as value. We could make it an abstract class. Then it should be possible 🙂

packages/common/src/services/integrations/cleeng/CleengService.ts
ChristiaanScheermeijer: I can update this, but there are probably more of these in the codebase. Shall we do this in a separate update linting PR?

packages/common/src/services/logging/LogService.ts
ChristiaanScheermeijer: See https://github.com/jwplayer/ott-web-app/pull/539#discussion_r1625828136

packages/common/src/services/logging/transporters/ConsoleTransporter.ts
ChristiaanScheermeijer: Sure! I removed the level to keep the messages cleaner, but I'm not against re-adding it.

packages/common/src/services/logging/LogService.ts
ChristiaanScheermeijer: Great suggestion! Do we want to do this for all log methods? Perhaps error doesn't need to be required for error and fatal logs.

The only downside using objects as params is that it will take more room in some scenarios:

```ts
getLogger().error('ApiService', 'Failed to retrieve the playlist', {
  error,
  extra: {
    response,
  },
});
```

packages/common/src/services/logging/LogService.ts
ChristiaanScheermeijer: I purposely placed that logic in the transporter so that the implementation is able to decide what to do with the messages.

This is useful when, for example, the transporter only collects info/debug/warn logs and publishes this alongside error/fatal logs.

For example, using Sentry, you can decide to either capture all messages as logs or to do this just before capturing errors.

packages/common/src/utils/common.ts
ChristiaanScheermeijer: TBH I think it doesn't make a difference because the `logDev("This stupid ..")` call would still be present in the compiled code. I think the compiler only removes `if(false)` blocks which is just the body of the `logDev` function.

packages/common/vitest.setup.ts
ChristiaanScheermeijer: That is possible, but I'm not sure if this needs to be enabled by default, right?

platforms/web/src/modules/register.ts
ChristiaanScheermeijer: I don't have a better solution than using environment variables. The only alternative I can think of is adding a script tag in the index.html so it can be changed per environment:

```html
<script>window.LOG_LEVEL = 'info';</script>
```

This only works for the ConsoleTransporter. If publishers add custom transporters, the configuration can also be customized.

If we want to use the SettingsService (ini-files), we could consider moving the `getSettings` call to the index.tsx.

packages/common/src/utils/common.ts
budzworthy: I think I did test it by searching the production build for the strings and at least at one point they weren't showing up. That might have been several refactors ago though. Not the most important thing, but I thought it was worth mentioning.

packages/common/vitest.setup.ts
budzworthy: I mean wouldn't it be just as easy to do something like:
```suggestion
  mockService(LogService, new LogService(new ConsoleTransporter()))
```

packages/common/src/log.ts
budzworthy: Ya, that's also probably fine. The only reason we have the log service is in case someone wants to override it at a higher level than the transporters, but the transporters look like they give us pretty good flexibility already, so we can probably collapse the hierarchy 1 level.

packages/common/src/log.ts
budzworthy: OK, `logger.ts`? `logging.ts`? `log.ts` just looks funny to me.

packages/common/src/modules/types.ts
budzworthy: Yes, great, let's do that!

packages/common/src/services/logging/LogService.ts
budzworthy: True, but it does make it more readable. Maybe you can play around with typings and find a good middle ground?

packages/common/vitest.setup.ts
ChristiaanScheermeijer: That would probably work, but I do not recommend instantiating Inversify modules manually because it can break when changing injection arguments.

I'm still unsure if we really want the logging output mixed with the Vitest output. If you think this is useful why don't we do this instead:

```ts
const makeLogger = (level: string) => {
  return (scope: string, message: string, extra: unknown) => {
    return import.meta.env.APP_DEBUG && console.log(`${level} [${scope}] ${message}`, extra);
  };
};
 
mockService(LogService, {
  debug: makeLogger('debug'),
  info: makeLogger('info'),
  warn: makeLogger('warn'),
  error: makeLogger('error'),
  fatal: makeLogger('fatal'),
});
```

packages/common/src/modules/types.ts
ChristiaanScheermeijer: Done

packages/common/src/log.ts
ChristiaanScheermeijer: Updated the PR

packages/common/src/utils/common.ts
ChristiaanScheermeijer: Hmm, you might be right...

packages/common/vitest.setup.ts
ChristiaanScheermeijer: After the latest change, the above becomes:

```ts
mockService(LogTransporter, {
  log: console.log,
});
```

budzworthy: That looks good to me. I'm fine if you want to make it conditional with an environment variable, just so long as it's easy enough to enable when someone is stuck trying to debug tests.

packages/common/src/services/logging/LogService.ts
budzworthy: Hm, interesting use case. 2 asks then:
1) Add a comment explaining why we don't filter in LogService ("Call all transporters, regardless of log level, because some transporters use custom logic like batching or grouping logs of different levels."
2) Remove logLevel field from the LogTransporter class, because it's not indicative to all implementations if different implementations may not have a clear cut, single log level

ChristiaanScheermeijer: Good point. I did both 😄

packages/common/vitest.setup.ts
ChristiaanScheermeijer: Updated the PR! 

Although, we will have to refactor the globals variables (`__debug__`, etc...) at some point because it doesn't play well with other platforms.

packages/common/src/services/logging/LogService.ts
ChristiaanScheermeijer: @dbudzins I have one alternative:

```ts
export type LogParams = { error?: unknown; [key: string]: unknown };

export const makeLogFn =
  (logLevel: LogLevel) =>
  (scope: string, message: string, { error, ...extra }: LogParams = {}) => {
    const transporters = getAllModules(LogTransporter);

    // call log on all transporters, the transporters should decide to handle the call or not
    transporters.forEach((transporter) => {
      transporter.log(logLevel, scope, message, extra, error ? wrapError(error) : undefined);
    });
  };
```

The `extra` property is obtained using spread instead of a nested property. This removes the nested object:

```ts
logError('ApiController', 'Something is wrong', { error, extra: { response } });
``` 

Becomes:

```ts
logError('ApiController', 'Something is wrong', { error, response });
``` 

What do you think?

ChristiaanScheermeijer: @dbudzins @AntonLantukh I think this is the last open discussion. I am curious about your opinion so we can finish this PR.

budzworthy: Good alternative :+1:

AntonLantukh (approved):
Looks good!

Issue #518, opened 2024-04-24
Integration offerId inconsistency


ChristiaanScheermeijer:
We found some issues with the offerIds being used differently between the JWP and Cleeng integrations.

The difference between JWP assets and Cleeng offers is that JWP uses a single asset with two pricing options (monthly & annual). But for Cleeng, two offers need to be created. In the codebase, JWP imitates having two offers and this has some problems with backtracking the correct assetId for some places.

We should consider making this more generic and not integration-specific in the business logic.

See #492 for more information about this.

ChristiaanScheermeijer:
@dbudzins @AntonLantukh 

Unfortunately, we also ran into this problem while integrating IAP and need to solve this for a good flow.

## Problem

A bit more context to this problem.

The JWP integration is configured using one asset. Based on this asset, the access fees configured in the dashboard will be transformed into two `Offer` objects.

The `offerId` and `id` are generated using the `AccessFee#id`, and the asset ID is not added. 

https://github.com/jwplayer/ott-web-app/blob/71469721cacb6224e080910e5491da7935f5eea9/packages/common/src/services/integrations/jwp/JWPCheckoutService.ts#L49-L55

When the PayPal payment succeeds, the user is sent to the following URL, which adds the `offerId` of the selected offer to the query params. For a JWP offer, this will be `S123456` (S + access fee)

https://github.com/jwplayer/ott-web-app/blob/71469721cacb6224e080910e5491da7935f5eea9/packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx#L105

The WaitingForPayment component will poll until it has access to the given offerId:

https://github.com/jwplayer/ott-web-app/blob/71469721cacb6224e080910e5491da7935f5eea9/packages/ui-react/src/components/WaitingForPayment/WaitingForPayment.tsx#L15

In the `useCheckAccess` hook, which does the polling, we use the given `offerId` and call the `checkEntitlements` function.

Also note the fallback here: when no `offerId` is given, the `offerId` is obtained from the `checkoutController.getSubscriptionOfferIds` function. This uses the `offerId` from the app config, which is the asset ID and not the access fee ID.

https://github.com/jwplayer/ott-web-app/blob/71469721cacb6224e080910e5491da7935f5eea9/packages/hooks-react/src/useCheckAccess.ts#L22-L31

Finally, we call it `checkAccessForAsset` using the `offerId`. 

[...]

PR #541, opened 2024-06-04
Fix / waiting for payment PPV/TVOD offers

ChristiaanScheermeijer:
## Description

This PR fixes issue #518. I've updated the formatted `offerId` in the JWP integration. The `offerId` is now composed of the offer prefix, asset ID, and access fee ID. The composed offer ID is added to the URL when the user is sent to the waiting for payments modal. The integration is responsible for retrieving the asset ID when needed.

I also encountered a problem because the `getSubscriptionSwitches` method threw an error in the `initialiseOffers` method. This caused the query to retry a few times before resolving, making the Choose offers modal really slow.

The third change is that the welcome modal isn't shown to users buying PPV access since the welcome modal shows a message stating "You now have access to all content", which isn't true when buying access to a single item.

### DIFF
--- a/packages/common/src/controllers/CheckoutController.ts
+++ b/packages/common/src/controllers/CheckoutController.ts
@@ -27,6 +27,7 @@ import { useCheckoutStore } from '../stores/CheckoutStore';
 import { useAccountStore } from '../stores/AccountStore';
 import { FormValidationError } from '../errors/FormValidationError';
 import { determineSwitchDirection } from '../utils/subscription';
+import { logDev } from '../utils/common';
 
 @injectable()
 export default class CheckoutController {
@@ -52,10 +53,14 @@ export default class CheckoutController {
       useCheckoutStore.setState({ subscriptionOffers });
     }
 
-    if (!useCheckoutStore.getState().switchSubscriptionOffers.length) {
-      const subscriptionSwitches = await this.getSubscriptionSwitches();
-      const switchSubscriptionOffers = subscriptionSwitches ? await this.getOffers({ offerIds: subscriptionSwitches }) : [];
-      useCheckoutStore.setState({ switchSubscriptionOffers });
+    try {
+      if (!useCheckoutStore.getState().switchSubscriptionOffers.length) {
+        const subscriptionSwitches = await this.getSubscriptionSwitches();
+        const switchSubscriptionOffers = subscriptionSwitches ? await this.getOffers({ offerIds: subscriptionSwitches }) : [];
+        useCheckoutStore.setState({ switchSubscriptionOffers });
+      }
+    } catch (error) {
+      logDev('Failed to get subscription switches', error);
     }
   };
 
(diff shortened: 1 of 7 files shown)


langemike (approved):
These look like good fixes to me.

packages/hooks-react/src/useCheckAccess.ts
```diff
@@ -23,20 +23,28 @@ const useCheckAccess = () => {
...
     ({ interval = 3000, iterations = 5, offerId, callback }: IntervalCheckAccessPayload) => {
-      if (!offerId && offers?.[0]) {
-        offerId = offers[0];
+      if (!offerId) {
+        offerId = offers?.[0] || '';
+      }
+
+      if (!offerId) {
+        callback?.({ hasAccess: false, offerId: '' });
+        return;
```
royschut: Could we default the arg? 
```javascript
({ interval = 3000, iterations = 5, offerId = offers?.[0], callback }: IntervalCheckAccessPayload) => {
```

Or otherwise perhaps:
```suggestion
      if (!offerId) {
        if (!offers?.[0]) {
          callback?.({ hasAccess: false, offerId: '' });
          
          return;
        }
        offerId = offers[0];
      }
```

packages/ui-react/src/components/WaitingForPayment/WaitingForPayment.tsx
```diff
@@ -23,11 +23,18 @@ const WaitingForPayment = () => {
...
-      callback: (hasAccess) => {
+      callback: ({ hasAccess, offerId }) => {
         if (!hasAccess) return;
 
         announce(t('checkout.payment_success'), 'success');
-        navigate(modalURLFromLocation(location, 'welcome'));
+
+        // close the modal for PPV/TVOD offers
+        if (offerId.startsWith('C') || offerId.startsWith('P')) {
+          // should we show a dedicated modal for TVOD access?
```
royschut:
```suggestion
          // @todo: should we show a dedicated modal for TVOD access?
```

royschut (approved):
Really great work!

packages/common/src/controllers/CheckoutController.ts
```diff
@@ -52,10 +53,14 @@ export default class CheckoutController {
       useCheckoutStore.setState({ subscriptionOffers });
     }
 
-    if (!useCheckoutStore.getState().switchSubscriptionOffers.length) {
-      const subscriptionSwitches = await this.getSubscriptionSwitches();
-      const switchSubscriptionOffers = subscriptionSwitches ? await this.getOffers({ offerIds: subscriptionSwitches }) : [];
-      useCheckoutStore.setState({ switchSubscriptionOffers });
+    try {
+      if (!useCheckoutStore.getState().switchSubscriptionOffers.length) {
+        const subscriptionSwitches = await this.getSubscriptionSwitches();
```
AntonLantukh: Do you think it could make sense to return an empty array for `getSubscriptionSwitches` as this method is used as part of the init and we could expect to throw errors every time we open the page?

packages/hooks-react/src/useCheckAccess.ts
ChristiaanScheermeijer: Yes, good suggestion!

packages/common/src/controllers/CheckoutController.ts
ChristiaanScheermeijer: Yes, you're right. I updated the `getSubscriptionSwitches` so it does not throw when it's not implemented.

AntonLantukh (approved):

PR #544, opened 2024-06-07
Remove Discount codes from Dashboard and WebApp

borkopetrovicc:
## Description

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


ChristiaanScheermeijer:
Why is this feature removed? Customers depend on this and it is still supported by both integrations. I suggest that we remove this when we have an alternative.

marcovandeveen:
@darkoatanasovski @borkopetrovicc 

Let's keep this feature in: 
- We are still rolling out InPlayer 'standalone' implementations
- We don't have an alternative yet

We should still hide coupons in the dashboard.

borkopetrovicc:
> Why is this feature removed? Customers depend on this and it is still supported by both integrations. I suggest that we remove this when we have an alternative.

This is part of the initiative to remove features that we will not be supporting at the moment, and this applies only to the JWP integration.

PR #545, opened 2024-06-07
Feat / show external payment explanation

ChristiaanScheermeijer:
## Description

If a user has purchased a subscription via Google, Apple or Roku stores, managing the subscription via the web app is not possible (or very difficult).

This PR adds a message that the subscription is managed through an external payment provider and a link with information on how to manage your subscription instead.

The "Change subscription" button is also removed when an external payment provider is used.

image

### DIFF
--- a/packages/ui-react/src/components/Payment/Payment.tsx
+++ b/packages/ui-react/src/components/Payment/Payment.tsx
@@ -58,13 +58,9 @@ type Props = {
   setSelectedOfferId: (offerId: string | null) => void;
   isUpgradeOffer: boolean | undefined;
   setIsUpgradeOffer: (isUpgradeOffer: boolean | undefined) => void;
-};
-
-const EXTERNAL_PAYMENT_METHODS = ['Apple In-App', 'Android In-App', 'Roku In-App'];
-const STORE_LINKS: Record<string, string> = {
-  apple: 'https://support.apple.com/en-qa/118428',
-  android: 'https://support.google.com/googleplay/answer/7018481?hl=en',
-  roku: 'https://support.roku.com/article/208756478',
+  isExternalPaymentProvider: boolean;
+  paymentProvider?: string;
+  paymentProviderLink?: string;
 };
 
 const Payment = ({
@@ -93,6 +89,9 @@ const Payment = ({
   setSelectedOfferId,
   isUpgradeOffer,
   setIsUpgradeOffer,
+  isExternalPaymentProvider,
+  paymentProvider,
+  paymentProviderLink,
 }: Props): JSX.Element => {
   const subscriptionDetailsId = useOpaqueId('subscription-details');
   const paymentMethodId = useOpaqueId('payment-method');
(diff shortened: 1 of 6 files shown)


royschut (approved):

packages/ui-react/src/components/Payment/Payment.tsx
```diff
@@ -203,22 +211,31 @@ const Payment = ({
...
-                    }
-                  }}
-                  fullWidth={isMobile}
-                  color="primary"
-                  data-testid="change-subscription-button"
-                />
+              {isExternalPaymentProvider ? (
+                <p className={styles.explanation}>
+                  {t('account:external_payment.explanation', { paymentProvider })}{' '}
+                  <Link href={paymentProviderLink} target="_blank">
```
langemike, line 217: Maybe add `rel="noopener noreferrer"` for safety?

packages/ui-react/src/containers/PaymentContainer/PaymentContainer.tsx
```diff
@@ -44,6 +44,13 @@ const processBillingReceipt = (receipt: Blob | string, transactionId: string) =>
   }
 };
 
+const EXTERNAL_PAYMENT_METHODS = ['Apple In-App', 'Android In-App', 'Roku In-App'];
+const STORE_LINKS: Record<string, string> = {
+  apple: 'https://support.apple.com/en-qa/118428',
+  android: 'https://support.google.com/googleplay/answer/7018481?hl=en',
+  roku: 'https://support.roku.com/article/208756478',
```
langemike: I don't want to overcomplicate things; but somewhere in the future we need to support multilingual links

packages/ui-react/src/components/Payment/Payment.tsx
langemike: This comment is outdated. Since 2021 it is not needed. `noreferrer` is optional

langemike (approved):
Looks good

packages/ui-react/src/components/Payment/Payment.tsx
ChristiaanScheermeijer: The Link component does that automagically when the `target` is set to `_blank` 😄

packages/ui-react/src/containers/PaymentContainer/PaymentContainer.tsx
ChristiaanScheermeijer: Good point. I tested it without language parameters, and this seems to redirect you to the correct language. I think this is the best solution instead that we choose the language.

MelissaDTH (approved):

packages/ui-react/src/containers/PaymentContainer/PaymentContainer.tsx
ChristiaanScheermeijer: I removed the language parameters and made a fix for Android (Google) IAP payments not being recognized.

langemike: It doesn't redirect me to the correct language when I set my browser to German using this method: https://stackoverflow.com/a/41190151

After a small research, these are the correct (auto redirect) URLs:

Apple: https://support.apple.com/118428
Android: https://support.google.com/googleplay/answer/7018481 (removed the `hl` query string)
Roku: unknown. But these are mostly targetted to US audience, but I consider this less a problem.

PR #546, opened 2024-06-11
Feat / make button states consistent

ChristiaanScheermeijer:
## Description

This aligns and fixes some button state inconsistencies. Most importantly, the accessibility mixins weren't used correctly due to it being used in a CSS module file which converts `body.is-tabbing` to a scoped class name, e.g. `body.is-tabbing-g34tsda`.

- Fix a11y tabbing visibility for the primary button
- Remove the button scale effect when tabbing
- Separate focus and hover states (this fixes the button being in a selected state after pressing it)
- Align featured card hover and focus scale
- Added separate `:active` state for all buttons to indicate the pressing action (like it should)

**Before "Start watching" button focus**

image

**After "Start watching" button focus**

image

**Before "Favorites" button focus**

image

**After "Favorites" button focus**

image

### DIFF
--- a/packages/ui-react/src/components/Button/Button.module.scss
+++ b/packages/ui-react/src/components/Button/Button.module.scss
@@ -1,6 +1,5 @@
 @use '@jwp/ott-ui-react/src/styles/variables';
 @use '@jwp/ott-ui-react/src/styles/theme';
-@use '@jwp/ott-ui-react/src/styles/accessibility';
 @use '@jwp/ott-ui-react/src/styles/mixins/responsive';
 
 $small-button-height: 28px;
@@ -58,8 +57,7 @@ $large-button-height: 40px;
     background-color: theme.$btn-default-bg;
 
     &:focus {
-      @include accessibility.accessibleOutline;
-      outline-offset: 1px;
+      outline-offset: 2px;
     }
   }
 
@@ -68,8 +66,7 @@ $large-button-height: 40px;
     background-color: var(--highlight-color, theme.$btn-primary-bg);
 
     &:focus {
-      @include accessibility.accessibleOutline;
-      outline-offset: 1px;
+      outline-offset: 2px;
     }
   }
 
--- a/packages/ui-react/src/components/Button/Button.module.scss
+++ b/packages/ui-react/src/components/Button/Button.module.scss
@@ -25,14 +25,17 @@ $large-button-height: 40px;
   border-radius: 4px;
   outline: none;
   cursor: pointer;
-  transition: background-color 0.1s ease, transform 0.1s ease;
+  transition: background-color 0.1s ease, transform 0.1s ease, opacity 0.1s ease;
 
   @media (hover: hover) and (pointer: fine) {
     &:not(.disabled) {
-      &:hover,
-      &:focus {
+      &:hover {
         z-index: 1;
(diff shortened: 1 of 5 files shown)


packages/ui-react/src/components/IconButton/IconButton.module.scss
```diff
@@ -12,11 +13,19 @@
...
 
-  &:hover,
-  &:focus {
+  &:hover {
     transform: scale(1.1);
     opacity: 1;
   }
+
+  &:focus {
+    @include accessibility.accessibleOutline;
```
royschut: This mixin is also added globally. But perhaps we prefer it more explicit.

https://github.com/jwplayer/ott-web-app/blob/3329a0916bde27a66559ce492fd1af7013450311/packages/ui-react/src/styles/accessibility.scss#L41-L52

packages/ui-react/src/components/Button/Button.module.scss
```diff
@@ -53,28 +56,31 @@ $large-button-height: 40px;
...
   }
 
   &.primary {
     color: var(--highlight-contrast-color, theme.$btn-primary-color);
     background-color: var(--highlight-color, theme.$btn-primary-bg);
 
     &:focus {
-      @include accessibility.accessibleOutlineContrast;
+      @include accessibility.accessibleOutline;
+      outline-offset: 1px;
```
royschut: Since you've taken away the scaling, considering a11y, I have a minor doubt if 1px offset is enough for the primary buttons. The color of the button and the outline are the same, with only 1px between. Should we increase to 2px?

Example, current vs. accessibility.nl:

Screenshot 2024-06-12 at 10 47 55

Screenshot 2024-06-12 at 10 52 40

langemike (approved):
Nice improvement! Well spotted!

I have tested it myself locally to confirm

packages/ui-react/src/components/IconButton/IconButton.module.scss
ChristiaanScheermeijer: You found another problem, indeed. The usage of the `:global` pseudo class name prevents this selector from working again. I have fixed this by not using the `@include` but explicitly defining the CSS rules...

packages/ui-react/src/components/Button/Button.module.scss
ChristiaanScheermeijer: I have increased the outline offset for buttons to 2px, curious what others think

langemike (approved):
Great!

packages/ui-react/src/components/Button/Button.module.scss
langemike: Both look good

MelissaDTH: The [WCAG guidelines](https://www.w3.org/TR/WCAG22/#focus-appearance) for Level AAA focus indicators (a level we’re not necessarily aiming for, but hey, it’s great if we can achieve it) state that the focus area should be at least as large as the area of a 2 CSS pixel thick perimeter. 

So, increasing the outline offset to `2px` here is the minimum requirement for that and thus a good change!

MelissaDTH (approved):
LGTM, nice changes!

royschut (approved):

Issue #333, opened 2023-07-10
Font to large of social login buttons


marcovandeveen:
Suggestion: 

<img width="824" alt="image" src="https://github.com/jwplayer/ott-web-app/assets/42120041/0305a108-e6b4-46b1-8565-66a5148eddd0">

ChristiaanScheermeijer:
Social login buttons are removed from the web app.

PR #550, opened 2024-06-12
Handle multiple plans from the AppConfig in the Web App 

mirovladimitrovski:
## Description

This PR solves #PS-226.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
```diff
@@ -50,42 +60,24 @@ export default class JWPCheckoutService extends CheckoutService {
...
-        return parseInt(offerId.slice(1).split('_')[0]);
-      }
-
-      // offer id format `<assetId>`
-      return parseInt(offerId);
-    }
-
-    return offerId;
+  private formatOfferId(offer: PlanPrice) {
+    return `${offer.access.type === 'subscription' ? 'S' : 'C'}${offer.id}`; //`${offer.access.type === 'subscription' ? 'S' : 'C'}S${offer.item_id}_${offer.id}`;
```
ChristiaanScheermeijer: > [!WARNING]
> This change undos the fix from PR: https://github.com/jwplayer/ott-web-app/pull/541

packages/hooks-react/src/useEntitlement.ts
```diff
@@ -1,11 +1,14 @@
...
+import { useQueries, useQuery } from 'react-query';
 import type { PlaylistItem } from '@jwp/ott-common/types/playlist';
 import type { GetEntitlementsResponse } from '@jwp/ott-common/types/checkout';
 import type { MediaOffer } from '@jwp/ott-common/types/media';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
 import { useAccountStore } from '@jwp/ott-common/src/stores/AccountStore';
+import { useCheckoutStore } from '@jwp/ott-common/src/stores/CheckoutStore';
 import CheckoutController from '@jwp/ott-common/src/controllers/CheckoutController';
+import JWPEntitlementService from '@jwp/ott-common/src/services/JWPEntitlementService';
```
ChristiaanScheermeijer, line 11: This defeats the purpose of using an IoC Container. We should only import generic/abstract classes in the implementation.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -42,7 +42,8 @@ const Checkout = () => {
       onSubmitPaypalPaymentSuccess: ({ redirectUrl }) => {
         window.location.href = redirectUrl;
       },
-      onSubmitStripePaymentSuccess: () => navigate(modalURLFromLocation(location, 'waiting-for-payment'), { replace: true }),
+      onSubmitStripePaymentSuccess: () =>
+        navigate(modalURLFromLocation(location, 'waiting-for-payment', { offerId: selectedOffer?.planOriginalId }), { replace: true }),
```
ChristiaanScheermeijer: This is not compatible with generic integrations

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -102,7 +103,7 @@ const Checkout = () => {
   }
 
   const cancelUrl = modalURLFromWindowLocation('payment-cancelled');
-  const waitingUrl = modalURLFromWindowLocation('waiting-for-payment', { offerId: selectedOffer?.offerId });
+  const waitingUrl = modalURLFromWindowLocation('waiting-for-payment', { offerId: selectedOffer?.planOriginalId });
```
ChristiaanScheermeijer: This is also not compatible with generic integrations

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
mirovladimitrovski: I'm aware of this, the fix was necessary for the previous implementation of offers with the JWP integration, but with the implementation introduced with this PR it's no longer needed.

packages/hooks-react/src/useEntitlement.ts
mirovladimitrovski: The service in question (`JWPEntitlementService`) was already used the same way in the `useContentProtection` hook: https://github.com/jwplayer/ott-web-app/blob/38f8a637489c497a07551450323dd21df0fc66a1/packages/hooks-react/src/useContentProtection.ts#L6

Since this PR is already big as it is and its scope isn't re-evaluating that implementation, I thought I'd just go along with it, because the contrary would've meant refactoring it with the possibility of breaking it.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
mirovladimitrovski: Good catch.

ChristiaanScheermeijer:
I'm a bit hesitant of some parts of this PR because it doesn't leverage the existing IoC pattern but adds another layer of complexity on top. I think our goal should be to make the services/integrations more simple and flexible by generalising and abstracting implementation logic out of the hooks and views.

By skipping this now, it will make it even more difficult to simplify the integrations later.

**Simplified flow of the current situation**

```mermaid
graph TD
    subgraph ChooseOffers
        A[View]
        B[useOffers Hook]
        C[CheckoutController]
        D[IoC Container]
        E[Cleeng Service]
        F[JWP Service]
    end

    A --> B
    B --> C
    C --> D
    D --> E
    D --> F
    C == Offers ==> B
    B == Offers ==> A

    style E fill:#ff9,stroke:#333,stroke-width:4px,color:black
    style F fill:#9ff,stroke:#333,stroke-width:4px,color:black
```

**After the plans addition**

```mermaid
graph TD
    subgraph ChooseOffers
        A[View]
        AA[Choose Offer Component]
        AB[List Plans Component]
        B[useOffers Hook]
        BB[usePlan Hook]
        C[CheckoutController]
        D[IoC Container]
        E[Cleeng Service]
        F[JWP Service]
    end

    A --> AA
    A --> AB

    AB --> BB
    BB --> F

    AA --> B
    
    B --> C
    C --> D
    D --> E
    D --> F
    C == Offers ==> B
    B == Offers ==> A

    style E fill:#ff9,stroke:#333,stroke-width:4px,color:black
    style F fill:#9ff,stroke:#333,stroke-width:4px,color:black
```

I don't think it needs a lot more work to improve this. Let me know what you think 😄

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
mirovladimitrovski: Also a good catch.

packages/ui-react/src/components/Button/Button.tsx
```diff
@@ -29,6 +29,7 @@ type Props = {
   busy?: boolean;
   id?: string;
   activeClassname?: string;
+  navLinkState?: any;
```
ChristiaanScheermeijer: We are using `strict` mode, so `any` is not allowed 😃

packages/ui-react/src/components/ListPlans/ListPlans.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const mediaId = match?.params.id || '';
+
+  useLayoutEffect(() => {
+    if (mediaId) {
+      queryClient.invalidateQueries(['plans', mediaId]);
+    }
+  }, [mediaId, queryClient]);
+
+  const { isLoading, data: plans } = usePlansForMedia(mediaId);
```
ChristiaanScheermeijer: This is container logic and shouldn't be added to components

packages/ui-react/src/containers/AccountModal/forms/CancelSubscription.tsx
```diff
@@ -48,17 +48,25 @@ const CancelSubscription = () => {
...
+    } else {
+      closeHandler();
+    }
+  };
+
   if (!subscription) return null;
 
   return (
-    <React.Fragment>
+    <div style={{ padding: '16px' }}>
```
ChristiaanScheermeijer, line 62: How come this is needed?

packages/ui-react/src/components/Button/Button.tsx
mirovladimitrovski: This is a prop we're forwarding to `<NavLink state={navLinkState} />` (`NavLink` from React Router) and the type of the `state` prop is `any`. This is the only reason I would ever use `any` like this. 😃

Edit: Maybe explaining it in a comment would be the way to go with this.

packages/ui-react/src/components/ListPlans/ListPlans.tsx
mirovladimitrovski: I wasn't aware we were following a rule like this, but it makes sense and I'll make the necessary changes. Thank you.

packages/ui-react/src/containers/AccountModal/forms/CancelSubscription.tsx
mirovladimitrovski: Not exactly needed, it was just a remark from a designer for the sake of better visuals.

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
ChristiaanScheermeijer: Do you mean that the previous integration can't/shouldn't be used anymore?

packages/hooks-react/src/useEntitlement.ts
ChristiaanScheermeijer: Ah that's right. That one isn't generalised yet unfortunately. 

I do wonder why it needs the additional access check? Isn't this covered by the `CheckoutService#getEntitlements` call?

mirovladimitrovski:
> I'm a bit hesitant of some parts of this PR because it doesn't leverage the existing IoC pattern but adds another layer of complexity on top. I think our goal should be to make the services/integrations more simple and flexible by generalising and abstracting implementation logic out of the hooks and views.
> 
> By skipping this now, it will make it even more difficult to simplify the integrations later.
> 
> **Simplified flow of the current situation**
> 
> ```mermaid
> graph TD
>     subgraph ChooseOffers
>         A[View]
>         B[useOffers Hook]
>         C[CheckoutController]
>         D[IoC Container]
>         E[Cleeng Service]
>         F[JWP Service]
>     end
> 
>     A --> B
>     B --> C
>     C --> D
>     D --> E
>     D --> F
>     C == Offers ==> B
>     B == Offers ==> A
> 
>     style E fill:#ff9,stroke:#333,stroke-width:4px,color:black
>     style F fill:#9ff,stroke:#333,stroke-width:4px,color:black
> ```
> **After the plans addition**
> 
> ```mermaid
> graph TD
>     subgraph ChooseOffers
>         A[View]
>         AA[Choose Offer Component]
>         AB[List Plans Component]
>         B[useOffers Hook]
>         BB[usePlan Hook]
>         C[CheckoutController]
>         D[IoC Container]
>         E[Cleeng Service]
>         F[JWP Service]
>     end
> 
>     A --> AA
>     A --> AB
> 
>     AB --> BB
>     BB --> F
> 
>     AA --> B
>     
>     B --> C
>     C --> D
>     D --> E
>     D --> F
>     C == Offers ==> B
>     B == Offers ==> A
> 
>     style E fill:#ff9,stroke:#333,stroke-width:4px,color:black
>     style F fill:#9ff,stroke:#333,stroke-width:4px,color:black
> ```
> I don't think it needs a lot more work to improve this. Let me know what you think 😄

With the introduction of support for multiple plans which this PR is implementing, we have a revised layout for `plans`, but we're keeping the current layout for `offers` (meaning we don't want to change the UI for Cleeng users). This meant we had to look at plans and offers as two separate concepts, as a cleaner alternative to blending them together in a single common layout (such as the `ChooseOfferForm` component). But in order to avoid the trap of accessing `JW` or `Cleeng` services directly, we're introducing a generic variable called `accessMethod` which can either be `plan` or `offer`, although they implicitly mean `JWP` and `Cleeng` respectively at the moment. I agree I've made an exception to the rule in my new hook that I'm introducing - `usePlansForMedia`, which I'm going to fix right away. But I think that's as far as we can go in terms of being generic, because the said distinction between plans and offers is generic enough and from that point on plans and offers are inevitably going to have their own exclusive hooks and, as it turns out, their own exclusive components as well. Anyway, I'm open to suggestions.

packages/common/src/modules/register.ts
```diff
@@ -82,6 +82,7 @@ container.bind(SubscriptionService).to(CleengSubscriptionService).whenTargetName
 // JWP integration
 container.bind(DETERMINE_INTEGRATION_TYPE).toConstantValue(isJwpIntegrationType);
 container.bind(JWPEntitlementService).toSelf();
+container.bind(JWPCheckoutService).toSelf().whenTargetNamed(INTEGRATION.JWP);
```
ChristiaanScheermeijer: This seems wrong 🤔

packages/hooks-react/src/usePlansForMedia.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const extractCustomParameters = (playlistItem: PlaylistItem): { key: string; value: unknown }[] => {
+  const excludedKeys = Object.keys(samplePlaylistItem);
+  return Object.entries(playlistItem)
+    .filter(([key]) => !excludedKeys.includes(key))
+    .map(([key, value]) => ({ key, value }));
+};
+
+export default function usePlansForMedia(_mediaId: string) {
+  const [mediaId, setMediaId] = useState(_mediaId);
```
ChristiaanScheermeijer: I don't think this `state` and `useEffect` is really needed, correct?

packages/ui-react/src/components/Button/Button.tsx
ChristiaanScheermeijer: Yeah, but I believe running `yarn lint:ts` will fail when using `any`, right? Probably need to change it to `unknown`

packages/hooks-react/src/usePlansForMedia.tsx
```diff
@@ -0,0 +1,126 @@
...
+      const customParameters = extractCustomParameters(mediaData);
+
+      const searchString = `q=${buildFilterQuery(tags, customParameters)}`;
+
+      checkoutController.siteId = siteId;
+
+      return await checkoutController.getPlansWithPriceOffers(searchString);
+    },
+    enabled: !!checkoutController && !!mediaData,
+  });
```
ChristiaanScheermeijer: Hoping that we can avoid having additional hooks specific for plans

packages/hooks-react/src/useEntitlement.ts
mirovladimitrovski: The problem is that offers are standalone entities, while plans are just containers of prices. Unlike offers, which are holders of access/entitlement, with the case of plans the access/entitlement is attached to the plan, rather than to the price. For this reason the mapping between offers and plans isn't consistent throughout the code - somewhere offer means plan (i.e. when we need to check for access), somewhere offer means plan price (i.e. when we need to list them visually). So, one way of checking for access is iterating through every plan, making a request for each of them until we find which plan we have access for, which is what `getEntitlements` would do if we utilised it for the plans implementation. A quicker and more optimised way for this is to simply make a request with `mediaId` and let the backend do the said iteration through the plans and then let us know if we have access for any of them, which is what `getJWPMediaToken` does. It's confusing, even more so with the constraints that go along with the need for generalisation.

packages/hooks-react/src/usePlansForMedia.tsx
mirovladimitrovski: Ah yes, good catch, this is a left-over from a workaround from a few commits ago.

packages/hooks-react/src/useEntitlement.ts
ChristiaanScheermeijer: Thanks for the explanation. That does make things clear 😄

But I think that I still prefer to refactor the `getEntitlements` to make it fit the new APIs. Perhaps we should draw this out, but I don't have access to all the new APIs.

One thing that comes to mind is that instead of asking for entitlements for offer with id X, we could pass the full playlist item to the `getEntitlements` function. The service can then use whatever information it needs to figure out if the user has access.

mirovladimitrovski: I understand what you mean and I would've definitely done that by now, but what we have is two totally different react queries with different variables. Just their query keys are indicative of that, where on the one hand we have `['token', 'media', playlistItem?.mediaid, {}]`, and on the other we have an array of queries with keys `['entitlements', offerId]`. Trying to reconcile them into a unifying one seems like too far beyond the scope of this feature - it would mean refactoring the existing implementation with the risk of breaking it, and I'm not even sure I understand why it has to be an array of `['entitlements', playlistItem.mediaOffers[i]]` query keys instead of a single one with `['entitlements', playlistItem.mediaid]`.

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
mirovladimitrovski: No, I just reverted back to the `parseOfferId` and `formatOffer` methods the way they were done with your fix.

packages/hooks-react/src/useEntitlement.ts
mirovladimitrovski: I even forgot to bring up that the JWP integration still needs `getEntitlements` the way it is, meaning still being able to check for entitlement by `plan.original_id` (InPlayer's `asset.id`), it uses it from the `useCheckAccess` hook. I think we'll have to have two methods for checking for entitlements generically - `getEntitlementByAccessItem` (`access item` as a generic name for *offer* or *plan*) and `getEntitlementByMedia`, but I would leave it for a separate task.

packages/ui-react/src/components/Button/Button.tsx
mirovladimitrovski: It doesn't fail. But I did leave a comment.

packages/ui-react/src/containers/AccountModal/forms/PersonalDetails.tsx
```diff
@@ -40,10 +40,12 @@ const PersonalDetails = () => {
   );
 
   const nextStep = useCallback(() => {
-    const hasOffers = accessModel === ACCESS_MODEL.SVOD || (accessModel === ACCESS_MODEL.AUTHVOD && hasMediaOffers);
+    const hasSubscriptionPlansOrOffers = !!subscriptionOffers.length;
+
+    const hasOffers = accessModel === ACCESS_MODEL.SVOD || (accessModel === ACCESS_MODEL.AUTHVOD && hasMediaOffers) || hasSubscriptionPlansOrOffers;
```
ChristiaanScheermeijer, line 45: Are you sure this change is needed? When the access model is SVOD, we can already assume that there are subscription offers that will be fetched in the next step. Or are we not configuring global SVOD plans in the new setup?

mirovladimitrovski: There might be no subscription offers yet, in which case it makes sense to skip the `choose-offer` modal and go straight to `welcome`.

ChristiaanScheermeijer:
@mirovladimitrovski what is your timeline for this feature? Do you have a config ID that we can use to test the new plans feature? 😄

PS. Don't take my feedback as resistance to this change, I mainly worry about features being removed or breaking changes which has a direct impact for users of this repo (including us). I'm hoping that we can do these changes in a "next" or "v7" branch because we're basically removing the existing integration and forcing users to migrate.

When working in a different branch it will also become easier to refactor services and integrations to shape them around the plans model. In theory both Cleeng and InPlayer can be used with plans, right?

mirovladimitrovski:
> @mirovladimitrovski what is your timeline for this feature?
>
As soon as possible. This feature got to be a lot more complex on the OTT app and took some more dev time than expected, so we're a even a little behind schedule.

> Do you have a config ID that we can use to test the new plans feature? 😄
>
I'm mostly using `aecedypm` (with access model `free`) and `wxpmaqis` (with access model `svod`), but the environment in this PR is temporarily switched to `dev`, so you can conveniently test with any config from https://dev-dashboard.jwplayer.com/. The appropriate changes are already done in the dashboard, so you can create apps and play with plans and prices you want to test as you see fit. Let me know if I can help with anything.

> PS. Don't take my feedback as resistance to this change,
>
I don't, no worries. :) You're giving a valuable feedback and I appreciate it.

> I mainly worry about features being removed or breaking changes which has a direct impact for users of this repo (including us). I'm hoping that we can do these changes in a "next" or "v7" branch because we're basically removing the existing integration and forcing users to migrate.
>
We're not removing any existing integration at all. If you're referring to the previous implementation of JWP plans, we're not removing that either. :) We're only expanding it from supporting a single plan to supporting multiple plans per app. In any case, the `offers` integration is supposed to remain intact.

> When working in a different branch it will also become easier to refactor services and integrations to shape them around the plans model. In theory both Cleeng and InPlayer can be used with plans, right?
>
In theory, yes.

PR #551, opened 2024-06-12
Fix / Generic consents validation message & show message after submission

langemike:
I introduced a bug with this PR https://github.com/jwplayer/ott-web-app/pull/537

## Bug
After my change, the error message always showed up initially (before form submission) and I noticed that the `<CustomRegisterField />` can be used for all types of fields. In the situation I tested it with I have only seen checkboxes.

## Fix
I reused the already present (generic) `field_required` error message which equals the message "This field is required" without any specificity or guidance.  It originally was "This field is required, please fill in {{field}} to continue.", but the field name can be "technical" such as "broadcaster_terms" which we considered an issue. We originally wanted the extra information (such as _please fill in_ or _check the box_) to guide users, but keeping this intact for all types but this requires a lot of work and translation maintenance. In the end we think (and hope) the message is clear enough.

I also wrote an e2e test to prevent the bug from happening.

Ticket: https://videodock.atlassian.net/browse/OTT-1928

### DIFF
--- a/packages/ui-react/src/components/RegistrationForm/RegistrationForm.tsx
+++ b/packages/ui-react/src/components/RegistrationForm/RegistrationForm.tsx
@@ -126,7 +126,7 @@ const RegistrationForm: React.FC<Props> = ({
                 value={consentValues[consent.name] || ''}
                 required={consent.required}
                 error={!!consentError}
-                helperText={t('registration.consent_required')}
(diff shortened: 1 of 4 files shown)


ChristiaanScheermeijer (approved):
LGTM!

royschut (approved):
👍

MelissaDTH (approved):

PR #555, opened 2024-06-12
Fix/ Remove features from web app

borkopetrovicc:
## Description

https://jwplayer.atlassian.net/jira/software/c/projects/PS/boards/575?selectedIssue=PS-242

This PR solves # .

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


packages/ui-react/src/components/Account/Account.tsx
```diff
@@ -384,46 +384,6 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
                 </div>
               ),
             }),
-          canExportAccountData &&
-            formSection({
-              label: t('account.export_data_title'),
-              content: (section) => (
-                <div className={styles.textWithButtonContainer}>
-                  <div>
-                    <Trans t={t} i18nKey="account.export_data_body" values={{ email: section.values.email }} />
```
AntonLantukh, line 393: Do we still need translations for delete and export?

packages/ui-react/src/components/Account/Account.tsx
```diff
@@ -72,7 +72,7 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
     shallow,
   );
 
-  const { canChangePasswordWithOldPassword, canExportAccountData, canDeleteAccount } = accountController.getFeatures();
+  const { canChangePasswordWithOldPassword, canExportAccountData } = accountController.getFeatures();
```
AntonLantukh: Do we still need "canExportAccountData"?

borkopetrovicc: 01a95c9ceda9cf3127a3fe29a2af2a8109642d2e

AntonLantukh:
@borkopetrovicc it looks like tests are failing because of the code that was removed. Can you look into it?

AntonLantukh:
@borkopetrovicc it looks looks good to me in general but I didn't work a lot with the payments logic. @ChristiaanScheermeijer any concerns you have here?

ChristiaanScheermeijer:
I'm afraid this also removal doesn't take the Cleeng integration into consideration. We want to keep that logic intact as much as possible.

Also, when using Cleeng, I now see two cancel subscription buttons:

image

borkopetrovicc:
> I'm afraid this also removal doesn't take the Cleeng integration into consideration. We want to keep that logic intact as much as possible.
> 
> Also, when using Cleeng, I now see two cancel subscription buttons:
> 
> image

What do you suggest ?

borkopetrovicc:
> I'm afraid this also removal doesn't take the Cleeng integration into consideration. We want to keep that logic intact as much as possible.
> 
> Also, when using Cleeng, I now see two cancel subscription buttons:
> 
> image

8da8eadb1203eebfccc7b980f0abd52455b2706d

packages/ui-react/src/components/Account/Account.tsx
```diff
@@ -72,7 +62,7 @@ const Account = ({ panelClassName, panelHeaderClassName, canUpdateEmail = true }
     shallow,
   );
 
-  const { canChangePasswordWithOldPassword, canExportAccountData, canDeleteAccount } = accountController.getFeatures();
+  const { canChangePasswordWithOldPassword } = accountController.getFeatures();
```
ChristiaanScheermeijer, line 65: This mechanism is in place to disable these features for specific integrations. Why would you remove this mechanism and implementation?

Just set `canExportAccountData` and `canDeleteAccount` to false and done, right? Custom integrations can still implement these features.

packages/ui-react/src/components/Payment/Payment.tsx
```diff
@@ -219,78 +201,18 @@ const Payment = ({
...
-                  />
-                ))}
-              <div className={styles.changePlanButtons}>
-                <Button
-                  label={t('user:account.save')}
-                  onClick={onChangePlanClick}
-                  disabled={selectedOfferId === activeSubscription?.accessFeeId || changeSubscriptionPlan.isLoading}
-                />
-                <Button label={t('user:account.cancel')} onClick={() => setIsChangingOffer(false)} variant="text" />
-                {activeSubscription?.status !== 'cancelled' && (
```
ChristiaanScheermeijer: This removes the following features for **all** integrations:

- Downgrade/upgrade subscription
- Renew subscription when it has been cancelled
- Complete subscription when user is not subscribed

I don't think we want to do that (yet)... @dbudzins @AntonLantukh ?

ChristiaanScheermeijer (changes requested):

ChristiaanScheermeijer:
@dbudzins perhaps we could add some more description/definition to the PR descriptions like we do? That helps us to better understand the desired outcome and review this more efficient.

packages/ui-react/src/components/Payment/Payment.tsx
borkopetrovicc: I will close these pr i've opened new pr

PR #556, opened 2024-06-13
feat(project): add new type of lists (recommendations)

AntonLantukh:
## Description

1. Add new "content_list" type support (recommendations).
2. Support it for Landing Page and Menu.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


AntonLantukh:
App Config for testing: `0i9cflkq`

packages/ui-react/src/pages/ScreenRouting/RecommendationsScreenRouter.tsx
```diff
@@ -0,0 +1,42 @@
...
+import ErrorPage from '../../components/ErrorPage/ErrorPage';
+import type { ScreenComponent } from '../../../types/screens';
+
+import PlaylistGrid from './sharedScreens/PlaylistGrid/PlaylistGrid';
+
+export const playlistScreenMap = new ScreenMap<Playlist, ScreenComponent<Playlist>>();
+
+playlistScreenMap.registerDefault(PlaylistGrid);
+
+const RecommendationsScreenRouter = () => {
```
ChristiaanScheermeijer: Do we need a separate router or could we re-use `<PlaylistScreenRouter />` and pass the list type along from a prop?

ChristiaanScheermeijer (commented):
The PR overall looks good, but I do have some questions:

- ~Do you have content list IDs to test with?~ You just added it 😄 
- Do we need to add e2e testing for this?
- Do we really need an abstraction for the content and playlist APIs or could it just be two separate services?
  - This could prevent some breaking changes in the `usePlaylist` hook for example
- Should we use different names for the service/controller/types?
  - Content list can be confused with contentTypes
  - `RecommendationsService` sounds like it's used for different purposes
  - I suggest renaming it to `PlaylistController`, `PlaylistService`, `PlaylistContentService` and `PLAYLIST_TYPE` removing the content analogy from most most places

packages/ui-react/src/pages/ScreenRouting/RecommendationsScreenRouter.tsx
AntonLantukh: So for menu items you think we could pass 'type' as a query param? We could reuse it. I was just thinking that it could be cleaner when we use a separate router. We can also tweak translations and avoid using LiveChannels screen.

AntonLantukh:
@ChristiaanScheermeijer 

 > Do we need to add e2e testing for this?

We do need. I will add it to the Q3 plan. 

> Do we really need an abstraction for the content and playlist APIs or could it just be two separate services? 
> This could prevent some breaking changes in the `usePlaylist` hook for example.

Do you mean having two hooks "usePlaylist" and "usePlaylistContent" which could be called conditionally + removing the controller completely? We still need to know the type to make a decision, so there will be also a "parent" hook receiving it as an argument.

> Should we use different names for the service/controller/types?

I agree naming is a problem. My idea was to consume playlists, content lists (recommendations), series using controller in the future. That is why I used "ContentController" name. We also have some plans to deliever an additional "media" endpoint returning data based on the content type (maybe series, stream data inside). Though it can be also overengineering to consider this case in advance and to avoid braking changes we could start with conditional calls.

packages/common/src/services/content/RecommendationsService.ts
budzworthy, line 1: Do we need to differentiate content lists from recommendations lists?

packages/common/src/services/content/ContentService.ts
budzworthy, line 1: I think we should avoid putting the Media vs. Content abstraction into the web app for now. It adds an extra layer above playlist and content lists that may make the sample card harder to understand.

packages/common/src/services/content/PlaylistService.ts
```diff
@@ -0,0 +1,58 @@
...
+    const response = await fetch(url);
+    const data = (await getDataOrThrow(response)) as Playlist;
+
+    return this.transformPlaylist(data, params.related_media_id);
+  };
+
+  /**
+   * Get playlist by id
+   */
+  getContentList = async ({ id, params }: { id: string | undefined; params: GetContentListParams }): Promise<Playlist | undefined> => {
```
budzworthy: I'm not sure this boilerplate is adding value. I would prefer to keep the getPlaylist and any other new api endpoints in the ApiService.

budzworthy (changes requested):
I'm still going through some parts, but I think we can keep this change simpler without adding the content service abstractions. If we add the new content list endpoint to the existing Api Service, we can do switching at the hook or controller level to get playlists or content lists as needed.

packages/common/src/services/content/ContentService.ts
AntonLantukh: I personally like that we define the contract in the Service and let the controller select the right endpoint. It adds complexity but looks clean. Though I also agree that maybe it is still early for more global refactoring here.

AntonLantukh:
@ChristiaanScheermeijer @dbudzins simplified approach without abstraction layer.

packages/common/src/services/ApiService.ts
```diff
@@ -44,20 +45,64 @@ export default class ApiService {
...
+  };
+
+  /**
+   * Transform incoming playlists
+   */
+  private transformPlaylist = (playlist: Playlist, relatedMediaId?: string) => {
+    playlist.playlist = playlist.playlist.map((item) => this.transformMediaItem(item, playlist));
+
+    // remove the related media item (when this is a recommendations playlist)
+    if (relatedMediaId) playlist.playlist.filter((item) => item.mediaid !== relatedMediaId);
```
budzworthy: https://www.imperialviolet.org/2014/02/22/applebug.html
Also, I think you need to assign for filter to have an effect.
```suggestion
    if (relatedMediaId) {
      playlist.playlist = playlist.playlist.filter((item) => item.mediaid !== relatedMediaId);
    }
```

packages/common/types/config.ts
```diff
@@ -39,20 +41,20 @@ export type Drm = {
...
+  type: PlaylistType;
   featured?: boolean;
   backgroundColor?: string | null;
 };
 
 export type Menu = {
   label: string;
   contentId: string;
-  type?: Extract<ContentType, 'playlist'>;
+  type?: Extract<PlaylistType, 'playlist' | 'content_list'>;
```
budzworthy: This can be easier since you have a type for it now:
```suggestion
  type?: PLAYLIST_TYPE.playlist | PLAYLIST_TYPE.content_list;
```

packages/common/types/playlist.ts
```diff
@@ -27,41 +25,15 @@ export type PlaylistItem = {
...
-  scheduleDemo?: string;
-  catchupHours?: string;
-  mediaStatus?: MediaStatus;
-  scheduledStart?: Date;
-  scheduledEnd?: Date;
-  markdown?: string;
-  scheduleType?: string;
-  [key: string]: unknown;
-};
+} & CustomParams;
```
budzworthy, line 36: Nice!

packages/common/src/utils/urlFormatting.ts
```diff
@@ -103,8 +103,8 @@ export const mediaURL = ({
   return createPath(PATH_MEDIA, { id: media.mediaid, title: slugify(media.title) }, { r: playlistId, play: play ? '1' : null, e: episodeId });
 };
 
-export const playlistURL = (id: string, title?: string) => {
-  return createPath(PATH_PLAYLIST, { id, title: title ? slugify(title) : undefined });
+export const playlistURL = ({ id, title, type }: { id: string; title?: string; type?: 'playlist' | 'content_list' }) => {
+  return createPath(PATH_PLAYLIST, { id, title: title ? slugify(title) : undefined }, { t: type });
```
budzworthy: I would keep this as succinct as possible and try to avoid changing the existing playlist routing:
```suggestion
  return createPath(PATH_PLAYLIST, { id, title: title ? slugify(title) : undefined }, type === 'content_list' ? { t: 'c' } : {});
```

packages/hooks-react/src/usePlaylists.ts
budzworthy, line 1: We should probably only have usePlaylist (singular) or usePlaylists (plural) otherwise there's a big risk of code divergence (seen here with different query keys whether content lists are retrieved with this hook or the useContentList hook)

packages/hooks-react/src/usePlaylistContent.ts
budzworthy, line 1: I think the approach you took in usePlaylists was probably more maintainable. 1 hook that has a tiny bit of playlist and content list logic is better than 3 hooks.

packages/ui-react/src/pages/ScreenRouting/sharedScreens/PlaylistGrid/PlaylistGrid.tsx
budzworthy, line 1: I would keep the 2 concepts of media and lists here, instead of adding shared. Shared can be confusing because it's unclear if it's shared between media and playlist, media and content list, or playlist and content list. If you want to rename playlistScreens to listScreens that's fine as a compromise.

packages/common/src/services/ApiService.ts
AntonLantukh: Hm, looks like this is the way it worked before. I will assign it.

packages/ui-react/src/pages/ScreenRouting/sharedScreens/PlaylistGrid/PlaylistGrid.tsx
AntonLantukh: I think this is what is left after removing "RecsScreen". I will move it to Playlists.

packages/common/src/services/ApiService.ts
budzworthy: Probably just a bug then.. good to fix while we're here

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
```diff
@@ -37,6 +36,12 @@ const PlaylistScreenRouter = () => {
     return <ErrorPage title={t('empty_shelves_heading')} message={t('empty_shelves_description')} />;
   }
 
+  // register playlist screens
+  playlistScreenMap.registerDefault(PlaylistGrid);
+  if (type === PLAYLIST_TYPE.playlist) {
+    playlistScreenMap.registerByContentType(PlaylistLiveChannels, PLAYLIST_CONTENT_TYPE.live);
+  }
```
ChristiaanScheermeijer: This feels a bit hacky, why is this needed?

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
```diff
@@ -2,27 +2,26 @@ import React from 'react';
...
-// register playlist screens
-playlistScreenMap.registerByContentType(PlaylistLiveChannels, CONTENT_TYPE.live);
-playlistScreenMap.registerDefault(PlaylistGrid);
-
 const PlaylistScreenRouter = () => {
   const params = useParams();
   const id = params.id || '';
-  const { isLoading, isFetching, error, data } = usePlaylist(id);
+  const type = useQueryParam('t') === 'c' ? PLAYLIST_TYPE.content_list : PLAYLIST_TYPE.playlist;
+
```
ChristiaanScheermeijer: Hmm, I have doubts about using query params for this. Do we include this in the canonical URL?

Alternatively, we could do this:

```tsx
export const PATH_PLAYLIST_CONTENT = '/p/:id/c/*' // or something similar

// ...

<Route path={PATH_PLAYLIST_CONTENT} element={<PlaylistScreenRouter type="content" />} />
<Route path={PATH_PLAYLIST} element={<PlaylistScreenRouter type="playlist" />} />
```

packages/hooks-react/src/usePlaylist.ts
```diff
@@ -1,42 +1,84 @@
...
+  type,
+  params = {},
+  enabled = true,
+  usePlaceholderData = true,
+}: {
+  contentId?: string;
+  type: PlaylistMenuType;
+  params?: GetPlaylistParams;
+  enabled?: boolean;
+  usePlaceholderData?: boolean;
```
ChristiaanScheermeijer: Wouldn't it make more sense to leave the `usePlaylist` as is and introduce a `useContentList` hook?

ChristiaanScheermeijer: I'm now seeing @dbudzins feedback. Although I agree to that, this change will result in many changes and could be considered a breaking change. I would very much like to prevent that, but it's not a blocking issue.

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
AntonLantukh: I think we would prefer to avoid changing playlist routing, so I was choosing between a separate screen routing and a query param. If we change the path, then I think we should also change it for the playlist.

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
AntonLantukh: Not to register the screen for content_list where live screen is not supported yet. For now we define the screen based on the item itself, but the item doesn't tell anything about the type (though maybe we could add additional "type" property for both content list and playlists).

packages/hooks-react/src/usePlaylist.ts
AntonLantukh: That was the initial solution here:

https://github.com/jwplayer/ott-web-app/pull/556/commits/39c9f981fc88c9f04169018167224415c949102a

Would you prefer this one?

@dbudzins Danny suggested to use "usePlaylist" directly without additional hook. Not sure I have any strong preference here. I used hook approach with search before, worked fine.

AntonLantukh: Breaking change is a valid point here. Maybe we can start with hooks here and refactor usePlaylist / usePlaylists later.

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
budzworthy: Probably safest to keep them as separate paths. Why not just `/p/:id/*` and `/n/:id/*` (using n for coNtent list, because we might want to reserve c for individual content items if we ever get to that refactor.) I worry `/p/:id/c/*` will have collisions with `/p/:id/*` if we aren't careful.

packages/hooks-react/src/usePlaylist.ts
budzworthy: What is the breaking change? If it's just the addition of the type param, can we default it to `playlist` if not provided?

ChristiaanScheermeijer: We consume these hooks in all platforms. The breaking change is that the hook signature changes. It would be safe if it was:

```ts
export default function usePlaylist(playlistId?: string, params: GetPlaylistParams = {}, enabled: boolean = true, usePlaceholderData: boolean = true, type: PlaylistMenuType = PLAYLIST_TYPE.playlist) {
```

Although this require adding all optional arguments with the default values when only changing the type.

It's not a huge deal, but something to be aware of.

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
ChristiaanScheermeijer: That works for me 😄

packages/ui-react/src/pages/ScreenRouting/PlaylistScreenRouter.tsx
AntonLantukh: It will work then this way (out of component's scope):

```
// register playlist screens
playlistScreenMap.registerDefault(PlaylistGrid);
playlistScreenMap.register(PlaylistLiveChannels, (item) => item?.contentType === PLAYLIST_CONTENT_TYPE.live && item.type === PLAYLIST_TYPE.playlist);

```

ChristiaanScheermeijer: We don't need a separate router screen per se, but perhaps we can create a separate screen map for content? 

```ts
const screenMap = type === 'playlist' ? playlistScreenMap : contentScreenMap;
```

The `contentScreenMap` only contains the default screen using the PlaylistGrid.

AntonLantukh:
@ChristiaanScheermeijer @dbudzins 

1. Switched to a separate endpoint for content lists.
2. Reused PlaylistScreen for content lists.
3. Changed interface of `usePlaylist` and added optional params (was needed in one place).
4. Separated query options into a separate function to reuse in `usePlaylists` / `usePlaylist`

ChristiaanScheermeijer (approved):
Perfect @AntonLantukh! This looks good to me.

Only the `playlistURL` method has changed, but that's only used once in our platforms 😄

AntonLantukh:
@ChristiaanScheermeijer rolled back `playlistURL` to the previous interface.

packages/hooks-react/src/usePlaylist.ts
```diff
@@ -1,42 +1,78 @@
...
+  enabled: boolean = true,
+  usePlaceholderData: boolean = true,
+  type: PlaylistMenuType = PLAYLIST_TYPE.playlist,
+) {
+  const queryClient = useQueryClient();
+  const siteId = useConfigStore((state) => state.config.siteId);
+
+  const queryOptions = getPlaylistQueryOptions({ type, contentId, siteId, params, queryClient, enabled, usePlaceholderData });
+
+  return useQuery<Playlist | undefined, ApiError>(queryOptions);
```
budzworthy, line 77: You've gotta add usePlacehoderData and params inputs to usePlaylists, but then you avoid the getQueryOptions construct, which is a bit non-standard I think.

(We can do this later. I know you want to get the changes out, and the double playlist hooks wasn't added from this PR)

```suggestion
  return usePlaylists([{contentId, type}], enabled ? 1 : 0, usePlaceholderData, params)[0];
```

platforms/web/src/containers/AppRoutes/AppRoutes.tsx
```diff
@@ -24,6 +24,7 @@ export default function AppRoutes() {
         <Route element={<Layout />} errorElement={<RootErrorPage />}>
           <Route index element={<Home />} />
           <Route path={PATH_PLAYLIST} element={<PlaylistScreenRouter />} />
+          <Route path={PATH_CONTENT_LIST} element={<PlaylistScreenRouter />} />
```
budzworthy: I would suggest we either use @ChristiaanScheermeijer 's suggestion to have the `type` as a param on PlaylistScreenRouter or alternatively, use a regex for the path and share one <Route> for playlist and content list. I think the former is preferred, because you don't need the extra path match checking inside PlaylistScreenRouter.

budzworthy (approved):
2 small comments. The one about the Routes and PlaylistScreenRouter component, I'd strongly suggest doing, but I'll approve as-is, because I know you're pressed for time.

packages/hooks-react/src/usePlaylist.ts
AntonLantukh: I didn't want to do it because of the different use case we have here. usePlaylists is intended to work with a list of shelves on the Landing Page. So it also covers the cases of Favorites and Continue Watching, while usePlaylist only supports `content_list` and `playlist` types which makes the purpose a bit different.

budzworthy (approved):
lgtm

PR #562, opened 2024-06-27
Fix/ Remove features from web app 

borkopetrovicc:
## Description

https://jwplayer.atlassian.net/jira/software/c/projects/PS/boards/575?selectedIssue=OWA-78

This PR solves # .

### Steps completed:

According to our definition of done, I have completed the following steps:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Docs updated (including config and env variables)
- [ ] Translations added
- [ ] UX tested
- [ ] Browsers / platforms tested
- [ ] Rebased & ready to merge without conflicts
- [ ] Reviewed own code

PR #563, opened 2024-06-27
Fix / remove profiles from the web app

borkopetrovicc:
## Description
[
](https://jwplayer.atlassian.net/jira/software/c/projects/PS/boards/575?selectedIssue=OWA-78)

https://www.figma.com/design/qLl8KxBOOr640h18ICx0H6/Web-app?node-id=3004-9048&t=RpZJoBJfnVp5w6Us-0

1.Disable "Export account" feature for JWP
2.Disable "Delete account" feature for JWP
3.Disable "Switch plans" for JWP

This PR solves # .

### Steps completed:

According to our definition of done, I have completed the following steps:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Docs updated (including config and env variables)
- [ ] Translations added
- [ ] UX tested
- [ ] Browsers / platforms tested
- [ ] Rebased & ready to merge without conflicts
- [ ] Reviewed own code

PR #565, opened 2024-07-01
Refactor / align payment (upgrade) integration flows

ChristiaanScheermeijer:
## Description

This proposal PR is my attempt to align the subscription switch logic discrepancy between the JWP and Cleeng integration. These two features (doing the same) were implemented differently.

Motivated by #562 and #555 I'm hoping we can simplify flows and eliminate differences between providers. To disable features, we should be using the integration feature flags to keep the code intact for other providers (as long we don't have alternatives).


packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
```diff
@@ -276,13 +283,50 @@ export default class JWPCheckoutService extends CheckoutService {
...
+
+    try {
+      const response = await InPlayer.Subscription.changeSubscriptionPlan({
+        access_fee_id: accessFeeId,
+        inplayer_token: String(subscription.subscriptionId),
+      });
+
+      await this.accountService.updateCustomer({
+        metadata: {
+          [`${subscription.subscriptionId}_pending_downgrade`]: toOfferId,
```
AntonLantukh, line 308: Oh...this part was in the hook previously. Makes sense to have it here.

borkopetrovicc:
Screenshot 2024-07-03 at 13 28 40

I see the Renew subscription button on JWP side. This is not aligned wit our design where we should only have cancel subscription you can see in the link: https://www.figma.com/design/qLl8KxBOOr640h18ICx0H6/Web-app?t=UrJZI5RpFBxrWajc-0 
Screenshot 2024-07-03 at 13 36 40

I have this https://github.com/jwplayer/ott-web-app/pull/565 pr that solves the problem without impacting the cleeng flow simply by using feature flag.I think we should merge this and then if we need we can refractor it later

packages/common/src/services/integrations/jwp/JWPAccountService.ts
```diff
@@ -61,7 +61,7 @@ export default class JWPAccountService extends AccountService {
       canUpdateEmail: false,
       canSupportEmptyFullName: false,
       canChangePasswordWithOldPassword: true,
-      canRenewSubscription: false,
+      canRenewSubscription: true,
```
AntonLantukh, line 64: It looks like it can't be done for JWP now.

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
```diff
@@ -276,13 +283,50 @@ export default class JWPCheckoutService extends CheckoutService {
...
+
+    const subscriptionSwitches = data?.filter((accessFee) => accessFee.item.plan_switch_enabled).map((accessFee) => this.formatOffer(accessFee)) || [];
+
+    return { responseData: subscriptionSwitches, errors: [] };
+  };
 
   getOrder = undefined;
 
-  switchSubscription = undefined;
+  switchSubscription: SwitchSubscription = async (payload) => {
```
AntonLantukh, line 296: Can we also remove `changeSubscription` method from subscription service?

packages/ui-react/src/components/Payment/Payment.tsx
```diff
@@ -159,58 +115,39 @@ const Payment = ({
     }
   }
 
-  const showChangeSubscriptionButton = (!isExternalPaymentProvider && offerSwitchesAvailable) || (!isChangingOffer && !canRenewSubscription);
+  const showChangeSubscriptionButton = !isExternalPaymentProvider && offerSwitchesAvailable;
```
AntonLantukh, line 118: Maybe we should also add `canChangeSubscription` feature flag to hide the button even though we get some offers?

AntonLantukh:
@ChristiaanScheermeijer looks great, thank you! Just some small comments. Probably some e2e tests also need changes.

packages/ui-react/src/components/Payment/Payment.tsx
ChristiaanScheermeijer: Yes, that would even be better. I think the `canRenewSubscription` flag is used were it shouldn't and this can fix that as well.

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
ChristiaanScheermeijer: Yes!

packages/common/src/services/integrations/jwp/JWPAccountService.ts
ChristiaanScheermeijer: Yes, this should be reverted and solved with a `canSwitchSubscription` flag instead.

ChristiaanScheermeijer:
@AntonLantukh thanks, this is still pretty POC'ish. Do you think it's worth to finish this PR? Perhaps we should do that when #563 is merged?

packages/ui-react/src/components/Payment/Payment.tsx
```diff
@@ -223,24 +160,15 @@ const Payment = ({
...
                     color="primary"
                     data-testid="change-subscription-button"
                   />
                 )
               )}
-              {(activeSubscription.status === 'active' || activeSubscription.status === 'active_trial') &&
-              !isGrantedSubscription &&
-              !isChangingOffer &&
-              canRenewSubscription ? (
+              {(activeSubscription.status === 'active' || activeSubscription.status === 'active_trial') && !isGrantedSubscription && canRenewSubscription ? (
```
ChristiaanScheermeijer, line 171: This condition should be checked. I think the `canRenewSubscription` should be removed for the cancel subscription button. So:

1. has subscription and is active -> Show "Cancel Subscription" button
2. has subscription and is not active and canRenewSubscription -> show "Renew Subscription" button

AntonLantukh:
> @AntonLantukh thanks, this is still pretty POC'ish. Do you think it's worth to finish this PR? Perhaps we should do that when #563 is merged?

@ChristiaanScheermeijer I find this approach cleaner than what we have when we try to build additional complexity on top of current logic. You can finish it if you have some time.

PR #566, opened 2024-07-02
Chore / Upgrade codeceptjs from 3.5.5 to 3.6.4

langemike:
Fixed a few issues when upgrading `codeceptjs` and `playwright` to their latest versions.

Original internal PR: https://github.com/Videodock/ott-web-app/pull/199


ChristiaanScheermeijer:
@dbudzins @AntonLantukh I think we need to update the required status checks because the name changed.

Since we're not using `node-version` in the matrix, we could also remove this and just configure it directly in the `actions/setup-node`

AntonLantukh:
@ChristiaanScheermeijer I can change it when we have them (different PR). For now I can only choose "18.x" or "20".

.github/workflows/web-test-e2e.yml
```diff
@@ -19,13 +19,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [18.x]
+        node-version: ['20']
```
AntonLantukh: Looks like we will need to update it when your second PR is merged.

langemike: I rebased the changes. Awaiting the outcome of the tests now..

langemike:
I set this PR to draft since there is still a failing tests... I will fix this in the upcoming sprint.

AntonLantukh (approved):

PR #569, opened 2024-07-04
Remove the InPlayer SDK usage from the Web App and use API endpoints instead

mirovladimitrovski:
## Description

Instead of using the InPlayer SDK in the Web App, we need to replace it with appropriate API endpoints for those SDK methods in the Web App. The flow, requests, and responses should remain the same

This PR solves [#OWA-80](https://jwplayer.atlassian.net/browse/OWA-80).

Note: The `inplayer.js` dependency is removed from the `react-hooks` and `ui-react` packages, but there's handful of code lines where we're still using it in the web platform and that's why we're still leaving it there for the time being. The reason is because we have websocket-based notifications implemented with `aws-iot-device-sdk` as a dependency. This websocket implementation is to be removed entirely soon enough, and that's why we're leaving it out of the web app, avoiding replacing one external dependency with another.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code


packages/common/src/services/integrations/jwp/base/JWPAPIService.ts
```diff
@@ -0,0 +1,128 @@
...
+          key && (value || value === 0)
+            ? value.toString() === '[object Object]'
+              ? Object.entries(value)
+                  .map(([subkey, subval]) => `${key}[${subkey}]=${encodeURIComponent(subval)}`)
+                  .join('&')
+              : `${key}=${typeof value === 'string' ? encodeURIComponent(value) : value}`
+            : '',
+        )
+        .filter(Boolean)
+        .join('&');
```
ChristiaanScheermeijer: Should we use https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData?

packages/common/src/services/integrations/jwp/base/JWPAPIService.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const endpoint = /^\//.test(path) ? `${this.getBaseUrl()}${path}` : path;
```
ChristiaanScheermeijer: This is more semantical (easier to understand what is happening).

```suggestion
    const endpoint = path.startsWith('/') ? `${this.getBaseUrl()}${path}` : path;
```

Although, this is more safe (i.e. when passing a path without a starting slash)

```suggestion
    const endpoint = path.startsWith('http') ? path : `${this.getBaseUrl()}${path}`;
```

packages/hooks-react/src/useProfiles.ts
```diff
@@ -1,9 +1,9 @@
-import type { ProfilesData } from '@inplayer-org/inplayer.js';
 import { useMutation, useQuery, type UseMutationOptions, type UseQueryOptions } from 'react-query';
 import { useTranslation } from 'react-i18next';
 import type { GenericFormErrors } from '@jwp/ott-common/types/form';
 import type { CommonAccountResponse } from '@jwp/ott-common/types/account';
 import type { ListProfilesResponse, ProfileDetailsPayload, ProfileFormSubmitError, ProfileFormValues, ProfilePayload } from '@jwp/ott-common/types/profiles';
+import type { ProfilesData } from '@jwp/ott-common/src/services/integrations/jwp/base/types';
```
AntonLantukh: I am not sure this is the right way to import types. Should we consider storing them in `@jwp/ott-common/types` ?

packages/ui-react/src/components/CustomRegisterField/CustomRegisterField.tsx
```diff
@@ -1,6 +1,6 @@
 import { type ChangeEventHandler, type FC, type ReactNode, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
-import type { RegisterFieldOptions } from '@inplayer-org/inplayer.js';
+import type { RegisterFieldOptions } from '@jwp/ott-common/src/services/integrations/jwp/base/types';
```
AntonLantukh: Same question about types importing

packages/hooks-react/src/useProfiles.ts
mirovladimitrovski: Agreed.

packages/ui-react/src/components/CustomRegisterField/CustomRegisterField.tsx
mirovladimitrovski: Agreed.

packages/common/src/modules/register.ts
```diff
@@ -39,6 +39,7 @@ import CleengCheckoutService from '../services/integrations/cleeng/CleengCheckou
 import CleengSubscriptionService from '../services/integrations/cleeng/CleengSubscriptionService';
 
 // InPlayer integration
+import JWPAPIService from '../services/integrations/jwp/base/JWPAPIService';
```
AntonLantukh: We could probably remove base folder here.

packages/common/src/services/JWPEntitlementService.ts
```diff
@@ -1,11 +1,22 @@
...
+  private readonly apiService;
+
+  constructor(@inject(JWPAPIService) apiService: JWPAPIService) {
+    this.apiService = apiService;
+  }
+
   getJWPMediaToken = async (configId: string = '', mediaId: string) => {
     try {
-      const { data } = await InPlayer.Asset.getSignedMediaToken(configId, mediaId);
+      const data = await this.apiService.get<SignedMediaResponse>(`v2/items/jw-media/token?app_config_id=${configId}&media_id=${mediaId}`, {
```
AntonLantukh: Could be nice to pass query params to "get" in the future as a separate argument. Looks cleaner to me.

packages/hooks-react/src/useProfiles.ts
ChristiaanScheermeijer: Types in the `@jwp/ott-common/types` directory can be used by dependants. Types in the integration folder are scoped for the integration. E.g. for API response typings that are mapped to application typings.

ChristiaanScheermeijer: In a typings refactor PR (not merged) I have separating the typings completely. I did a example in the Cleeng integration folder with a types folder.

The reason for this is to prevent the usage of integration specific types in the application logic. Services should transform models to application data.

Now we are mixing even more integration (JWP) typings. Perhaps this is a good moment to separate this for the JWP integration already?


- Create `./packages/common/src/services/integrations/jwp/types` directory
- Move inplayer DTO typings used only by services (not exposed) to this directory
- When typings are used outside the integration, we could duplicate them in the `./packages/common/typings` folder at feature level

```mermaid
graph TD
    subgraph JWP[JWP Integration]
        JWPTypes[Types]
        JWPClient[Client]
        JWPService[Service]
        
        direction RL
        JWPTypes --> JWPClient
        JWPClient --> JWPService
    end

    subgraph Cleeng[Cleeng Integration]
        direction LR
        CleengTypes[Types]
        CleengClient[Client]
        CleengService[Service]

        CleengTypes --> CleengClient
        CleengClient --> CleengService
    end

    subgraph Common
        Controller        
        Types --> Controller
    end

    Cleeng -- Service calls --> Common
    JWP -- Service calls --> Common
```

mirovladimitrovski: I would agree, but we're almost rid of this problem, at least with `Inplayer`-specific imports: There are only two instances of integration-specific types being imported and used in the web app. One of which is `RegisterFieldOptions` which I'll simply replace with an inline `Record<string, string>` and avoid importing it. The other one is `ProfilesData`, which is practically not even used anymore because we've disabled `Profiles` recently and everything related to that is dead code now, and is to be removed altogether in another upcoming task.

packages/common/src/services/JWPEntitlementService.ts
mirovladimitrovski: Done.

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (includeFullResponse) {
+      return { ...resp, data: resParsed };
+    }
+
+    return resParsed;
+  };
+
+  setup = (sandbox: boolean, siteId: string) => {
+    this.sandbox = sandbox;
+    this.siteId = siteId;
```
AntonLantukh: Do we use `site_id` here?

mirovladimitrovski: We don't, I just removed it.

packages/hooks-react/src/useProfiles.ts
AntonLantukh: Maybe we could start moving `inplayer.ts` types to `src/services/integrations/jwp/types` directory? I am not sure we should have them in `@jwp/ott-common/types`

In case anything is needed outside of the service, we could reexport them from `@jwp/ott-common/types`. @ChristiaanScheermeijer or would you prefer duplicates?

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    return { token: '', refreshToken: '', expires: 0 };
+  };
+
+  removeToken = async () => {
+    await Promise.all([this.storageService.removeItem(INPLAYER_TOKEN_KEY), this.storageService.removeItem(INPLAYER_IOT_KEY)]);
+  };
+
+  isAuthenticated = async () => {
+    const token = await this.getToken();
```
AntonLantukh:
```suggestion
    const data = await this.getToken();
```

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    const formData = new URLSearchParams();
+
+    if (body) {
+      Object.entries(body).forEach(([key, value]) => {
+        if (value || value === 0) {
+          if (typeof value === 'object') {
+            Object.entries(value as Record<string, string | number>).forEach(([innerKey, innerValue]) => {
+              formData.append(`${key}[${innerKey}]`, innerValue.toString());
```
AntonLantukh, line 89: Is it just to support custom register fields?

packages/common/types/inplayer.ts
```diff
@@ -11,3 +11,362 @@ export type InPlayerError = {
...
+  pin_required: boolean;
+  created_at: number;
+  updated_at: number;
+  credentials: {
+    access_token: string;
+    expires: number;
+  };
+};
+
+export type InplayerSubscription = {
```
AntonLantukh:
```suggestion
export type JWPSubscription = {
```

packages/common/src/services/integrations/jwp/JWPAPIService.ts
mirovladimitrovski: Probably yes. There's a nested `metadata` object that gets updated.

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -0,0 +1,142 @@
...
+            });
+          } else {
+            formData.append(key, value.toString());
+          }
+        }
+      });
+    }
+
+    const endpoint = `${path.startsWith('http') ? path : `${this.getBaseUrl()}${path}`}${
```
AntonLantukh, line 98: Do you think we also need to support `http` case here?

mirovladimitrovski: Yes, we have one use case where we use an absolute path:
https://github.com/jwplayer/ott-web-app/blob/ff97156d08f59d17c163d79a3b8ae74db8d17a00/packages/common/src/services/integrations/jwp/JWPSubscriptionService.ts#L259

packages/hooks-react/src/useProfiles.ts
mirovladimitrovski: I made duplicates, there's very few of them. But I would vote for re-exporting to avoid mismatching.

ChristiaanScheermeijer (approved):
Nice work! 🎉 Remember to squash commit or rebase this first to make the changelog a bit nicer 😄

packages/hooks-react/src/useProfiles.ts
ChristiaanScheermeijer: Re-exporting them is also fine, but @mirovladimitrovski already solved it nicely 👍

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -0,0 +1,142 @@
...
+  includeFullResponse?: boolean;
+};
+
+@injectable()
+export default class JWPAPIService {
+  private readonly storageService: StorageService;
+
+  private sandbox = true;
+
+  static isCommonError = (error: unknown): error is InPlayerError => {
```
AntonLantukh:
```suggestion
  static isCommonError = (error: unknown): error is JWPError => {
```

AntonLantukh: Same for constants. I think we can use JWP now

AntonLantukh (approved):

packages/common/src/services/integrations/jwp/JWPAPIService.ts
mirovladimitrovski: I just did that. This is the second time today this happening.

PR #573, opened 2024-07-09
Feat / support more items in header navigation (opinionated)

ChristiaanScheermeijer:
## Description

This one was not planned, but was mentioned multiple times internally. Our config has 7 menu items which currently break over multiple lines. This breaks the layout a bit, but I initially decided that it can be solved by removing menu items.

Some time later I reconsidered this and had an idea that I wanted to test. I think it works pretty well and solves the layout issues in our config.

**Before**

image

**After**

image

**How it works**

https://github.com/jwplayer/ott-web-app/assets/3996119/bc0466b5-e475-4e55-b969-1ee34d09a809

When a button receives focus, the offset position within the scroll container is calculated. If needed, the navigation is scrolled to the target element with an offset.

Preview: https://ottwebapp--pr573-feat-header-navigati-c9ryg6am.web.app/?app-config=eoncwnho

### DIFF
--- a/packages/ui-react/src/components/Header/Header.module.scss
+++ b/packages/ui-react/src/components/Header/Header.module.scss
@@ -10,7 +10,7 @@
 
 .header {
   height: variables.$header-height;
-  padding: 10px calc(#{variables.$base-spacing} * 2);
+  padding: 0 calc(#{variables.$base-spacing} * 2);
   color: var(--header-contrast-color, variables.$white);
   background: var(--header-background, transparent);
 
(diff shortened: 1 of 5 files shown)


royschut (approved):
I find it convenient and intuitive. Curious what others think!

AntonLantukh:
@ChristiaanScheermeijer looks great, thank you! I also like spiegel's implementation where simple arrows were added:
https://www.spiegel.de/

AntonLantukh (approved):

langemike (approved):
Nice solution! Have not seen a solution like this before, but I like it.

ChristiaanScheermeijer:
> @ChristiaanScheermeijer looks great, thank you! I also like spiegel's implementation where simple arrows were added: https://www.spiegel.de/

Yes, that looks better indeed. Let's wait for feedback first, because I think we should advise to put less items in the menu.

PR #575, opened 2024-07-10
Remove `Profiles` and all dead code associated with it

mirovladimitrovski:
## Description

We have recently hardcoded the `Profiles` section and all of its logic as permanently disabled, which practically makes it dead code. In the meantime we’ve been making such general changes in the app that somewhat forced us to update even these dead code parts despite knowing we no longer need them, just so that TypeScript or ESLint wouldn’t complain and the app can be built. Such example was removing the InPlayer SDK (the `@inplayer-org/inplayer.js` dependency) and replacing it with direct API requests using `fetch()`. There will be other changes of this kind and in order to avoid unnecessary work it’d be best for us to remove all the code related to Profiles.

This PR solves [#OWA-82](https://jwplayer.atlassian.net/browse/OWA-82).

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code


packages/ui-react/src/pages/User/User.tsx
```diff
@@ -87,29 +72,17 @@ const User = (): JSX.Element => {
         <nav className={styles.leftColumn}>
           <div className={styles.panel}>
             <ul>
-              {accessModel === 'SVOD' && profilesEnabled && profileAndFavoritesPage && (
```
AntonLantukh, line 90: Should we keep `accessModel` here?

AntonLantukh: Looks like it can be removed

AntonLantukh (approved):

AntonLantukh:
I think we also want to highlight in the changelog that profiles were removed. "chore" type is probably ignored by default

packages/ui-react/src/pages/User/User.tsx
mirovladimitrovski: It is removed, you're looking at a deleted line. 😃

AntonLantukh: Yeah, that is what I meant. I think we are good here

mirovladimitrovski: Ah, got it. Yeah, this entire block is removed, not just the condition we are looking at. As you already figured out.

PR #576, opened 2024-07-11
Feat / Add backClick event support for player

langemike:
`backClick` is a [documented](https://github.com/jwplayer/jwplayer-tizen-app?tab=readme-ov-file#creating-your-own-player-screen) event of JWPlayer which is only used on the Tizen platform. It was not implemented yet in the OTT platform.

The conditional `playerRef.current.remove();` is needed because otherwise the player tries to remove itself twice (also in the internal JW event `backClick` [handler](https://github.com/jwplayer/jwplayer/blob/30353cd1e1f3017a96ef2854ef758fb4f479cd7a/src/js/view/controls/tizen/tizen-controls.ts#L218)) causing an error.

Ticket: https://videodock.atlassian.net/browse/OTT-1613

### DIFF
--- a/packages/ui-react/src/components/Player/Player.tsx
+++ b/packages/ui-react/src/components/Player/Player.tsx
@@ -244,9 +244,12 @@ const Player: React.FC<Props> = ({
     return () => {
       if (playerRef.current) {
         // Detaching events before component unmount
-        if (!backClickRef.current) {
-          playerRef.current.remove();
+        detachEvents();
+        if (backClickRef.current) {
+          backClickRef.current = false;
+          return;
         }
+        playerRef.current.remove();
       }
     };
   }, [detachEvents, backClickRef]);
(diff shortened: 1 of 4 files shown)


packages/ui-react/src/components/Player/Player.tsx
```diff
@@ -235,11 +244,12 @@ const Player: React.FC<Props> = ({
     return () => {
       if (playerRef.current) {
         // Detaching events before component unmount
-        detachEvents();
-        playerRef.current.remove();
+        if (!backClickRef.current) {
+          playerRef.current.remove();
+        }
```
royschut: Have you removed `detachEvents` intentionally?

We should also reset the backClickRef:
```suggestion
        if (backClickRef.current) {
          backClickRef.current = false;
          return;
        }
        playerRef.current.remove();
```

langemike: Good idea. Thanks. I accidentally removed `detachEvents` 😅 . I restored it.

ChristiaanScheermeijer (approved):

AntonLantukh (approved):

PR #577, opened 2024-07-12
Feat / add cancel functions for debounce and throttle utils

ChristiaanScheermeijer:
## Description

Simple change to allow cancelation of a debounced/throttled function. This is needed to prevent the focus call in the LayoutGrid when the focus moves out.

### DIFF
--- a/packages/common/src/utils/common.ts
+++ b/packages/common/src/utils/common.ts
@@ -1,17 +1,21 @@
 import type { Playlist, PlaylistItem } from '../../types/playlist';
 
 export function debounce<T extends (...args: any[]) => void>(callback: T, wait = 200) {
-  let timeout: NodeJS.Timeout | null;
-  return (...args: unknown[]) => {
+  let timeout: NodeJS.Timeout | undefined;
+  function debounced(...args: unknown[]) {
     if (timeout) clearTimeout(timeout);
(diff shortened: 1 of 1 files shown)


royschut (approved):

langemike (approved):
Handy addition

packages/common/src/utils/common.ts
```diff
@@ -1,17 +1,21 @@
...
-  let timeout: NodeJS.Timeout | null;
-  return (...args: unknown[]) => {
+  let timeout: NodeJS.Timeout | undefined;
+  function debounced(...args: unknown[]) {
     if (timeout) clearTimeout(timeout);
     timeout = setTimeout(() => callback(...args), wait);
-  };
+  }
+
+  debounced.cancel = () => clearTimeout(timeout);
```
AntonLantukh, line 10: Any reference implementation you plan to add?

ChristiaanScheermeijer: Yes, in LayoutGrid.tsx:

```tsx
    // cancel the scrollIntoView action when the focus leaves the layout grid
    if (isLeaving) scrollIntoViewThrottled.cancel();
```

and:

```tsx
useEffect(() => {
  return () => scrollIntoViewThrottled.cancel();
}, []);
```

AntonLantukh (approved):

PR #580, opened 2024-07-16
Feat / implement tile-slider dependency

ChristiaanScheermeijer:
## Description

We've been working on this quite some time. Unfortunately, due to a few unforeseen issues this PR was delayed a few times. 

This PR implements the tile-slider dependency. The `tile-slider` package is similar to the TileDock component, but is maintained separately and contains fixes and features that are not implemented in TileDock.

The main reason we want to use `tile-slider` is for improved touch support. Currently, only swiping without visual feedback is supported. Using `tile-slider`, the slider behaves similar to a scrollable element but has the same performance optimisations (flat list).

We postponed this PR because of a performance issue on mobile phones. This was caused by the tiles rendered in the slider loading the image and animating the opacity when out of view. This happens multiple times when swiping the slider and causes hiccups in the slide animation.

To prevent this, @royschut has done optimisations and we disabled the card opacity on mobile.

**Preview**

https://github.com/user-attachments/assets/8423b0b2-8f1f-4ac3-8c05-c214367248a6


royschut (approved):

langemike (approved):
I already reviewed it internally. I approve.

packages/ui-react/src/components/Card/Card.module.scss
```diff
@@ -103,12 +103,6 @@
     left: 0;
     width: 100%;
     height: 100%;
-    opacity: 0;
```
AntonLantukh, line 106: Does it also disable opacity for desktop devices?

packages/ui-react/src/components/Shelf/Shelf.tsx
```diff
@@ -10,8 +11,8 @@ import ChevronLeft from '@jwp/ott-theme/assets/icons/chevron_left.svg?react';
 import ChevronRight from '@jwp/ott-theme/assets/icons/chevron_right.svg?react';
 import useBreakpoint, { Breakpoint, type Breakpoints } from '@jwp/ott-ui-react/src/hooks/useBreakpoint';
 import type { PosterAspectRatio } from '@jwp/ott-common/src/utils/collection';
+import '@videodock/tile-slider/lib/style.css';
```
AntonLantukh, line 14: Hm, why can't we use styles from `TileSlider` itself?

packages/ui-react/src/components/Image/Image.tsx
```diff
@@ -16,14 +16,56 @@ const setWidth = (url: string, width: number) => {
   return createURL(url, { width });
 };
 
+const cache = new Map();
+
+const resolveImageURL = async (imgUrl: string, width: number) => {
```
AntonLantukh, line 21: Is it something we need because of the thumbnail fallback returning 1920 image?

packages/ui-react/src/components/Card/Card.module.scss
ChristiaanScheermeijer: This is moved to the `Shelf.module.scss` file because we want to limit card re-renders while sliding

packages/ui-react/src/components/Shelf/Shelf.tsx
ChristiaanScheermeijer: This is using the styles from TileSlider, right? 😅

packages/ui-react/src/components/Image/Image.tsx
ChristiaanScheermeijer: Yes, that's correct. But also because we found that the redirect causes the browser caching mechanism to fail. Not sure if this is caused by invalid/missing cache headers, but this fix resolves the redirect URL once and caches it for the next time.

AntonLantukh: We have open PRs to fix both now (adding Cache-Control header + fallback image size fix). Hope it will be delivered soon.

packages/ui-react/src/components/Shelf/Shelf.tsx
AntonLantukh: Right, I mean, can we import them in `TileSlider.tsx` and get them from component and not separately?
Or is your idea to redefine them for some use cases?

packages/ui-react/src/components/Image/Image.tsx
ChristiaanScheermeijer: That would be awesome! Do you have an ETA? 

I'm asking because currently the homepage images are loading really slow in different regions (Thailand for example). This fix drastically improves that, so we could decide to merge this and remove it when the CDN PRs are landed?

packages/ui-react/src/components/Shelf/Shelf.tsx
ChristiaanScheermeijer: It has to be separate because it's a compiled CSS file. We don't include it by default in TileSlider, because users can decide to exclude the default styling and do it from scratch (full control).

ChristiaanScheermeijer:
Deploy preview: https://pr-200.ott-previews.videodock.com/?app-config=eoncwnho

packages/ui-react/src/components/Image/Image.tsx
AntonLantukh: Yeah, we can also merge these changes before. I hope this / next week.

AntonLantukh (approved):

PR #581, opened 2024-07-18
Fix / Wait for isGeoBlocked status and cache it’s value

langemike:
Wait for the `isGeoBlocked` status to be known before mounting/unmounting the player. Otherwise you can run into player errors, because the player is destructed before it has time to be fully initialized caused by the conditional component rendering in `PlayerContainer`.

We also utilize TanStack Query to cache it's value.

These errors happen in the console while the player is not present in the DOM. It does not interfere with the user.
We experienced this issue on Tizen while pressing keys (LRUD) when the PlayerError is shown.

### DIFF
--- a/packages/hooks-react/src/useProtectedMedia.ts
+++ b/packages/hooks-react/src/useProtectedMedia.ts
@@ -26,6 +26,6 @@ export default function useProtectedMedia(item: PlaylistItem) {
   return {
     ...contentProtectionQuery,
     isGeoBlocked,
-    isLoading: contentProtectionQuery.isLoading && isLoading,
+    isLoading: contentProtectionQuery.isLoading || isLoading,
   };
 }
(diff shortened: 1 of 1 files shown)


ChristiaanScheermeijer (approved):
LGTM!

packages/hooks-react/src/useProtectedMedia.ts
```diff
@@ -7,21 +7,25 @@ import useContentProtection from './useContentProtection';
...
+    },
+    {
+      enabled: contentProtectionQuery.isFetched,
+    },
+  );
 
   return {
     ...contentProtectionQuery,
     isGeoBlocked,
+    isLoading: contentProtectionQuery.isLoading && isLoading,
```
royschut: Shouldn't this be `||`?
```suggestion
    isLoading: contentProtectionQuery.isLoading || isLoading,
```

royschut (approved):

packages/hooks-react/src/useProtectedMedia.ts
langemike: You are right. I changed it

PR #582, opened 2024-07-22
[IDM-161] - feat: IAM Middleware

kmitrovv:
## Description

This PR sets up the initial project structure for the Access Bridge service. 
A service that facilitates seamless communication between SIMS (Subscriber Identity Management System) and Access Control services. It provides endpoints to generate and refresh access passports for authenticated viewers, ensuring secure and efficient access management.

**Create Initial Project Structure**

- Set up the project structure following the service and controller patterns.
- Installed necessary dependencies and added the required configurations.

**Generate Passport**

- Implemented the endpoint to generate a new passport.
- Created the corresponding controller and service logic.

**Refresh Passport**

- Implemented the endpoint to refresh an existing passport, generating a new expiry and refresh token.
- Created the corresponding controller and service logic.

**Add Tests**

- Added unit tests for the small functionalities as well as util helper functions.
- Added integration tests for the generate passport and refresh passport functionalities to ensure proper behavior and stability.

Design document: [AC - Sims Integration](https://www.notion.so/jwplayer/AC-Sims-Integration-1a61b486d8d74567971358901f41a34b)

Related Jira cards: 

- [Create initial project structure](https://jwplayer.atlassian.net/browse/IDM-166)
- [Generate passport](https://jwplayer.atlassian.net/browse/IDM-170)
- [Refresh passport](https://jwplayer.atlassian.net/browse/IDM-210)
- [Tests](https://jwplayer.atlassian.net/browse/IDM-212)

**Note**:
There will be follow up PRs related to this one, that will include the Stripe logic needed for fetching Stripe products metadata, as well as payment initiation.

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


ChristiaanScheermeijer:
Nice PR @kiremitrov123! I haven't looked much into what is needed, but I assume this middleware is needed for the frontend. Just in case, are you really sure that we can't fetch the offers from the public Stripe APIs?

Perhaps we can discuss the following here:

- Should we use a lightweight HTTP framework (https://fastify.dev/)?
- Should we consider this a `platform` or create a new packages folder called `api` or `backend`?
- Should we use Inversify to make it easier to replace the Stripe services with a different integration?

I will review the code more thoroughly soon 😄

kmitrovv:
> Nice PR @kiremitrov123! I haven't looked much into what is needed, but I assume this middleware is needed for the frontend. Just in case, are you really sure that we can't fetch the offers from the public Stripe APIs?
> 
> Perhaps we can discuss the following here:
> 
> * Should we use a lightweight HTTP framework (https://fastify.dev/)?
> * Should we consider this a `platform` or create a new packages folder called `api` or `backend`?
> * Should we use Inversify to make it easier to replace the Stripe services with a different integration?
> 
> I will review the code more thoroughly soon 😄

@ChristiaanScheermeijer thanks for the input!

Yes, right! This middleware (access-bridge) will be used by the frontend to connect with both the SIMS and Access Control systems. It will generate/refresh passports, list Stripe products, and handle Stripe checkout sessions.

Key Points:

1. **Stripe Products and Checkout Sessions**: Accessing Stripe products requires a secret key, as detailed [here](https://docs.stripe.com/api/products/list). The secret key is also necessary for Stripe checkout sessions, which will be addressed in a subsequent PR.
2. **Framework Usage:** While I initially aimed to keep it simple with minimal libraries, I’m open to using a framework if the team agrees.
3. **Platform vs. Package**: After discussing with @dbudzins, it seems more logical to treat this as a platform, given that it will need to be deployed on Firebase alongside the web (as a POC), otherwise, it will serve as a starting point for the customers, where they can either just replace keys and host it somewhere on their side or add additional logic as they require. Feels more like a platform than a package I guess 😄 .
4. **Integration and Replaceability**: The goal is not to make Stripe replaceable but to add more integrations as needed. However, if it makes sense, we can consider making it more flexible. 😄

platforms/access-bridge/README.md
```diff
@@ -0,0 +1,66 @@
...
+Create a `.env.local` file in the root of this project and add the following variables:
+
+- API_SECRET=customer_v1_secret
+- BIND_ADDR=localhost
+- BIND_PORT=3000
+- ACCESS_CONTROL_CLIENT=https://cdn-dev.jwplayer.com
+- PLANS_CLIENT=https://daily-sims.jwplayer.com
+
+Make sure to replace the placeholder values (e.g., customer_v1_secret) with the actual values from your JW Dashboard.  
+You can also copy and paste the contents of `.env.example` into `.env.local` and just ajust the API_SECRET.
```
budzworthy:
```suggestion
You can also copy and paste the contents of `.env.example` into `.env.local` and just adjust the API_SECRET.
```

platforms/access-bridge/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@jwp/access-bridge",
+  "version": "1.0.0",
+  "main": "build/main.js",
+  "private": true,
+  "author": "JW Player",
+  "type": "module",
+  "scripts": {
+    "build": "tsc && [ -f .env ] && cp .env ./build/.env || true && [ -f .env.local ] && cp .env.local ./build/.env.local || true",
```
budzworthy: Did you look at using vite to do the node build? It would probably be useful to align with the rest of the web app code.

platforms/access-bridge/src/app-config.ts
```diff
@@ -0,0 +1,46 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+import * as dotenv from 'dotenv';
+
+// Load environment variables from .env files
+function loadEnvFiles() {
```
budzworthy: I'd suggest we use vite to inject the (non-secret) env variables directly into code and a secret manager to set env vars for the secrets: https://cloud.google.com/run/docs/configuring/services/secrets

platforms/access-bridge/src/controllers/access-controller.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Controller class responsible for handling access-related services.
+ */
+export class AccessController {
+  private accountService: AccountService;
+  private accessService: AccessService;
+  private plansService: PlansService;
+
+  constructor() {
```
budzworthy, line 17: Did you look at using inversify to inject these services?

platforms/access-bridge/src/controllers/access-controller.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.plansService = new PlansService();
+  }
+
+  /**
+   * Service handler for generating passport access tokens.
+   * @param req The HTTP request object.
+   * @param res The HTTP response object.
+   * @param params The request parameters containing site_id.
+   */
+  generatePassport = async (req: IncomingMessage, res: ServerResponse, params: { [key: string]: string }) => {
```
budzworthy: Is there a reason that params isn't typed?

platforms/access-bridge/src/controllers/access-controller.ts
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      if (!isValidSiteId(params.site_id)) {
+        sendErrors(res, new ParameterInvalidError({ parameterName: 'site_id' }));
+        return;
+      }
+
+      // unauthorized is default for viewers without an authorization token
+      let viewer: Viewer = { id: 'unauthorized', email: '' };
```
budzworthy: Is this a standard from the Sims side or should we get a consensus on what id/email to use in this case?

platforms/access-bridge/src/errors.ts
```diff
@@ -0,0 +1,205 @@
...
+ * Send one or more errors.
+ *
+ * @param res - The response object to send the error to. This ends the response.
+ * @param error - The error to send.
+ * @param errors - Any other errors to send. Their status code must match error's.
+ */
+export function sendErrors(res: ServerResponse, error: AccessBridgeError, ...errors: AccessBridgeError[]) {
+  // All errors must share the same error code
+  const statusCode = error.statusCode;
+  if (errors.find((e) => e.statusCode !== statusCode)) {
```
budzworthy: It looks like you're not using the `...errors` param in any of the calls of this method, so I would suggest removing it. Otherwise, you should probably refactor it to prevent this condition from happening. Otherwise you have what is essentially a compile time error causing a runtime error.

platforms/access-bridge/src/server.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.httpServer = null;
+    this.address = address;
+    this.port = port;
+    this.endpointsHandler = endpointsHandler;
+  }
+
+  public async listen(): Promise<number> {
+    return new Promise((resolve, reject) => {
+      const server = createServer(async (req: IncomingMessage, res: ServerResponse) => {
+        res.setHeader('Content-Type', 'application/json');
```
darkoatanasovski:
```suggestion
        res.setHeader('Content-Type', 'application/json');
        res.setHeader('Access-Control-Allow-Headers', '*'); //or we can use env variable here
```

platforms/access-bridge/src/controllers/access-controller.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+  };
+
+  /**
+   * Service handler for refreshing passport access tokens.
+   * @param req The HTTP request object.
+   * @param res The HTTP response object.
+   * @param params The request parameters containing site_id.
+   */
+  refreshPassport = async (req: IncomingMessage, res: ServerResponse, params: { [key: string]: string }) => {
```
budzworthy: Do we need refresh passport to go through the middleware or can that call go directly to AC?

platforms/access-bridge/src/errors.ts
```diff
@@ -0,0 +1,205 @@
+import { ServerResponse } from 'node:http';
+
+import { JWError, JWErrorResponse } from '@jwp/ott-common/types/errors.js';
+
+import { RequestMethod } from './http.js';
+
+/**
+ * List of possible error codes based on RFC-JW03-Delivery API
+ * https://www.notion.so/jwplayer/RFC-JW03-Delivery-API-fc35c5f5d300445eba4004de970d06b4
```
budzworthy: Can you reference the public docs instead of notion? https://docs.jwplayer.com/platform/reference/error-codes

This is in the public repo, so partners and customers won't have access to our notion pages.

platforms/access-bridge/src/errors.ts
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * This function replaces placeholders in the error description with the provided `value` and `reason`.
+ *
+ * @param {ErrorCode} code - The error code for which the description is needed.
+ * @param {string} [value] - An optional value to replace the `{value}` placeholder in the description.
+ * @param {string} [reason] - An optional reason to replace the `{reason}` placeholder in the description.
+ * @returns {string} The description corresponding to the error code, with the `{value}` and `{reason}`
+ * placeholders replaced by the provided arguments if applicable.
+ */
+export function getErrorCodeDescription(code: ErrorCode, value?: string, reason?: string): string {
```
budzworthy: There seems to be a lot of boilerplate repeat of getErrorCodeDescription. Can you put it in the abstract class?

platforms/access-bridge/src/errors.ts
```diff
@@ -0,0 +1,205 @@
...
+
+import { JWError, JWErrorResponse } from '@jwp/ott-common/types/errors.js';
+
+import { RequestMethod } from './http.js';
+
+/**
+ * List of possible error codes based on RFC-JW03-Delivery API
+ * https://www.notion.so/jwplayer/RFC-JW03-Delivery-API-fc35c5f5d300445eba4004de970d06b4
+ */
+export enum ErrorCode {
```
budzworthy: I wonder if these 3 type definitions (ErrorCode, ErrorDescription, and ErrorStatusCode) in addition to the const error instances at the bottom could all be combined into 1 typed object like you did with the Endpoint Handlers. It fees like the 4 pieces can safely be bundled together to keep the data here more compact and keep the codes, types, and descriptions all together that belong together.

platforms/access-bridge/src/endpoints.ts
```diff
@@ -0,0 +1,20 @@
+import { IncomingMessage, ServerResponse } from 'node:http';
+
+import { AccessController } from './controllers/access-controller.js';
+
+export type EndpointHandler = {
```
budzworthy: This is an elegant, simple implementation, but did you look at using a framework like express or nestjs? Since this is a demo repo, using a framework would help keep the code to just what's interesting for demo purposes. While this is a nice implementation, it adds complexity for someone trying to go through and understand what the JWP specific integration code should look like.

platforms/access-bridge/src/http.ts
```diff
@@ -0,0 +1,97 @@
...
+    method: 'PUT',
+    headers,
+    body: JSON.stringify(body),
+  });
+
+  return await handleResponse(response);
+}
+
+// Handles the response from a fetch request, throwing an error if the response is not ok.
+async function handleResponse<T>(response: Response): Promise<T> {
```
dbudzins, line 81: Is this the same as `getDataOrThrow` from the web app code? Could we move that to a common project to share it and fully leverage the monorepo?

platforms/access-bridge/src/http.ts
```diff
@@ -0,0 +1,97 @@
+// Allowed backend request methods.
+export const ALLOWED_REQUEST_METHODS = ['GET', 'PUT', 'POST'] as const;
```
budzworthy: This looks very similar to JWPAPIService. Is it possible to move that to a common shared project so it can be used in both places and fully leverage the monorepo setup?

platforms/access-bridge/src/services/access-service.ts
```diff
@@ -0,0 +1,118 @@
+import jwt from 'jsonwebtoken';
+import {
+  AccessTokensResponse,
```
budzworthy: Is there a reason to define these types external to this file? It seems they are only used in this code and could be better encapsulated by putting them here.

platforms/access-bridge/src/services/access-service.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import { ACCESS_CONTROL_CLIENT, API_SECRET } from '../app-config.js';
+import { BadRequestError, ForbiddenError, NotFoundError, ParameterInvalidError, isJWError } from '../errors.js';
+import { put } from '../http.js';
+
+/**
+ * AccessService handles interactions with the access management APIs.
+ * It provides methods to generate and refresh access tokens.
+ */
+export class AccessService {
```
budzworthy: Did you consider calling this PassportService?

platforms/access-bridge/src/services/access-service.ts
```diff
@@ -0,0 +1,118 @@
...
+  async refreshAccessTokens({ siteId, refreshToken }: RefreshAccessTokensParams): Promise<AccessTokensResponse> {
+    try {
+      const url = await this.generateSignedUrl(`/v2/sites/${siteId}/access/refresh`);
+      const payload = {
+        refresh_token: refreshToken,
+      };
+
+      return await put<AccessTokensResponse, typeof payload>(url, payload);
+    } catch (e) {
+      if (isJWError(e)) {
```
budzworthy: Is there a reason to define these switch cases for each method? This seems like something that can be universal.

platforms/access-bridge/src/services/access-service.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  // URL signer - needed for validating requests on Delivery Gateway
+  // More about this: https://docs.jwplayer.com/platform/reference/protect-your-content-with-signed-urls
+  async generateSignedUrl(path: string, host: string = this.accessControlClient) {
+    const now = new Date();
+    const token = jwt.sign(
+      {
+        exp: Math.ceil((now.getTime() + 3600) / 300) * 300,
```
budzworthy: Would be nice to comment what this math represents.

platforms/access-bridge/src/services/access-service.ts
```diff
@@ -0,0 +1,118 @@
...
+ * AccessService handles interactions with the access management APIs.
+ * It provides methods to generate and refresh access tokens.
+ */
+export class AccessService {
+  private apiSecret: string;
+  private accessControlClient: string;
+
+  constructor() {
+    this.apiSecret = API_SECRET;
+    this.accessControlClient = ACCESS_CONTROL_CLIENT;
```
budzworthy: Suggest naming this ACCESS_CONTROL_HOST

platforms/access-bridge/src/services/access-service.ts
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * AccessService handles interactions with the access management APIs.
+ * It provides methods to generate and refresh access tokens.
+ */
+export class AccessService {
+  private apiSecret: string;
+  private accessControlClient: string;
+
+  constructor() {
+    this.apiSecret = API_SECRET;
```
budzworthy: Is there any value to putting the global values into private fields? It seems like it would be easier to debug and inspect code if we just used the globals directly where needed.

platforms/access-bridge/src/global.d.ts
```diff
@@ -0,0 +1,13 @@
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv {
+      API_SECRET: string;
+      STRIPE_SECRET: string;
+      BIND_ADDR: string;
+      BIND_PORT: string;
+      ACCESS_CONTROL_CLIENT: string;
+      SIMS_CLIENT: string;
```
budzworthy: These seem like hosts, not clients, right?
Naming tweak to be more meaningful than our internal team names.
```suggestion
      PASSPORT_API_HOST: string;
      SUBSCRIPTION_API_HOST: string;
```

platforms/access-bridge/src/http.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// Type to describe allowed request methods.
+export type RequestMethod = (typeof ALLOWED_REQUEST_METHODS)[number];
```
dbudzins, line 5: I thought I added a comment about this already, but don't see it. Isn't this very similar to JwpApiService? Can we reuse it as a shared monorepo project?

platforms/access-bridge/src/services/account-service.ts
budzworthy, line 1: Let's call this "identity service" to keep with our SIMS nomenclature

platforms/access-bridge/src/services/plans-service.ts
budzworthy, line 1: Let's call this "subscription service" to keep with our SIMS nomenclature

platforms/access-bridge/src/services/plans-service.ts
```diff
@@ -0,0 +1,93 @@
...
+  private simsClient: string;
+
+  constructor() {
+    this.simsClient = SIMS_CLIENT;
+  }
+
+  /**
+   * Retrieves access control plans based on the specified endpoint type.
+   *
+   * Depending on the `endpointType` parameter, this function either:
```
budzworthy: I would probably split these 2 endpoints into 2 separate methods. Does the 'plans' endpoint need to be in the middleware or could it be called from the FE directly?

platforms/access-bridge/src/services/plans-service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+          if (!id || !exp) {
+            return null;
+          }
+
+          return {
+            id,
+            exp,
+            external_providers: {
+              stripe: plan.metadata?.external_providers?.stripe,
```
budzworthy: When we're listing the plans to buy, we should only show the stripe ones in the web app

platforms/access-bridge/src/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+import { BadRequestError } from './errors.js';
+
+export function isValidSiteId(siteId: string): boolean {
+  // Regular expression to match exactly 8 alphanumeric characters
+  const alphanumericRegex = /^[a-zA-Z0-9]{8}$/;
+  return alphanumericRegex.test(siteId);
+}
+
+// parse request body provided in an API call
+export const parseJsonBody = <T>(req: IncomingMessage): Promise<T> => {
```
budzworthy: If you use express framework or other libs, you can avoid this extra implementation logic

platforms/access-bridge/test/integration/access.ts
```diff
@@ -0,0 +1,179 @@
...
+          const responseBody = JSON.parse(body);
+          assert.strictEqual(responseBody.passport, ACCESS_TOKENS.PASSPORT.VALID);
+          assert.strictEqual(responseBody.refresh_token, ACCESS_TOKENS.REFRESH_TOKEN.VALID);
+          done();
+        });
+      })
+      .end();
+  });
+
+  await test('should generate passport access tokens with valid authorization', (t, done) => {
```
budzworthy: Can this test and the 1 above be 2 test cases with different params but the same logic?

platforms/access-bridge/src/controllers/access-controller.ts
budzworthy: Technically isn't this "unauthenticated" not "unauthorized"?

platforms/access-bridge/test/unit/errors.ts
```diff
@@ -0,0 +1,79 @@
...
+  InternalError,
+  MethodNotAllowedError,
+  NotFoundError,
+  ParameterInvalidError,
+  ParameterMissingError,
+  UnauthorizedError,
+} from '../../src/errors.js';
+import { RequestMethod } from '../../src/http.js';
+
+describe('Error Classes', () => {
```
budzworthy: I'm not sure how valuable these tests are. I would prefer to test the error codes through a response path that throws the expected type of error. Otherwise, it's a bunch of boilerplate to test a bunch of boilerplate.

platforms/access-bridge/test/integration/access.ts
budzworthy: https://mochajs.org/#dynamically-generating-tests

platforms/access-bridge/test/unit/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+import { IncomingMessage } from 'node:http';
+import { Socket } from 'node:net';
+import { describe, test } from 'node:test';
+
+import { SITE_ID } from '../fixtures.js';
+import { isValidSiteId, parseJsonBody } from '../../src/utils.js';
+import { AccessBridgeError, ErrorCode } from '../../src/errors.js';
+
+describe('isValidSiteId', () => {
+  test('should return true for valid site IDs', () => {
```
budzworthy: Again, I would parameterize these: https://mochajs.org/#dynamically-generating-tests

budzworthy (commented):
There's a lot of work here that you implemented in a short stretch! Some regular PR comments here and there. Otherwise, the main suggestion is to see what we can strip out by using some well known frameworks or external libs so that the code here is just the interesting stuff for writing a middleware. Lastly, I think there are some small things that we might be able to share between the BE and FE using the monorepo. Nice work though! We're almost there!

platforms/access-bridge/src/http.ts
darkoatanasovski: do we need this? :)

darkoatanasovski:
```suggestion
export const ALLOWED_REQUEST_METHODS = ['OPTIONS','GET', 'PUT', 'POST'] as const;
```

platforms/access-bridge/src/controllers/access-controller.ts
kmitrovv: Introducing a dependency injection framework seems like a premature optimization, given that the project's requirements and scope are still evolving. By keeping things simple for now, we can ensure that the codebase remains easy to understand and adapt as needed. We can always revisit and incorporate such optimizations when we have a clearer picture of the project's direction and specific needs.

platforms/access-bridge/src/controllers/access-controller.ts
kmitrovv: updated with express.

platforms/access-bridge/src/controllers/access-controller.ts
kmitrovv: This is something we should introduce for free and unauthenticated viewers. You are right, I would rename it to unauthenticated, but probably something we should bring up within our checkup.

platforms/access-bridge/src/errors.ts
kmitrovv: updated.

platforms/access-bridge/src/server.ts
kmitrovv: updated using express, no longer needed.

platforms/access-bridge/src/controllers/access-controller.ts
kmitrovv: you are right, removing refreshing passport, as we can directly call it from the passport service itself.

platforms/access-bridge/src/errors.ts
kmitrovv: good suggestion! grouped them into one object, seems lot clearer now.

platforms/access-bridge/src/endpoints.ts
kmitrovv: I did introduce express, but still not seeing any differences in addressing your concern. If you have any other suggestions how we should define routes, feel free to recommend. I think it's very nice to have them separately defined like this.

platforms/access-bridge/src/http.ts
kmitrovv: deleted as we are using express now.

kmitrovv: @dbudzins , reusing the JWPAPIService as it is now, seems like an overkill. It is coupled with StorageService which we don't need in this case, so leaving that refactor for future improvements.

platforms/access-bridge/src/services/access-service.ts
kmitrovv: Some of them are required on the FE part as well, e.g. the PassportResponse, but stripped off the non-required ones, so we are good.

platforms/access-bridge/src/services/access-service.ts
kmitrovv: Makes sense.

platforms/access-bridge/src/services/access-service.ts
kmitrovv: Good catch, defined a global handler for this that uses the entire error definition object.

platforms/access-bridge/src/services/access-service.ts
kmitrovv: you are right, approved this approach.

platforms/access-bridge/src/global.d.ts
kmitrovv: The Subscription doesn't really make anymore sense, since we are using external providers and not internal anymore. So, I did name it SIMS_API_HOST.

platforms/access-bridge/src/services/plans-service.ts
kmitrovv: I wouldn't name it lake that since it does not do anything else than handling plans. Subscription and payment related stuff is a Stripe job now.

platforms/access-bridge/src/services/plans-service.ts
kmitrovv: It will be bunch of boilerplate as they share the same interface. These are both good to have them here in the middleware since one is used to generate passport (entitlements) and the other to match the stripe products. 
Only the entitlements one will be probably exposed on the FE side (for locks check and stuff), but still, not quite secure to send the plans from the frontend as they can be manipulated.

platforms/access-bridge/src/services/plans-service.ts
kmitrovv: this is internal, so we can later match the stripe products. the other ones are not really used for now, but for future inapp implementations they will be.

platforms/access-bridge/src/controllers/access-controller.ts
kmitrovv: @dbudzins looking back at why I did this to be unauthorized.. Actually the response for invalid token or expired, by our standards is used `unauthorized`.. So, I think it makes more sense to be compatible or any other suggestion?

platforms/access-bridge/src/services/plans-service.ts
kmitrovv: On the other hand, I did split them and made the services to be straight forward without any logic in them. All the logic goes in the controllers now. Thanks!

platforms/access-bridge/src/errors.ts
```diff
@@ -1,196 +1,132 @@
...
-  protected abstract get defaultDescription(): string;
+interface ParameterInvalidContext extends BaseContext {
+  parameterName: string;
+  reason?: string;
+}
 
-  get description(): string {
-    return this.descriptionOverride || this.defaultDescription;
+// Unified Error Definitions with details and creation functions
+export const ErrorDefinitions = {
```
budzworthy, line 23: Did you consider making these all classes that inherit from `AccessBirdgeError`? The `create` prop looks like it could be a constructor.

platforms/access-bridge/src/server.ts
```diff
@@ -1,123 +1,106 @@
...
+  private notFoundErrorHandler = (req: Request, res: Response, next: NextFunction) => {
+    sendErrors(res, ErrorDefinitions.NotFoundError.create());
+    return;
+  };
+
+  /**
+   * Global error handler middleware for the server.
+   */
+  private globalErrorHandler = (err: unknown, req: Request, res: Response, next: NextFunction) => {
+    if (err instanceof AccessBridgeError) {
```
budzworthy: Can you put the 'isJwError' logic here instead of having it in each try-catch?

kmitrovv: But then the error instance will not be AccessBridgeError and we will not be able to sendErrors here as they will not be handled. Using isJwError is possible but we still need to handle them before sending. 
Also why we wanna lose the service and controller error level handling. This suggestion also means just passing the errors from service level to controller and from controller to here.

platforms/access-bridge/src/services/passport-service.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // URL signer - needed for validating requests on Delivery Gateway
+  // More about this: https://docs.jwplayer.com/platform/reference/protect-your-content-with-signed-urls
+  async generateSignedUrl(path: string, host: string = ACCESS_CONTROL_API_HOST): Promise<string> {
+    try {
+      const now = new Date();
+      const token = jwt.sign(
+        {
+          // The expiration timestamp is calculated by taking the current time,
+          // adding 3600 milliseconds (which is 1 second)
```
budzworthy: I think the logic is ok here, but the comment is wrong. 3600ms is 3.6 seconds. it should be 3600 seconds = 1 hour, with 300 seconds = 5 minute boundaries, right?

platforms/access-bridge/src/services/passport-service.ts
```diff
@@ -0,0 +1,77 @@
...
+          // and rounding the result up to the nearest multiple of 300 milliseconds.
+          // This ensures that the expiration time is always on a 300-millisecond boundary.
+          exp: Math.ceil((now.getTime() + 3600) / 300) * 300,
```
budzworthy: What is the reason for doing the 5 minute boundaries?

platforms/access-bridge/test.html
budzworthy, line 1: Is this file worth keeping around or was it for temporary testing?

kmitrovv: Ah this shouldn't be here.. will remove it.

platforms/access-bridge/src/services/passport-service.ts
kmitrovv: https://docs.jwplayer.com/platform/reference/protect-your-content-with-signed-urls
That's suggested in our docs.

platforms/access-bridge/src/services/passport-service.ts
kmitrovv: you are right, will update this. thanks!

platforms/access-bridge/src/server.ts
budzworthy: Generally the idea with most of these request / response pipeline frameworks is to use the pipeline to your advantage so there can be a single place that 'universal' things happen, instead of having to duplicate it for each endpoint. This feels to me like one of those types of places, but if it's a difficult refactor we can leave it for a later date when we have time to look at it more.

budzworthy (approved):
A few small comments on the latest, but nothing blocking or that's not an easy fix!

platforms/access-bridge/src/server.ts
kmitrovv: It makes sense what you're saying. I would leave this for later as the tests are also based on this structure, it would require a little bit more effort to refactor it.

kmitrovv: This is now applied 😅

PR #584, opened 2024-07-24
Refactor / add className prop to header components

ChristiaanScheermeijer:
## Description

This PR adds `className` props to the header components to make it easier to change styling in different platforms. We currently need to write `[class^="_shelf"]` selectors which isn't ideal.


AntonLantukh (approved):

royschut (approved):
Nice!

PR #585, opened 2024-07-25
[IDM-169] - feat: stripe products

kmitrovv:
## Description

This PR follows up on [PR #582](https://github.com/jwplayer/ott-web-app/pull/582) and introduces functionality for listing Stripe products (plans). Specifically, this PR addresses the process for displaying available subscription plans to viewers.

When a viewer attempts to purchase a subscription or access a service, this endpoint will be invoked. It first queries SIMS to retrieve the available access plans (if any) associated with the current request. Subsequently, the endpoint calls Stripe to filter and retrieve the relevant products based on the provided access plan IDs. The Stripe products include all necessary metadata for presenting the options to viewers on the user interface.

There will be a follow up [PR](https://jwplayer.atlassian.net/browse/IDM-209) for initiating a Stripe checkout session.

Jira card: [List available plans](https://jwplayer.atlassian.net/browse/IDM-169)
### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


platforms/access-bridge/README.md
```diff
@@ -62,6 +62,33 @@ You can also copy and paste the contents of `.env.example` into `.env.local` and
   }
   ```
 
+#### URL: `/v2/sites/{site_id}/products`
+
+- **Method:** GET
+- **Authorization:** None
+- **Summary:** Lists all the corresponding stripe products with prices that are connected to the SIMS plans.
+- **Response:** Stripe.Product type
```
budzworthy: Maybe it's worth it to include a link to the latest stripe format doc and re-emphasize that this is just a passthrough of what stripe tells us

platforms/access-bridge/src/app-config.ts
budzworthy, line 1: Small nit, but can we rename this file to avoid confusion with the 'app config' json from OTT? Maybe just `utils/consts` or something?

platforms/access-bridge/src/controllers/products-controller.ts
```diff
@@ -0,0 +1,49 @@
...
+   * Service handler for fetching and returning Stripe products with prices based on available plans
+   * for a given site ID. Validates the site ID, retrieves and filters plans, and matches them with
+   * Stripe products via external provider IDs. Sends appropriate error responses for invalid requests.
+   *
+   * @param req - Express request object
+   * @param res - Express response object
+   * @param next - Express next middleware function
+   */
+  async getProducts(req: Request, res: Response, next: NextFunction): Promise<void> {
+    const siteId = req.params.site_id;
```
budzworthy: Do we need site ID to be a param to the middleware or can it be hard-coded in the ini/env values? It will have to match the api keys anyhow which are already statically set as env vars.

platforms/access-bridge/src/app-config.ts
```diff
@@ -6,6 +6,10 @@ function requireEnvVar(env: string | undefined, name: string): string {
   return env;
 }
 
+// Secrets resonsible for authenticating requests
+export const API_SECRET = requireEnvVar(process.env.APP_API_SECRET, 'APP_API_SECRET');
```
budzworthy, line 14: We've had a couple conversations about vite and env vars but I don't know if we've talked about these... are these intended to be injected at build time or runtime? Runtime is probably better, since we can then re-use the same build for different customers.

platforms/access-bridge/src/controllers/products-controller.ts
```diff
@@ -0,0 +1,49 @@
...
+import { StripeService } from '../services/stripe-service.js';
+import { ErrorDefinitions, sendErrors } from '../errors.js';
+import { isValidSiteId } from '../utils.js';
+
+/**
+ * Controller class responsible for handling Stripe-related services.
+ */
+export class ProductsController {
+  private plansService: PlansService;
+  private stripeService: StripeService;
```
budzworthy: Can we call this something more generic like `PaymentService`? And then`StripePaymentService` is an implementation of `PaymentService`.

platforms/access-bridge/src/controllers/products-controller.ts
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      const availablePlans = await this.plansService.getAvailablePlans({ siteId });
+      const stripeProductIds: string[] = availablePlans
+        .map((plan) => plan.metadata.external_providers?.stripe ?? [])
+        .flat();
+
+      const products = await this.stripeService.getProductsWithPrices(stripeProductIds);
+      res.json(products);
+    } catch (error) {
+      console.error('Controller: failed to get Stripe products.', error);
```
budzworthy: Do you need the try/catch here or can this be captured in a pipeline handler?

platforms/access-bridge/src/server.ts
```diff
@@ -66,6 +74,11 @@ export class Server {
       return;
     }
 
+    if (err instanceof Stripe.errors.StripeError) {
```
budzworthy: GlobalErrorHandler is getting big enough that it can probably move to its own file. Maybe consider something like a `/pipeline` folder where you can move `endpoints.ts` and `globalErrorHandler.ts` and maybe some other things.

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Service class responsible for interacting with the Stripe API to fetch products.
+ */
+export class StripeService {
+  private stripe: Stripe;
+
+  constructor() {
+    this.stripe = new Stripe(STRIPE_SECRET, {
+      apiVersion: '2024-06-20',
```
budzworthy: Can you leave a comment here explaining what this is for?

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  /**
+   * Retrieves Stripe products with prices and filters them based on the provided productIds.
+   * @param productIds The array of product IDs to filter the products.
+   * @returns A Promise resolving to an array of filtered ProductsWithMetadata objects.
+   */
+  async getProductsWithPrices(productIds: string[]): Promise<StripeProduct[]> {
```
budzworthy: Maybe this method should return the already-mapped results that we plan to use like:
```
{
  plan_id: 123456
  price: $55
  recurrence: daily
  description: ....
}
```
This is related to the suggestion above that this implement a generic `PaymentService` interface and decouples us from stripe at higher levels, keeping stripe specific logic encapsulated in the stripe service.

budzworthy: (the json above is just a sample. I don't know what we actually plan to pick out and use to show to viewers)

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -0,0 +1,60 @@
...
+          const prices = await this.stripe.prices.list({ product: product.id });
+          return {
+            ...product,
+            prices: prices.data,
+          };
+        } catch (priceError) {
+          console.error(`Failed to fetch prices for product ${product.id}:`, priceError);
+          return {
+            ...product,
+            prices: [], // Return an empty array if price retrieval fails
```
budzworthy: What's the idea behind this error path? Does it result in us showing viewers a plan that they can't buy? Maybe it's better to just let this fail the entire flow and we can show a message to viewers that there was a problem fetching plans with a refresh button.

platforms/access-bridge/test/fixtures.ts
```diff
@@ -146,3 +151,96 @@ export const AUTHORIZATION = {
...
+    divide_by: 1,
+    round: 'up',
+  },
+  type: 'recurring',
+  unit_amount: 1000,
+  unit_amount_decimal: '10.00',
+};
+
+// Stripe product mock
+export const STRIPE_PRODUCT: StripeProduct = {
```
budzworthy: This comment isn't really about this file, but it made me think. I'm really nervous about how we're going to nicely show the huge amount of data that stripe returns. What are your thoughts on it? I worry that we are going to have many of the same problems that Sims was having on the management side (there are just too many features for us to keep up with.)

Part of me want to just use the stripe pricing tables, so it's off of our plate, but that's a big product architecture change. Alternatively, maybe we could think of a slick way to define mappings that could be easily changed or even controlled via app config to let customers deal with it as they want.

Some limitations I see:
* showing a description about what the plan gives you access to and special features (downloads, number of streams, etc.) Do we just say that we use name and description?
* Making sure we show all the recurrence details properly (what are all the possible options from stripe?)
* Showing the right currency and tax for the viewer
* Showing all the details of the free trials (again what are all the possible options?)

Whatever we decide, this is definitely something we need to document clearly.

platforms/access-bridge/test/integration/products.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  const testCases = [
+    {
+      description: 'should list stripe products',
+      requestOptions: {
+        method: 'GET',
+        path: ENDPOINTS.PRODUCTS.replace(':site_id', SITE_ID.VALID),
+      },
+      mockBehavior: 'default' as MockBehavior,
+      expectedStatusCode: 200,
+      expectedResponse: [STRIPE_PRODUCT],
```
budzworthy: Might be nice to have a couple products defined so you can test the filtering

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -0,0 +1,60 @@
...
+   * @param productIds The array of product IDs to filter the products.
+   * @returns A Promise resolving to an array of filtered ProductsWithMetadata objects.
+   */
+  async getProductsWithPrices(productIds: string[]): Promise<StripeProduct[]> {
+    if (!productIds.length) {
+      return [];
+    }
+
+    // Fetch all products (pagination might be necessary for large datasets)
+    const allProducts = await this.stripe.products.list();
```
budzworthy: Do we need to do any filtering on stripe data like active or expired?

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -0,0 +1,60 @@
...
+    const filteredProducts = allProducts.data.filter((product) => productIds.includes(product.id));
+
+    if (filteredProducts.length === 0) {
+      return [];
+    }
+
+    // Retrieve prices for each filtered product
+    const productsWithPrices = await Promise.all(
+      filteredProducts.map(async (product) => {
```
budzworthy: Any thoughts on sorting? Do we just let the FE handle that?

platforms/access-bridge/src/controllers/products-controller.ts
kmitrovv: I wouldn't do that, then we will need to rename the routes without site_id, and lose consistency and visibility. We already have site_id info, not seeing any benefit doing that.

platforms/access-bridge/src/app-config.ts
kmitrovv: I had them in run time using dotEnv, but one of your suggestion was to use vite to inject them, so now they are injected at build time.
Btw, how can we reuse single deployment to different clients? They will still need to deploy it on their side.

platforms/access-bridge/src/services/stripe-service.ts
kmitrovv: It is exactly what it does. It retrieves the mapped Stripe product. We are retrieving the entire Stripe product as we don't wanna use our own typings in case of Stripe changes.

platforms/access-bridge/test/fixtures.ts
kmitrovv: I would say for now we just follow the designs and show the minimum required stuff that is necessary to make a subscription. 

Stripe pricing tables look promising, but wouldn't change it now in sake of the spotlight and it may affect SIMS side as well. 
Also, not seeing the downside of returning the entire Stripe product, as the customers can add features by their preferences. Isn't this supposed to serve only as a starting point? Why do we wanna handle cases we don't know will exist..

We can document what we support and of course it can always be extended as we add support to more features.

platforms/access-bridge/src/services/stripe-service.ts
kmitrovv: By default they are sorted by date of creation. If we need additional logic it can always be managed on the FE.

platforms/access-bridge/src/services/stripe-service.ts
kmitrovv: I've updated to not return the product if the price fails. So in the end if the products array is empty, the FE will handle it.

platforms/access-bridge/src/app-config.ts
kmitrovv: It's well commented what it's used for, not seeing how it can be confused. I don't think this is utils, but it's the main source of the configuration so that the app can run.

PR #587, opened 2024-07-29
[IDM-209] - feat: stripe checkout

kmitrovv:
## Description

This pull request implements the backend logic for creating a checkout session within the Access Bridge component. It establishes the necessary interactions with the Sims and Stripe services to validate access plans and create a checkout session.

The Access Bridge now includes the following backend interactions:

**Retrieve Available Plans:**

- Access Bridge queries Sims for available access plans for a specified site and validates the provided access_plan_id.

**Create Stripe Checkout Session:**

- Upon successful plan validation, Access Bridge requests Stripe to create a checkout session. This includes sending details like price_id, mode, and redirect_url. Also, the SIMS `viewer_id` is added as a metadata to the checkout session.

**Endpoint**:
`POST:  /v2/checkout`

Jira card: [Create checkout session](https://jwplayer.atlassian.net/browse/IDM-209)

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -76,4 +78,50 @@ export class StripeService {
...
+    try {
+      const session = await this.stripe.checkout.sessions.create({
+        payment_method_types: ['card'],
+        line_items: [
+          {
+            price: params.price_id,
+            quantity: 1,
+          },
+        ],
+        metadata: {
```
darkoatanasovski: `metadata` is just for **single payments**, for **subscriptions**, we need `subscription_metadata` parameter here.

platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -0,0 +1,65 @@
...
+import { StripeService } from '../services/stripe-service.js';
+import { AccessBridgeError, ErrorDefinitions, sendErrors } from '../errors.js';
+import { IdentityService } from '../services/identity-service.js';
+import { isValidSiteId } from '../utils.js';
+
+/**
+ * Controller class responsible for handling Stripe Checkout sessions.
+ */
+export class CheckoutController {
+  private identityService: IdentityService;
```
budzworthy: These can be readonly, no?

platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -0,0 +1,65 @@
...
+      const authorization = req.headers['authorization'];
+      if (!authorization) {
+        sendErrors(res, ErrorDefinitions.UnauthorizedError.create());
+        return;
+      }
+
+      const checkoutParams = req.body;
+
+      // Validate required params
+      const requiredParams: (keyof StripeCheckoutParams)[] = ['price_id', 'mode', 'success_url', 'cancel_url'];
```
budzworthy: Should the urls be passed from the FE as params or should they be hard coded in the middleware?

platforms/access-bridge/src/server.ts
```diff
@@ -69,6 +69,11 @@ export class Server {
    * Global error handler middleware for the server.
    */
   private globalErrorHandler = (err: unknown, req: Request, res: Response, next: NextFunction) => {
+    if (err instanceof SyntaxError && 'body' in err) {
```
budzworthy: Might be nice to comment this to explain what you're catching here

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -57,4 +60,40 @@ export class StripeService {
...
+      metadata: {
+        viewer_id: viewer.id,
+      },
+      customer_email: viewer.email,
+      mode: params.mode,
+      success_url: params.success_url,
+      cancel_url: params.cancel_url,
+
+      // Conditionally include `subscription_data` only if mode is `subscription`
+      ...(params.mode === 'subscription' && {
```
budzworthy: What is mode used for besides 'subscription'?

platforms/access-bridge/src/services/stripe-service.ts
budzworthy, line 1: Same comment here as the last PR. Would probably be nice to have the controllers use a generic `checkoutService` which `stripeService` can implement

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -57,4 +60,40 @@ export class StripeService {
...
+        subscription_data: {
+          metadata: {
+            viewer_id: viewer.id,
+          },
+        },
+      }),
+    };
+
+    return await this.stripe.checkout.sessions.create(sessionParams);
```
budzworthy: How will this be used on the FE? Would it save some FE work if we just return a redirect to the stripe paywall URL with the session included?

kmitrovv: It is exactly what it does, it returns url for the stripe checkout page. It's already used in some of the Vladimir's PRs.

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -57,4 +60,40 @@ export class StripeService {
...
+      customer_email: viewer.email,
+      mode: params.mode,
+      success_url: params.success_url,
+      cancel_url: params.cancel_url,
+
+      // Conditionally include `subscription_data` only if mode is `subscription`
+      ...(params.mode === 'subscription' && {
+        subscription_data: {
+          metadata: {
+            viewer_id: viewer.id,
```
budzworthy: Please add a comment here for how important this is to the current implementation (it's our only way of connecting the payment back to this user.) I really want to revisit this approach in Q4, because this does not give me a warm fuzzy sense of reliability. Do you know why we aren't creating customers in stripe or using the reference id?
https://docs.stripe.com/api/customers/object
https://docs.stripe.com/api/checkout/sessions/object#checkout_session_object-client_reference_id

platforms/access-bridge/src/utils.ts
```diff
@@ -3,3 +3,18 @@ export function isValidSiteId(siteId: string): boolean {
...
   return alphanumericRegex.test(siteId);
 }
+
+/**
+ * Checks if the required parameters are present in the provided object.
+ * @param params - The object containing the parameters to be validated.
+ * @param requiredParams - The list of required keys to check for presence.
+ * @returns An array of missing required keys.
+ */
+export const validateBodyParams = <T>(params: Partial<T>, requiredParams: (keyof T)[]): (keyof T)[] => {
```
budzworthy: Looks like this is unused? Did you mean to use it in checkout controller?

kmitrovv: This is forgotten here. It is already removed in one of the following PRs.

platforms/access-bridge/src/services/stripe-service.ts
kmitrovv: Stripe creates a customer out of the box when a payment is made. Sims team already tested and used this stuff and no complains were made about something missing. We can always add some stuff if needed.

platforms/access-bridge/src/services/stripe-service.ts
kmitrovv: Mode can be also `payment` (for a single payment and not a subscription), not that we support it atm, but it will be eventually.

platforms/access-bridge/src/controllers/checkout-controller.ts
kmitrovv: Which urls do you mean? The endpoints themselves? Why do we wanna hardcode them? what do we gain, when we have all the needed info on the FE. Plus the endpoints can be easily tested via API if the site_id is exposed in the route.

PR #589, opened 2024-07-31
Refactor / fix delivery api ad schedule typings

### LINKED ISSUE #472: AdSchedule type doesn't match actual data type
In the `Player.tsx`, an `adsData` prop uses the `AdSchedule` type. This type is defined as follows:

```ts
export type AdSchedule = {
  adscheduleid?: string;
  client: string;
  schedule: string;
};
```

However, the data seems to differ from the type above:

```json
{
    "rules": {
        "startOnSeek": "pre",
        "timeBetweenAds": 0
    },
    "client": "vast",
    "schedule": [
        {
            "tag": [
                "https://pubads.g.doubleclick.net/gampad/ads?iu=/21775744923/external/single_preroll_skippable&sz=640x480&ciu_szs=300x250%2C728x90&gdfp_req=1&output=vast&unviewed_position_start=1&env=vp&impl=s&correlator=__random-number__"
            ],
            "type": "linear",
            "offset": "pre",
            "skipoffset": 5
        }
    ],
    "adscheduleid": "S2FKhc4J"
}
```

ChristiaanScheermeijer:
## Description

This PR changes the adSchedule typings to include the correct typings when served from the delivery API.

Fixes #472

### DIFF
--- a/packages/hooks-react/src/useAds.ts
+++ b/packages/hooks-react/src/useAds.ts
@@ -3,7 +3,7 @@ import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
 import ApiService from '@jwp/ott-common/src/services/ApiService';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import { createURL } from '@jwp/ott-common/src/utils/urlFormatting';
-import type { AdSchedule } from '@jwp/ott-common/types/ad-schedule';
+import type { AdConfig } from '@jwp/ott-common/types/ad-schedule';
 
 const CACHE_TIME = 60 * 1000 * 20;
 
@@ -39,7 +39,7 @@ export const useAds = ({ mediaId }: { mediaId: string }) => {
         schedule: createURL(adScheduleUrls?.xml || '', {
           media_id: mediaId,
         }),
-      } as AdSchedule)
+      } as AdConfig)
     : undefined;
 
   return {
--- a/packages/hooks-react/src/useAds.ts
+++ b/packages/hooks-react/src/useAds.ts
@@ -3,6 +3,7 @@ import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
 import ApiService from '@jwp/ott-common/src/services/ApiService';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import { createURL } from '@jwp/ott-common/src/utils/urlFormatting';
+import type { AdSchedule } from '@jwp/ott-common/types/ad-schedule';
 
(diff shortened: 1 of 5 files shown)


ChristiaanScheermeijer:
@dbudzins @AntonLantukh what do you think? 😄

packages/hooks-react/src/useAds.ts
```diff
@@ -33,12 +34,12 @@ export const useAds = ({ mediaId }: { mediaId: string }) => {
...
   const { data: adSchedule, isLoading: isAdScheduleLoading } = useLegacyStandaloneAds({ adScheduleId, enabled: !!adScheduleId });
   const adConfig = useAdConfigFlow
-    ? {
+    ? ({
         client: 'vast',
         schedule: createURL(adScheduleUrls?.xml || '', {
           media_id: mediaId,
         }),
-      }
+      } as AdSchedule)
```
AntonLantukh: Can we rename it to AdConfig and use the following type:

```
export type AdConfig = {
  client: string;
  schedule: string;
};
```

And then we could use AdSchedule instead of DeliveryAdSchedule

ChristiaanScheermeijer: Sure! I also considered something like this, but didn't changed it because both types are actually ad configs containing schedules.

packages/hooks-react/src/useAds.ts
```diff
@@ -33,12 +34,12 @@ export const useAds = ({ mediaId }: { mediaId: string }) => {
...
   const { data: adSchedule, isLoading: isAdScheduleLoading } = useLegacyStandaloneAds({ adScheduleId, enabled: !!adScheduleId });
   const adConfig = useAdConfigFlow
-    ? {
+    ? ({
         client: 'vast',
         schedule: createURL(adScheduleUrls?.xml || '', {
           media_id: mediaId,
         }),
-      }
+      } as AdConfig)
```
AntonLantukh, line 42: Do you think we can also remove this one? Looks like there are no errors.

ChristiaanScheermeijer: I thought the data typing was a bit clearer when forcing it to `AdConfig`. Without it infers the typings and looks like this:

```ts
function useAds({mediaId}: {mediaId: string}): {isLoading: false | true, data: {schedule: string, client: string} | undefined | AdSchedule}
```

VS

```ts
function useAds({mediaId}: {mediaId: string}): {isLoading: false | true, data: AdConfig | undefined | AdSchedule}
```

Without the typing, it still satisfies to `AdConfig` though...

AntonLantukh (approved):

PR #593, opened 2024-08-13
OWA-85: New payment flow

mirovladimitrovski:
## Description

This PR solves [#OWA-85](https://jwplayer.atlassian.net/browse/OWA-85).

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code


packages/common/src/env.ts
```diff
@@ -14,7 +14,7 @@ export type Env = {
 
 const env: Env = {
   APP_VERSION: '',
-  APP_API_BASE_URL: 'https://cdn.jwplayer.com',
+  APP_API_BASE_URL: 'https://cdn-dev.jwplayer.com',
```
darkoatanasovski:
```suggestion
  APP_API_BASE_URL: 'https://cdn.jwplayer.com',
```

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -18,23 +18,36 @@ type RequestOptions = {
...
+  private getBaseUrl = (fromSimsClient = false) => {
+    if (fromSimsClient) {
+      return `http://localhost:3000/v2/sites/${this.siteId}`;
+    }
+
+    if (this.useSandboxEnv) {
+      return 'https://daily-sims.jwplayer.com';
+    }
+
+    return 'http://sims.jwplayer.com';
```
darkoatanasovski:
```suggestion
    return 'https://sims.jwplayer.com';
```

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -18,23 +18,36 @@ type RequestOptions = {
...
-  setup = (useSandboxEnv: boolean) => {
+  setup = (useSandboxEnv: boolean, siteId: string) => {
     this.useSandboxEnv = useSandboxEnv;
+    this.siteId = siteId; // 'a1NXSCNM';
   };
 
-  private getBaseUrl = () => (this.useSandboxEnv ? 'https://staging-sims.jwplayer.com' : 'http://sims.jwplayer.com');
+  private getBaseUrl = (fromSimsClient = false) => {
+    if (fromSimsClient) {
+      return `http://localhost:3000/v2/sites/${this.siteId}`;
```
darkoatanasovski: the host should be defined in ENV

darkoatanasovski (dismissed):

packages/common/src/env.ts
mirovladimitrovski: That way it'll fail fetching the config.

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
```diff
@@ -23,19 +25,21 @@ import type { ServiceResponse } from '../../../../types/service';
...
+  StripeProduct,
+  StripePrice,
 } from './types';
 import JWPAPIService from './JWPAPIService';
 
 @injectable()
 export default class JWPCheckoutService extends CheckoutService {
   private readonly cardPaymentProvider = 'stripe';
 
+  accessMethod: AccessMethod = 'plan';
```
kmitrovv: Do we need this?

packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -2,6 +2,7 @@ import { inject, injectable } from 'inversify';
```
kmitrovv, line 2: We can revert this as it was, and define the needed methods in the ApiService. They need to be usable with other integrations as well. The access bridge host can be used from the config settings, as it will have only one environment and is set up in the ini file.

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
mirovladimitrovski: Well, yes
image

kmitrovv: We don't need additional setter to determine that, as we will have to introduce one for the access bridge. 

If the environment variable for the Access Bridge is defined, we know for sure that the new flow will be used. It can be part of the config store even before initalization. 

I had similar case here, so I was thinking something like this that we can reuse throughout entire app, and not just on the payment side.
https://github.com/jwplayer/ott-web-app/pull/603/files#diff-5249be0266da3296ee9e5a0f29405b5d56bc6f629025e88fb421964ec0edff70R49

.github/workflows/test-unit-snapshot.yml
```diff
@@ -1,4 +1,4 @@
-name: Test - Unit and Snapshot
+name: Test - Unit, Snapshot
```
budzworthy: Let's not rename the workflows unless it's a meaningful change, otherwise it makes the github actions list messy
```suggestion
name: Test - Unit and Snapshot
```

kmitrovv: This comes probably from one of my PRs. I did rename it there to: `Test - Unit, Integration, Snapshot`, as there are tests for the middleware as well.. Probably just not reverted properly here.

mirovladimitrovski: Yes, the last change I had pulled from the previous base branch was `Test - Unit, Integration, Snapshot` and I simply removed the word `Integration, ` - corresponding to the changes I was to push.
Changing it back to `Unit and Snapshot` right away.

packages/common/src/services/integrations/jwp/JWPAccountService.ts
```diff
@@ -65,7 +57,7 @@ export default class JWPAccountService extends AccountService {
 
   protected clientId = '';
 
-  accessModel: AccessModel = ACCESS_MODEL.AUTHVOD;
+  accessModel: AccessModel = ACCESS_MODEL.SVOD;
```
kmitrovv: forgotten?

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
```diff
@@ -23,19 +25,21 @@ import type { ServiceResponse } from '../../../../types/service';
 
 import type {
   CommonResponse,
-  GetAccessFeesResponse,
-  AccessFee,
   MerchantPaymentMethod,
   GeneratePayPalParameters,
   VoucherDiscountPrice,
   GetItemAccessResponse,
+  StripeProduct,
```
kmitrovv: eliminate these, use the defined types from common/payment

packages/common/src/services/integrations/jwp/JWPCheckoutService.ts
```diff
@@ -136,20 +115,51 @@ export default class JWPCheckoutService extends CheckoutService {
...
-      payload.offerIds.map(async (offerId) => {
-        try {
-          const data = await this.apiService.get<GetAccessFeesResponse>(`/v2/items/${this.parseOfferId(offerId)}/access-fees`);
+  chooseOffer: ChooseOffer = async ({ offer: { offerId }, successUrl, cancelUrl }) => {
+    try {
+      const { url } = await this.apiService.post<{ url: string }>(
+        '/v2/sites/:siteId/checkout',
+        {
+          price_id: offerId,
+          mode: 'subscription',
```
kmitrovv: this is redundant now

packages/common/src/services/integrations/jwp/types.ts
```diff
@@ -345,3 +345,29 @@ export type SetDefaultCardResponse = {
   exp_month: string;
   exp_year: string;
 };
+
```
kiremitrov123, line 348: Remove all of these types and import from common/payment

packages/common/types/stripe.ts
```diff
@@ -0,0 +1,10 @@
+export type StripeCheckoutParams = {
```
kmitrovv: This file should be removed now.

packages/hooks-react/src/useEntitlement.ts
```diff
@@ -43,9 +46,12 @@ const useEntitlement: UseEntitlement = (playlistItem) => {
   );
 
   const checkoutController = getModule(CheckoutController, false);
+  const jwpEntitlementService = getModule(JWPEntitlementService);
 
-  const isPreEntitled = playlistItem && !isLocked(accessModel, !!user, !!subscription, playlistItem);
-  const mediaOffers = playlistItem?.mediaOffers || [];
+  const accessMethod = useCheckoutStore((state) => state.accessMethod);
```
kmitrovv: Revisit this logic here after removing the accessModel

packages/ui-react/src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -40,15 +44,38 @@ const ChooseOffer = () => {
 
       if (!offer) return;
 
-      await chooseOffer.mutateAsync(offer);
+      const baseUrl = window.location.origin;
+
+      const url = await chooseOffer.mutateAsync({
```
kiremitrov123, line 49: This will break current implementations. It is also used for update subscription and selecting an offer from Cleen implementation, should not be invoked if access bridge is not present.

packages/common/src/controllers/AccessController.ts
```diff
@@ -62,23 +61,27 @@ export default class AccessController {
...
-        const accessTokens = await this.refreshAccessTokens();
-        if (accessTokens?.passport) {
-          return await this.apiService.getMediaByIdWithPassport({ id: mediaId, siteId: this.siteId, planId: entitledPlan.id, passport: accessTokens.passport });
-        }
-
-        throw new Error('Failed to refresh access tokens and retrieve media.');
-      }
-      throw error;
+      // If the initial attempt fails, it may indicate the passport does not have the latest plans.
+      // Force new access tokens generation and retry fetching the media.
```
AntonLantukh, line 82: And additional time we need to wait before retrying?

packages/ui-react/src/containers/StartWatchingButton/StartWatchingButton.tsx
```diff
@@ -31,14 +32,20 @@ const StartWatchingButton: React.VFC<Props> = ({ item, playUrl, disabled = false
...
+  }));
+
   const user = useAccountStore((state) => state.user);
   const isLoggedIn = !!user;
 
   // watch history
   const watchHistoryItem = useWatchHistoryStore((state) => item && state.getItem(item));
   const videoProgress = watchHistoryItem?.progress;
 
+  useOffers();
```
AntonLantukh: What is the idea of using it this way?

platforms/web/test-e2e/tests/offers/choose_offer_test.ts
```diff
@@ -28,10 +28,10 @@ const cleengProps: ProviderProps = {
   hasInlineOfferSwitch: false,
 };
 
-runTestSuite(jwProps, 'JW Player');
+runTestSuite(jwProps, 'JW Player', true);
```
AntonLantukh: Does it make sense to have this suite if it is always skipped?

platforms/web/test-e2e/tests/payments/coupons_test.ts
```diff
@@ -31,10 +31,10 @@ const cleengProps: ProviderProps = {
   hasInlineOfferSwitch: false,
 };
 
-runTestSuite(jwProps, 'JW Player');
+runTestSuite(jwProps, 'JW Player', true);
```
AntonLantukh: Same question about this test suite.

platforms/web/test-e2e/tests/payments/payments_test.ts
```diff
@@ -29,10 +29,10 @@ const cleengProps: ProviderProps = {
   hasInlineOfferSwitch: false,
 };
 
-runTestSuite(jwProps, 'JW Player');
+runTestSuite(jwProps, 'JW Player', true);
```
AntonLantukh: Test suite question

platforms/web/test-e2e/tests/register_test.ts
```diff
@@ -149,8 +151,8 @@ function runTestSuite(config: typeof testConfigs.svod, providerName: string) {
     I.fillField('lastName', 'Doe');
 
     I.click('Continue');
-    I.waitForLoaderDone();
+    I.waitForLoaderDone(20);
 
-    I.see('Welcome to JW OTT Web App (AuthVod)');
+    I.see('Payment successful');
```
AntonLantukh: Not sure this is the right wording after registration.

platforms/web/test-e2e/utils/payments.ts
```diff
@@ -31,8 +31,8 @@ export function formatDate(date: Date) {
 
 export async function finishSubscription(I: CodeceptJS.I) {
   I.click('Continue');
-  I.waitForText(`Welcome to JW OTT Web App (SVOD)`, longTimeout);
-  I.waitForText(`Thank you for subscribing to JW OTT Web App (SVOD). Please enjoy all our content.`);
+  I.waitForText('Payment successful', longTimeout);
```
AntonLantukh: Why did we change the wording here?

platforms/web/public/locales/en/account.json
```diff
@@ -50,8 +50,8 @@
     "upgrade_success": "Subscription updated!",
     "upgrade_success_message": "Your subscription is updated and can be used immediately.",
     "waiting_for_payment": "Waiting for payment",
-    "welcome_description": "Thank you for subscribing to {{siteName}}. Please enjoy all our content.",
-    "welcome_title": "Welcome to {{siteName}}",
+    "welcome_description": "Thank you for your subscription, enjoy all our content.",
+    "welcome_title": "Payment successful",
```
AntonLantukh: Why did we change it here?

platforms/web/public/locales/en/account.json
```diff
@@ -30,7 +30,7 @@
     "monthly": "Monthly subscription",
     "no_payment_needed": "No payment needed",
     "payment_cancelled": "Payment cancelled",
-    "payment_cancelled_message": "You have cancelled the payment",
+    "payment_cancelled_message": "This payment attempt was cancelled and no charges were made. If this wasn’t intentional, please retry the payment process.",
```
AntonLantukh: Do we also use this wording for Cleeng?

packages/ui-react/src/pages/User/User.tsx
```diff
@@ -30,10 +36,11 @@ const User = (): JSX.Element => {
   const favoritesController = getModule(FavoritesController);
   const accountController = getModule(AccountController);
 
-  const { accessModel, favoritesList } = useConfigStore(
+  const { accessModel, favoritesList, isAccessBridgeEnabled } = useConfigStore(
```
AntonLantukh, line 39: I doubt we should name it this way. Maybe we should better use something like `isJWPSVODEnabled`?

packages/ui-react/src/components/Welcome/Welcome.tsx
```diff
@@ -17,10 +18,17 @@ const Welcome: React.FC<Props> = ({ onCloseButtonClick, onCountdownCompleted, si
   const countdown = useCountdown(10, 1, onCountdownCompleted);
 
   return (
-    <div className={styles.welcome}>
-      <h2>{t('checkout.welcome_title', { siteName })}</h2>
-      <p>{t('checkout.welcome_description', { siteName })}</p>
-      <Button variant="contained" color="primary" label={t('checkout.start_watching', { countdown })} onClick={onCloseButtonClick} size="large" fullWidth />
+    <div className={styles.topContainer}>
+      <div className={styles.circleHug}>
+        <CheckmarkSVG />
```
AntonLantukh: We also use this screen for AuthVOD where no payment is required.

packages/ui-react/src/components/UserMenuNav/UserMenuNav.tsx
```diff
@@ -29,6 +30,8 @@ const UserMenuNav = ({ showPaymentItems, small = false, onButtonClick, favorites
   const accountController = getModule(AccountController);
   const tabIndex = focusable ? 0 : -1;
 
+  const isAccessBridgeEnabled = useConfigStore((state) => !!state.settings?.apiAccessBridgeUrl);
```
AntonLantukh, line 33: I would also prefer to use a different name here indicating that we use it for the particular case.

packages/common/src/controllers/AccountController.ts
```diff
@@ -203,6 +196,8 @@ export default class AccountController {
 
       if (response) {
         const { user, customerConsents } = response;
+        await this.accessController?.generateAccessTokens();
+        await this.getEntitledPlans();
```
AntonLantukh, line 200: Do we also make these requests for Cleeng?

packages/common/src/controllers/AccountController.ts
```diff
@@ -173,6 +165,7 @@ export default class AccountController {
 
       if (response) {
         await this.accessController?.generateAccessTokens();
+        await this.getEntitledPlans();
```
AntonLantukh, line 168: Do we need to do it if cleeng is enabled?

packages/ui-react/src/components/Subscription/Subscription.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import userStyles from '../../pages/User/User.module.scss';
+import Button from '../Button/Button';
+import LoadingOverlay from '../LoadingOverlay/LoadingOverlay';
+
+import styles from './Subscription.module.scss';
+
+const ActiveSubscription: React.FC<{ subscription: SubscriptionType }> = ({ subscription }) => {
+  const { t, i18n } = useTranslation('user');
+  const isAccessBridgeEnabled = useConfigStore((state) => !!state.settings?.apiAccessBridgeUrl);
```
AntonLantukh, line 18: Do we need "Subscription" component without access bridge?

packages/ui-react/src/containers/StartWatchingButton/StartWatchingButton.tsx
mirovladimitrovski: If you first open the `choose-offer` modal it'll initiate fetching the offers, and if you then arrive at a media page where the `StartWatchingButton` is rendered, its label will be `Subscribe to watch` (or something similar, doesn't matter the exact wording) accordingly. If you arrive at that same page without visiting the `choose-offer` modal, the label will be "Start watching" and you'd be redirected to watch the video and be never be shown the offers modal. But since you won't have access to the media, you won't be able to watch it.

In short - the `StartWatchingButton` needs the offers fetched to have the correct message and behavior.

packages/common/src/controllers/AccessController.ts
kmitrovv: Noup, the most common usage of this is after a payment when there is a need of new passport. We can instantly get a new one if this is the case don't need to wait. Btw, this will be updated in Q4, when we add a mechanism to update the passport on SIMS side.

platforms/web/test-e2e/tests/offers/choose_offer_test.ts
mirovladimitrovski: It's not the entire suite that gets skipped, just specific scenarios which rely on requests to the access bridge. You can easily navigate which are those - the ones that use the `isAccessBridgeEnabled` parameter to skip.

kmitrovv: I think these are skipped until we decide how we wanna have the access bridge scenario tested.

platforms/web/test-e2e/tests/payments/coupons_test.ts
mirovladimitrovski: Same answer. :)

platforms/web/test-e2e/tests/payments/payments_test.ts
mirovladimitrovski: Same answer. :)

packages/ui-react/src/pages/User/User.tsx
kmitrovv: It's should be as is, as we see the access bridge to be injectable with other integrations as well in the future.

packages/common/src/controllers/AccountController.ts
mirovladimitrovski: @kiremitrov123 ?

packages/ui-react/src/components/UserMenuNav/UserMenuNav.tsx
kmitrovv: It should remain as is, it's not JWP specific. Should be / will be used by other integrations.

packages/common/src/controllers/AccountController.ts
kmitrovv: No, there is a condition in the method if the ac bridge is not enabled to not fire it.

packages/common/src/controllers/AccountController.ts
kmitrovv: It is not enabled per integration, but per access bridge presence.

packages/ui-react/src/components/Subscription/Subscription.tsx
mirovladimitrovski: Basically the correlation is rather: if it's offers (only Cleeng ATM) then "Payment", if it's acces plans (SVOD, only JWP ATM) then "Subscription".

packages/ui-react/src/components/Welcome/Welcome.tsx
mirovladimitrovski: You're right, good catch.

platforms/web/test-e2e/tests/offers/choose_offer_test.ts
AntonLantukh: I would suggest to comment out this part instead of adding condition to every test.

mirovladimitrovski: Don't we want to run the scenarios that can be run? It'll be easier to track the scenarios we'll need to revisit to unskip later.

packages/ui-react/src/components/Subscription/Subscription.tsx
mirovladimitrovski: To answer your question fully - no, the way things are right now, the `Subscription` component and what it's supposed to render relies on the access bridge.  It's the only way for the `Go to Stripe Billing portal` button to serve its purpose.

platforms/web/test-e2e/tests/register_test.ts
mirovladimitrovski: I reverted this and all the changes in the `Welcome` component and its accompanying `test.tsx`, `module.css` and snap files and moved all the changes previously done to a new component `PaymentSuccessful`.

platforms/web/test-e2e/utils/payments.ts
mirovladimitrovski: I reverted this and all the changes in the `Welcome` component and its accompanying `test.tsx`, `module.css` and snap files and moved all the changes previously done to a new component `PaymentSuccessful`.

platforms/web/public/locales/en/account.json
mirovladimitrovski: I reverted this and all the changes in the `Welcome` component and its accompanying `test.tsx`, `module.css` and snap files and moved all the changes previously done to a new component `PaymentSuccessful`.

packages/ui-react/src/components/Welcome/Welcome.tsx
mirovladimitrovski: I reverted this and all the changes in the `Welcome` component and its accompanying `test.tsx`, `module.css` and snap files and moved all the changes previously done to a new component `PaymentSuccessful`.

platforms/web/public/locales/en/account.json
mirovladimitrovski: I reverted this and all the changes in the `Welcome` component and its accompanying `test.tsx`, `module.css` and snap files and moved all the changes previously done to a new component `PaymentSuccessful`.

kmitrovv (approved):

platforms/web/test-e2e/tests/offers/choose_offer_test.ts
AntonLantukh: It makes sense if there is something to test, but here they are basically skipped. We can create a ticket to make them work again.

kmitrovv: I think somewhere in Q4 we need to decide how we wanna test the new flow.. We will definitely need some mocks as we don't want to expose a deployed access bridge version to the open source.

packages/common/src/controllers/AccountController.ts
AntonLantukh: Which condition do you mean?

```
generateAccessTokens = async (): Promise<AccessTokens | null> => {
    if (!this.siteId) {
      return null;
    }

    const auth = await this.accountService.getAuthData();

    const accessTokens = await this.accessService.generateAccessTokens(this.siteId, auth?.jwt);
    if (accessTokens) {
      await this.setAccessTokens(accessTokens);
      return accessTokens;
    }

    return null;
  };
```

packages/ui-react/src/components/Subscription/Subscription.tsx
AntonLantukh: Thanks. I was just thinking if we could add this condition on the top level to make it clear.

packages/common/src/controllers/AccountController.ts
kmitrovv: Here in the method definition if you are asking for the entitled plans: https://github.com/jwplayer/ott-web-app/blob/ec61a8f745345e49ad4291d93de3c06c1fbfbeec/packages/common/src/controllers/AccountController.ts#L396 

For the access controller, it won't fire if it's not initialized, that's why it's optional. On the access controller initalize we set the site id, so if it's not initialized, it won't be set and will return null.

packages/ui-react/src/components/Subscription/Subscription.tsx
mirovladimitrovski: The `Subscription` component is a route that gets redirected to from menu options that are rendered conditionally in `User.tsx` and `UserMenuNav.tsx`, the condition being the same `isAccessBridgeEnabled`, anyway.

packages/ui-react/src/containers/StartWatchingButton/StartWatchingButton.tsx
AntonLantukh: Maybe we could use "isLoading" state to pass it to the button while we fetch the data?

mirovladimitrovski: Done.

platforms/web/test-e2e/tests/offers/choose_offer_test.ts
mirovladimitrovski: I really think it was a good idea to have the affected scenarios marked, but I removed the `isAccessBridgeEnabled` altogether and disabled this line so we can proceed and merge.

platforms/web/test-e2e/tests/payments/payments_test.ts
mirovladimitrovski: I really think it was a good idea to have the affected scenarios marked, but I removed the `isAccessBridgeEnabled` altogether and disabled this line so we can proceed and merge.

platforms/web/test-e2e/tests/payments/coupons_test.ts
mirovladimitrovski: I really think it was a good idea to have the affected scenarios marked, but I removed the `isAccessBridgeEnabled` altogether and disabled this line so we can proceed and merge.

packages/common/src/controllers/AccessController.ts
```diff
@@ -62,23 +61,27 @@ export default class AccessController {
...
+      // TODO: Revisit this logic once we address the passport update possibility in Q4
```
AntonLantukh, line 83: @kiremitrov123 I would appreciate if you mention all the possible improvements marked as TODO in a separate ticket, so that we had more visibility here.

kmitrovv: We don't have tickets yet, this will all be a part of the new spotlight. But as soon as we create them I can provide an update with the corresponding tickets.

ChristiaanScheermeijer:
@kiremitrov123 @mirovladimitrovski thanks for your work on this feature! I'm currently reviewing and testing and I noticed that the JWP integration without access bridge doesn't work anymore. Is this expected or should it still be possible to use InPlayer (JWP) directly?

If not, it means that we can't upgrade when this is merged for existing customers + InPlayer.

image

kmitrovv:
> @kiremitrov123 @mirovladimitrovski thanks for your work on this feature! I'm currently reviewing and testing and I noticed that the JWP integration without access bridge doesn't work anymore. Is this expected or should it still be possible to use InPlayer (JWP) directly?
> 
> If not, it means that we can't upgrade when this is merged for existing customers + InPlayer.
> 
> image

@ChristiaanScheermeijer correct, we are eliminating the old flow for JW. It will be only possible with using the Access Bridge.

packages/ui-react/src/components/Button/Button.tsx
```diff
@@ -71,9 +75,11 @@ const Button: React.FC<ButtonProps> = ({
     </>
   );
 
-  if (to) {
+  if (to || toModal) {
+    const path = createURL(to || `${window.location.pathname}${window.location.search}`, { u: toModal });
```
ChristiaanScheermeijer: This seems unnecessary, I would just use `modalURLFromWindowLocation` in the implementation of the button. The Button component should be as it is used on other platforms as well (without modal routing).

packages/theme/assets/icons/check-green.svg
ChristiaanScheermeijer, line 1: I would make it colorless and use the `Icon` + CSS#fill property for colouring instead.

Should we use material-ui icons? Now we have two icons that don't match the icon set anymore 😅

packages/ui-react/src/components/ChoosePlanForm/ChoosePlanForm.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+        offers={svodOffers}
+      />,
+    );
+
+    fireEvent.submit(getByTestId('choose-offer-form'));
+
+    expect(onSubmit).toBeCalled();
+  });
+
+  test('WCAG 2.2 (AA) compliant', async () => {
```
AntonLantukh, line 103: I appreciate additional tests here! Really nice.

packages/common/src/services/PaymentService.ts
```diff
@@ -0,0 +1,93 @@
...
+import { API_ACCESS_BRIDGE_URL, INTEGRATION_TYPE } from '../modules/types';
+import type { IntegrationType } from '../../types/config';
+import type { Product } from '../../types/payment';
+import { logError } from '../logger';
+import { getNamedModule } from '../modules/container';
+
+import AccountService from './integrations/AccountService';
+
+@injectable()
+export default class PaymentService {
```
ChristiaanScheermeijer, line 12: Should this be renamed 🤔? This is actually the `AccessBridgeService` integration and not a payment service. It might be confusing with the CheckoutService.

packages/ui-react/src/components/PaymentSuccessful/PaymentSuccessful.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+  test('calls the onCloseButtonClick callback when clicking the close button', () => {
+    const onCloseButtonClick = vi.fn();
+    const { getByText } = render(<PaymentSuccessful onCloseButtonClick={onCloseButtonClick} siteName="Sitename!" />);
+
+    fireEvent.click(getByText('checkout.start_watching'));
+
+    expect(onCloseButtonClick).toBeCalled();
+  });
+
+  test('calls the onCloseButtonClick callback when clicking the close button', () => {
```
AntonLantukh:
```suggestion
  test('calls the onCloseButtonClick callback when timer expires', () => {
```

packages/common/src/services/PaymentService.ts
kmitrovv: The Access related methods are in the AccessService, these are the payment related definitions that will be the one source of truth hopefully :D. It is confirmed with @dbudzins for the naming.

platforms/web/test-e2e/tests/payments/payments_test.ts
```diff
@@ -31,7 +31,7 @@ const cleengProps: ProviderProps = {
   cardInfo: Array.of(['Card number', '•••• •••• •••• 1115'], ['Expiry date', '03/2030'], ['Security code', '******']),
 };
 
-runTestSuite(jwProps, 'JW Player');
+// runTestSuite(jwProps, 'JW Player');
```
ChristiaanScheermeijer, line 34: This is not finished yet? Will it be possible to do e2e testing with the access bridge? 🤔

kmitrovv: Actually we don't want to have exposed a deployed version.. We will need to come up with some mocks probably.

packages/ui-react/src/containers/AccountModal/forms/ChooseOffer.tsx
```diff
@@ -57,29 +84,54 @@ const ChooseOffer = () => {
...
     return (
       <div style={{ height: 300 }}>
         <LoadingOverlay inline />
       </div>
     );
   }
 
+  if (isAccessBridgeEnabled) {
+    return (
+      <ChoosePlanForm
```
AntonLantukh, line 123: @mirovladimitrovski do you see the way to fully switch to `ChoosePlanForm` and remove `ChooseOfferForm` in the future?

packages/theme/assets/icons/check-green.svg
mirovladimitrovski: Good suggestion, but I wouldn't delay merging this PR any further and would suggest leaving this for later.

packages/ui-react/src/components/Button/Button.tsx
mirovladimitrovski: Is it crucial to revert this? I usually prefer when something works as a link to behave that way in every way (including on hover). Also the link implementation was already at place and I just wanted to make a convenient way to implement opening a modal by clicking our common `Button` component. I really prefer that over doing it programmatically with `useNavigation` and `modalURLFromWindowLocation`.

packages/hooks-react/src/useBillingPortal.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  const redirectToBillingPortal = useCallback(async () => {
+    const billingPortalUrl = await billingPortalUrlRequest.mutateAsync();
+
+    setRedirecting(true);
+
+    if (billingPortalUrl) {
+      window.location.href = billingPortalUrl;
+    }
+  }, [billingPortalUrlRequest]);
```
ChristiaanScheermeijer: This assumes that this hook is only used in a browser. The [hooks-react](https://github.com/jwplayer/ott-web-app/blob/develop/docs/workspaces.md#react-hooks) is platform-agnostic and can't use `window.location` directly.

packages/ui-react/src/components/Button/Button.tsx
ChristiaanScheermeijer: It's not crucial, but if we don't keep things clean now (before merging), chances are it will not be cleaned up at all 😄

packages/theme/assets/icons/check-green.svg
mirovladimitrovski: Also, which icon set are you referring to? These are icons I got from our design team specifically to use in the new modals introduced here, and material-ui is not in place as a dependency.

packages/ui-react/src/containers/AccountModal/forms/ChooseOffer.tsx
mirovladimitrovski: It'll certainly be preferrable to have a unified flow, so yes.

packages/ui-react/src/components/Button/Button.tsx
mirovladimitrovski: Done.

packages/hooks-react/src/useBillingPortal.tsx
mirovladimitrovski: Fixed this.

AntonLantukh (approved):
LGTM. Still suggest to address url things.

mirovladimitrovski:
> LGTM. Still suggest to address url things.

If I'm correct about what you're referring to, I've already addressed them.

packages/hooks-react/src/useBillingPortalURL.tsx
```diff
@@ -0,0 +1,14 @@
...
+import CheckoutController from '@jwp/ott-common/src/controllers/CheckoutController';
+
+const useBillingPortalURL = () => {
+  const checkoutController = getModule(CheckoutController);
+
+  return useMutation({
+    mutationKey: 'billing-portal',
+    mutationFn: () => checkoutController.generateBillingPortalUrl(window.location.href),
+  });
+};
```
ChristiaanScheermeijer: The current URL should be forwarded (now it is still using `window.location`):

```tsx
const useBillingPortalURL = (returnUrl: string) => {
  // ...
 mutationFn: () => checkoutController.generateBillingPortalUrl(returnUrl),
```

packages/ui-react/src/components/Subscription/Subscription.tsx
```diff
@@ -61,7 +75,7 @@ const Subscription: React.FC<Props> = ({ subscription }) => {
       ) : (
         <>
           <p>{t('user:payment.no_subscription_new_payment')}</p>
-          <Button toModal="choose-offer" label={t('user:payment.view_plans')} />
+          <Button onClick={onClick} label={t('user:payment.view_plans')} />
```
ChristiaanScheermeijer: `onClick` shouldn't be needed, but it's fine for now:

```tsx
<Button to={modalURLFromLocation(location, 'choose-offer')} label={t('user:payment.view_plans')} />
```

mirovladimitrovski: The `onClick` handler is there for the `isLoading` state, because we don't redirect right away, but only after we get a redirection URL from an API request.

packages/hooks-react/src/useBillingPortalURL.tsx
mirovladimitrovski: Done. I did it a little bit differently, but I think you'll approve of it.

packages/ui-react/src/components/Subscription/Subscription.tsx
mirovladimitrovski: Sorry, my previous answer is not the correct one for this use case. You're actually right, I'm changing it right now.

mirovladimitrovski: Sorry, my previous answer to this is incorrect for this use case, I actually had a different instance of using `Button` in mind. You're actually right.

Done.

AntonLantukh:
@mirovladimitrovski could you periodically merge some changes from `develop` here?

PR #598, opened 2024-08-21
Feat / Injectable component wrappers

royschut:
## Injectable component wrappers

In #450, we introduced the `Injectable components` feature. This PR builds onto this and prepares the most common UI components. It will now be possible to implement a custom Card, Button, etc from the top-level (using Inversify), without having to replace the whole component tree. This should make it much easier to customize while keeping the source code sync, or build new platforms based on the same UI.

The process of implementing a custom component is described in the code comments of `web/src/modules/register`.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/platforms/web/src/modules/register.ts
+++ b/platforms/web/src/modules/register.ts
@@ -52,17 +52,18 @@ container.bind<LogTransporter>(LogTransporter).toDynamicValue(() => new ConsoleT
  *
  * @example
  * ```ts
- * // Define in types.ts (from `ui-react` for example)
- * const TAG_IDENTIFIER = 'TAG_IDENTIFIER';
+ * // Define an identifier from the component (from `ui-react` for example):
+ * export const CardIdentifier = Symbol('CARD');
  *
- * // Bind custom component
- * container.bind(TAG_IDENTIFIER).toConstantValue(CustomTag);
+ * // Make sure the component is wrapped in the HOC:
+ * export default createInjectableComponent(CardIdentifier, Card);
  *
- * // Then wrap the component in a HOC, from the component file itself
- * export default createInjectableComponent(TAG_IDENTIFIER, DefaultTag);
+ * // Override the component into the associated container, from register.ts:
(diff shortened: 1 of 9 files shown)


langemike (approved):

royschut:
I see that some e2e tests are broken. Although they are not caused by this PR, I spent a moment to research them:
- `I can see payment types - Cleeng`: It seems that the Cleeng Adyen configuration is broken (config `ozylzc5m`). The `/setup` call fails.
- `I can finish my subscription with credit card - JW Player`: It seems that the JWP configuration is broken as well (config `a2kbjdv0`). I can perform the payment, the access call succeeds, but the `/subscription` call returns a 500 server error:
`The API was not able to process this operation.`.

I suspect that the cancelation and invoices tests also fail because of this. @AntonLantukh or @dbudzins, perhaps you could check the configurations when you have a moment?

For now, I hope we can proceed with this PR.

platforms/web/src/modules/register.ts
```diff
@@ -52,17 +52,18 @@ container.bind<LogTransporter>(LogTransporter).toDynamicValue(() => new ConsoleT
...
+ * // Override the component into the associated container, from register.ts:
+ * import { container as uiComponentContainer } from '@jwp/ott-ui-react/src/modules/container';
  *
+ * uiComponentContainer.bind<React.FC<CardProps>>(CardIdentifier).toConstantValue(CustomCard);
  * ```
  */
 
 // Override ui-react component
-// uiComponentContainer.bind<React.FC<TagProps>>(TAG_IDENTIFIER).toConstantValue(Tag);
+// uiComponentContainer.bind<React.FC<CardProps>>(CardIdentifier).toConstantValue(CustomCard);
```
AntonLantukh, line 69: Looks like we already used this example right above.

AntonLantukh (approved):

PR #599, opened 2024-08-22
Feat / app metadata insertion

royschut:
## App metadata insertion

This PR makes it possible to add app metadata, to enable the typical [Smart app banner](https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners) (iOS) and [Native app banner](https://developer.chrome.com/blog/app-install-banners-native#prefer_related_applications) (Android), by adding the app bundle id through the env var.

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/platforms/web/scripts/build-tools/buildTools.ts
+++ b/platforms/web/scripts/build-tools/buildTools.ts
@@ -40,7 +40,7 @@ export const getFileCopyTargets = (mode: string): Target[] => {
   return fileCopyTargets;
 };
 
-export const createHeadMetaTags = (tagData: Record<string, string | undefined>): HtmlTagDescriptor[] => {
+export const getMetaTags = (tagData: Record<string, string | undefined>): HtmlTagDescriptor[] => {
   return Object.entries(tagData)
     .filter(([_, value]) => !!value)
     .map(([name, content]) => ({
@@ -98,7 +98,7 @@ export const getRelatedApplications = ({
   if (appleAppId) {
     relatedApplications.push({
       platform: 'itunes',
-      url: `https://apps.apple.com/nl/app/${appleAppId}`,
+      url: `https://apps.apple.com/app/${appleAppId}`,
     });
   }
 
--- a/platforms/web/scripts/build-tools/buildTools.ts
+++ b/platforms/web/scripts/build-tools/buildTools.ts
@@ -39,19 +39,14 @@ export const getFileCopyTargets = (mode: string): Target[] => {
   return fileCopyTargets;
 };
 
-export const getGoogleVerificationTag = (env: Record<string, string>): HtmlTagDescriptor[] => {
-  if (!env.APP_GOOGLE_SITE_VERIFICATION_ID) return [];
-
-  return [
-    {
+export const createHeadMetaTags = (tagData: Record<string, string | undefined>): HtmlTagDescriptor[] => {
+  return Object.entries(tagData)
+    .filter(([_, value]) => !!value)
+    .map(([name, content]) => ({
       tag: 'meta',
       injectTo: 'head',
-      attrs: {
-        content: process.env.APP_GOOGLE_SITE_VERIFICATION_ID,
-        name: 'google-site-verification',
-      },
-    },
-  ];
+      attrs: { name, content },
+    }));
 };
 
 // @todo: move to common?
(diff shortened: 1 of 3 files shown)


platforms/web/vite.config.ts
```diff
@@ -72,8 +80,8 @@ export default ({ mode, command }: ConfigEnv): UserConfigExport => {
           theme_color: '#DD0000',
           orientation: 'any',
           background_color: '#000',
-          related_applications: [],
-          prefer_related_applications: false,
+          related_applications: env.APP_GOOGLE_RELATED_APPLICATION_ID ? [{ platform: 'play', id: env.APP_GOOGLE_RELATED_APPLICATION_ID, url: '' }] : [],
```
langemike: I believe we can discard the (empty) URL.

According to the Google doc you mentioned the `url` property is not a requirement. Also in some other online examples, I don't see it. If we still want to provide the URL parameter maybe we can fill it with `https://play.google.com/store/apps/details?id=${env.APP_GOOGLE_RELATED_APPLICATION_ID}` an empty value is a bit weird.

Also, I find it a little bit more neat to also provide the itunes platform within `related_applications`, although this is not a requirement for the smart banner. It makes it more "complete".

To improve the readability I will also move this up in a `const`. You could move it to the buildTools.ts too but I leave this last decision up to you.

royschut: I only added the empty string `url` because the typings of VitePWA expect it to be set.

Great idea to also add `itunes`, but that one does require a url, which cannot really be generated based on the bundle id. So that would imply another env var, basically for no practical purpose. What do you think?

Do agree about the `const`.

langemike: > I only added the empty string url because the typings of VitePWA expect it to be set.

Then we could use my alternative approach with `https://play.google.com/store/apps/details?id=${env.APP_GOOGLE_RELATED_APPLICATION_ID}`

> Great idea to also add iTunes

Not important enough for now, then.

royschut: Agreed, added it.
Also, I was wrong about the itunesId: not the bundle id, but the apple id is required, which I can now also use in the related_application. Please see my latest commit. @langemike

platforms/web/vite.config.ts
```diff
@@ -42,12 +43,22 @@ export default ({ mode, command }: ConfigEnv): UserConfigExport => {
...
   const bodyFontsString = bodyFonts.map((font) => font.fontFamily).join(', ');
   const bodyAltFontsString = bodyAltFonts.map((font) => font.fontFamily).join(', ');
+
+  // Head tags
+  const fontTags = getGoogleFontTags([bodyFonts, bodyAltFonts].flat());
+  const metaTags = createHeadMetaTags({
+    'apple-itunes-app': env.APP_APPLE_ITUNES_APP ? `app-id=${env.APP_APPLE_ITUNES_APP}` : undefined,
+    'google-site-verification': env.APP_GOOGLE_SITE_VERIFICATION_ID,
+  });
+  const tags = [fontTags, metaTags, getGtmTags(env)].flat();
```
AntonLantukh: Should it be a `getTags` function?

platforms/web/scripts/build-tools/buildTools.ts
```diff
@@ -90,6 +86,33 @@ export const generateIconTags = (basePath: string, favIconSizes: number[], apple
...
+}: {
+  appleAppId?: string | undefined;
+  googleAppId?: string | undefined;
+} = {}): ManifestOptions['related_applications'] => {
+  const relatedApplications = [];
+
+  if (appleAppId) {
+    relatedApplications.push({
+      platform: 'itunes',
+      url: `https://apps.apple.com/nl/app/${appleAppId}`,
```
AntonLantukh: Can we avoid using `nl` in the url?

platforms/web/vite.config.ts
royschut: More correct indeed, but I'd say `getMetaTags` instead, to differentiate from tags such as the Google fonts.

platforms/web/scripts/build-tools/buildTools.ts
royschut: Apologees, the locale was not supposed to get it in there. 😅

platforms/web/vite.config.ts
langemike: Approved

langemike (approved):

AntonLantukh (approved):

PR #601, opened 2024-08-26
[IDM-232] - feat: stripe billing portal session

kmitrovv:
## Description

This pull request implements the backend logic for creating a billing portal session within the Access Bridge component. It establishes the necessary interactions with the Sims and Stripe services to validate the viewer and create a checkout session.

**Endpoint**:
`POST:  /v2/billing-portal`

Jira card: [Stripe Billing Portal URL](https://jwplayer.atlassian.net/browse/IDM-232)

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code


platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -63,4 +63,62 @@ export class CheckoutController {
...
+        sendErrors(res, ErrorDefinitions.UnauthorizedError.create());
+        return;
+      }
+
+      // Get the email address from the Authorization token
+      const viewer = await this.identityService.getAccount({ authorization });
+      if (!viewer.id || !viewer.email) {
+        sendErrors(res, ErrorDefinitions.UnauthorizedError.create());
+        return;
+      }
```
budzworthy: We are doing this exact same snippet in 3 places now. It seems to be pretty central to the main flows of the auth bridge. Would it make sense to put it in an Express middleware so that the downstream methods can just rely on the request context having a viewer property populated?

platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -63,4 +63,62 @@ export class CheckoutController {
...
+      const customerId = await this.stripeService.getCustomerIdByEmail({ email: viewer.email });
+      if (!customerId) {
+        sendErrors(
+          res,
+          ErrorDefinitions.NotFoundError.create({ description: 'The requested customer does not exist in Stripe.' })
+        );
+        return;
+      }
+
+      const { return_url } = req.body;
```
budzworthy: Do we need to do some validation on this url to make sure it's at least part of the expected domain?

platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -63,4 +63,62 @@ export class CheckoutController {
...
+
+      // Get the email address from the Authorization token
+      const viewer = await this.identityService.getAccount({ authorization });
+      if (!viewer.id || !viewer.email) {
+        sendErrors(res, ErrorDefinitions.UnauthorizedError.create());
+        return;
+      }
+
+      // Retrieve Stripe customer ID using the email
+      const customerId = await this.stripeService.getCustomerIdByEmail({ email: viewer.email });
```
budzworthy: Can you do the customer ID validation within the stripe service instead of passing it here and then passing it back again?

platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -63,4 +63,62 @@ export class CheckoutController {
...
+      }
+
+      const { return_url } = req.body;
+      if (!return_url) {
+        sendErrors(res, ErrorDefinitions.ParameterMissingError.create({ parameterName: 'return_url' }));
+        return;
+      }
+
+      // Generate a billing portal session
+      const session = await this.stripeService.createBillingPortalSession({ customerId, returnUrl: return_url });
```
budzworthy: Would it work to just return a redirect directly to the checkout session here?

platforms/access-bridge/src/controllers/checkout-controller.ts
```diff
@@ -63,4 +63,62 @@ export class CheckoutController {
...
+        sendErrors(res, ErrorDefinitions.UnauthorizedError.create());
+        return;
+      }
+
+      // Retrieve Stripe customer ID using the email
+      const customerId = await this.stripeService.getCustomerIdByEmail({ email: viewer.email });
+      if (!customerId) {
+        sendErrors(
+          res,
+          ErrorDefinitions.NotFoundError.create({ description: 'The requested customer does not exist in Stripe.' })
```
budzworthy: I'm confused. How will this work for new purchases? Is this PR only intended for the flow where a customer wants to view / update an existing subscription? If that's the case, maybe some method name adjustments and/or comments would make it more clear. Something like `subscriptionManagementPortal`?

platforms/access-bridge/src/services/stripe-service.ts
```diff
@@ -21,12 +21,28 @@ export class StripeService {
...
+   * @throws Error if there is an issue searching stripe customers by email.
+   */
+  async getCustomerIdByEmail({ email }: { email: string }): Promise<string | null> {
+    // Search for customers by email using Stripe's API.
+    const customers = await this.stripe.customers.search({
+      query: `email:'${email}'`,
+    });
+
+    // Return the first customer's ID if available, otherwise null.
+    return customers.data.length ? customers.data[0].id : null;
```
budzworthy: What does it mean if multiple are returned?

platforms/access-bridge/src/controllers/checkout-controller.ts
kmitrovv: I don't think anyone using this implementation will have hardcoded domain.. Maybe someone wants to redirect them to a certain landing page, so I don't like to put such a limitation.

platforms/access-bridge/src/controllers/checkout-controller.ts
kmitrovv: Added a comment explaining what it does. I would leave the naming as that's what stripe calls it. We can adjust new name when we add more integrations if it makes more sense.

platforms/access-bridge/src/services/stripe-service.ts
kmitrovv: I've updated the logic a little. I don;'t think it's possible to have more than 1 customers if we search by email, as we do. This is Stripe's search API and probably works with few characters only, instead of entire email so I assume that's why they have it structured like that.

platforms/access-bridge/src/controllers/checkout-controller.ts
kmitrovv: Agreed. The site id is moved to middleware. Not sure about the authorization as it's only mandatory here in the checkout controller. If makes sense later we can move it to a middleware as well.

darkoatanasovski (approved):

PR #602, opened 2024-08-28
[OWA-79] feat(i18n): add translations

CarinaDraganJW:
## Description

- This PR adds custom translations in the web app 

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code


baukef:
Great work. Could you maybe describe the scope of this ticket a bit more?

CarinaDraganJW:
> Great work. Could you maybe describe the scope of this ticket a bit more?

@baukef Definitely! In the dashboard, we have the `Translations` section on media details page, where users can add custom translations for the content type fields marked as `translatable`. For each field filled in we save a custom parameter of the following type: `param-languageCode: translation introduced in the field`

However, we did not integrate this into the OTT Web App yet, so these custom params translations are not retrieved. What this ticket is trying to do is to retrieve these custom params and transform the media item: if we for instance have `title` and `title-es`, and language is set to `es` (Spanish), we will retrieve value for `title-es` and set `title` to this value.

baukef:
It would be awesome if you could extend this pattern to playlists as well. This is not yet available in the dashboard, but we could still use the custom parameters on playlists. The main goal would be for the navigation (library) items to be translated this way.

marcovandeveen:
@baukef agree. It should probably be on a shelf level

Shelves have titles as well. The same goes for menu items. 
Shelves now have custom params as well. So should be easy to implement the pattern there as well.

langemike:
This is heading in the right direction. I suggest using `APP_DEFAULT_LANGUAGE` for the language fallback instead of the hard-coded `'en'`

packages/hooks-react/src/useSelectedLanguage.ts
```diff
@@ -0,0 +1,15 @@
+import { useTranslation } from 'react-i18next';
+import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
+import { useMemo } from 'react';
+
+import env from '../../common/src/env';
+
+const useSelectedLanguage = (): string => {
```
AntonLantukh: Hm, I am not sure we need to do it here as we already have this logic supported in `initI18n` function. So looks like we could simply use `i18n.language`.

packages/hooks-react/src/usePlaylist.ts
```diff
@@ -10,6 +10,8 @@ import type { PlaylistMenuType } from '@jwp/ott-common/types/config';
 import { PLAYLIST_TYPE } from '@jwp/ott-common/src/constants';
 import { useConfigStore } from '@jwp/ott-common/src/stores/ConfigStore';
 
+import env from '../../common/src/env';
```
AntonLantukh:
```suggestion
import env from '@jwp/ott-common/src/env';
```

packages/ui-react/src/pages/ScreenRouting/MediaScreenRouter.tsx
```diff
@@ -35,7 +36,11 @@ mediaScreenMap.register(MediaSeries, (item) => !!item && isLegacySeriesFlow(item
 const MediaScreenRouter = () => {
   const params = useParams();
   const id = params.id || '';
-  const { isLoading, isFetching, error, data } = useMedia(id);
+
+  // Determine currently selected language
+  const language = useSelectedLanguage();
```
AntonLantukh: Do you think we could get it right in `useMedia`?

packages/hooks-react/src/series/useNextEpisode.ts
```diff
@@ -3,14 +3,19 @@ import type { Series } from '@jwp/ott-common/types/series';
...
+  // Determine currently selected language
+  const { i18n } = useTranslation('menu');
+  const language = i18n.language;
+
   const { isLoading, data } = useQuery(
-    ['next-episode', series?.series_id, episodeId],
+    ['next-episode', series?.series_id, episodeId, language],
     async () => {
-      const item = await apiService.getEpisodes({ seriesId: series?.series_id, pageLimit: 1, afterId: episodeId });
+      const item = await apiService.getEpisodes({ seriesId: series?.series_id, pageLimit: 1, afterId: episodeId, language });
```
ChristiaanScheermeijer: Is the `'menu'` namespace given to `useTranslation` accidental or on purpose? I think we can remove it.

Alternatively, saving some lines, you can do this as well:

```tsx
import { getI18n } from 'react-i18next';

const item = await apiService.getEpisodes({ seriesId: series?.series_id, pageLimit: 1, afterId: episodeId, language: getI18n().language });
```

packages/hooks-react/src/usePlaylist.ts
```diff
@@ -62,17 +65,27 @@ export const getPlaylistQueryOptions = ({
...
+  enabled = true,
+  usePlaceholderData = true,
+  type = PLAYLIST_TYPE.playlist,
+}: {
+  contentId?: string;
+  params?: GetPlaylistParams;
+  enabled?: boolean;
+  usePlaceholderData?: boolean;
+  type?: PlaylistMenuType;
+}) {
```
ChristiaanScheermeijer: Should we revert this function signature change as well?

packages/hooks-react/src/useProtectedMedia.ts
```diff
@@ -2,15 +2,22 @@ import { useQuery } from 'react-query';
...
+import { useTranslation } from 'react-i18next';
 
 import useContentProtection from './useContentProtection';
 
 export default function useProtectedMedia(item: PlaylistItem) {
   const apiService = getModule(ApiService);
-  const contentProtectionQuery = useContentProtection('media', item.mediaid, (token, drmPolicyId) => apiService.getMediaById(item.mediaid, token, drmPolicyId));
+  const contentProtectionQuery = useContentProtection('media', item.mediaid, (token, drmPolicyId) =>
+    apiService.getMediaById(item.mediaid, language, token, drmPolicyId),
+  );
```
ChristiaanScheermeijer: Isn't the `language` constant being used before definition? It probably works because it's a callback...

packages/hooks-react/src/useMedia.ts
```diff
@@ -3,13 +3,18 @@ import type { PlaylistItem } from '@jwp/ott-common/types/playlist';
 import ApiService from '@jwp/ott-common/src/services/ApiService';
 import { getModule } from '@jwp/ott-common/src/modules/container';
 import { isScheduledOrLiveMedia } from '@jwp/ott-common/src/utils/liveEvent';
+import { useTranslation } from 'react-i18next';
 
 export type UseMediaResult<TData = PlaylistItem, TError = unknown> = UseBaseQueryResult<TData, TError>;
 
-export default function useMedia(mediaId: string, enabled: boolean = true): UseMediaResult {
+export default function useMedia({ mediaId, enabled = true }: { mediaId: string; enabled?: boolean }): UseMediaResult {
```
ChristiaanScheermeijer: Revert this function signature as well?

packages/hooks-react/src/useProtectedMedia.ts
```diff
@@ -2,15 +2,22 @@ import { useQuery } from 'react-query';
...
+
+  // Determine currently selected language
+  const { i18n } = useTranslation('menu');
+  const language = i18n.language;
 
   const { isLoading, data: isGeoBlocked } = useQuery(
-    ['media', 'geo', item.mediaid],
+    ['media', 'geo', item.mediaid, language],
```
ChristiaanScheermeijer: I don't think language is needed here. The GEO country doesn't change, so we don't want to revalidate GEO blocking...

packages/hooks-react/src/useProtectedMedia.ts
ChristiaanScheermeijer: The `useContentProtection` hook contains the `useQuery` for the media item. The language should be passed to that hook, so it can revalidate when the `language` changes.

Although... the media item returned there is only used for media playback, so perhaps we don't need to translate it at all.

packages/hooks-react/src/series/useNextEpisode.ts
ChristiaanScheermeijer: Forget the getI18n suggestion. That doesn't make it reactive to language changes

AntonLantukh:
@ChristiaanScheermeijer what if we make the language change reload the page? That way we could simply import i18n once in the ApiService and use the language property. So basically everything using "transformMedia" function will get translations.

CarinaDraganJW:
> @ChristiaanScheermeijer what if we make the language change reload the page? That way we could simply import i18n once in the ApiService and use the language property. So basically everything using "transformMedia" function will get translations.

@ChristiaanScheermeijer We can do the reload implementation very easily:

- In `HeaderLanguageSwitcher` when we set the language, we can do `window.location.reload();`
- In `ApiService.ts`, in the newly added function `getTranslatedFields`, we can just get the language with `i18next.language`

Screenshot 2024-09-04 at 12 58 21
Screenshot 2024-09-04 at 12 58 30



This is tested and it works, this way we could easily remove all the passed down language props 🙂
Let me and Anton know what you think, if you consider it would be ok to reload on language switch, I can commit the changes 🚀
Thank you ❤️

ChristiaanScheermeijer:
@CarinaDraganJW @AntonLantukh Hmm... I'm not a big fan of reloading the page actually and I'm afraid it causes issues on other platforms 😬 

Passing the language is fine, I was trying to make the language getter a bit more simple, but it doesn't really bother me how it is now. 

Only a bit of cleanup and it should be good to go!

If we really .. really want to make it cleaner, we could reintroduce the `getLanguage` hook that uses `useTranslation` and returns `i18n.langage`...

AntonLantukh:
@ChristiaanScheermeijer yeah, this is why I wanted to ask, was not sure how it could behave on different platforms. Using it in the hook is fine, it just requires you to support it for every new hook which has something to do with media item transformation. We will also look into how we could support it for favorites and continue_watching which we keep in store.

CarinaDraganJW:
@AntonLantukh @ChristiaanScheermeijer Code is ready and working, feel free to have another look 🚀

packages/hooks-react/src/useProtectedMedia.ts
```diff
@@ -7,7 +7,9 @@ import useContentProtection from './useContentProtection';
 
 export default function useProtectedMedia(item: PlaylistItem) {
   const apiService = getModule(ApiService);
-  const contentProtectionQuery = useContentProtection('media', item.mediaid, (token, drmPolicyId) => apiService.getMediaById(item.mediaid, token, drmPolicyId));
+  const contentProtectionQuery = useContentProtection('media', item.mediaid, (token, drmPolicyId) =>
+    apiService.getMediaById(item.mediaid, undefined, token, drmPolicyId),
```
AntonLantukh: Can we make the language the last argument of the function? That way the signature remains the same.

packages/hooks-react/src/usePlaylists.ts
```diff
@@ -25,6 +26,9 @@ const usePlaylists = (content: Content[], rowsToLoad: number | undefined = undef
   const favorites = useFavoritesStore((state) => state.getPlaylist());
   const watchHistory = useWatchHistoryStore((state) => state.getPlaylist());
 
+  // Determine currently selected language
+  const { i18n } = useTranslation('menu');
```
AntonLantukh: Do we need "menu" to retrieve it? Can we import "i18n" directly?

packages/hooks-react/src/useBootstrapApp.ts
```diff
@@ -19,11 +20,13 @@ export type OnReadyCallback = (config: Config | undefined) => void;
...
 export const useBootstrapApp = (url: string, onReady: OnReadyCallback) => {
   const queryClient = useQueryClient();
+  const { i18n } = useTranslation('menu');
+
   const refreshEntitlements = () => queryClient.invalidateQueries({ queryKey: ['entitlements'] });
 
   const { data, isLoading, error, isSuccess, refetch } = useQuery<Resources, Error | AppError>(
     'config-init',
-    () => applicationController.initializeApp(url, refreshEntitlements),
+    () => applicationController.initializeApp(url, i18n.language, refreshEntitlements),
```
AntonLantukh, line 29: Hm, I think that here we can simply import i18n library and get language from there. This one doesn't need to be dynamic.

packages/common/src/services/ApiService.ts
```diff
@@ -46,10 +46,29 @@ export default class ApiService {
     return date ? parseISO(date) : undefined;
   };
 
+  private getTranslatedFields = (item: PlaylistItem, language?: string) => {
```
AntonLantukh:
```suggestion
  protected getTranslatedFields = (item: PlaylistItem, language?: string) => {
```

packages/hooks-react/src/usePlaylists.ts
AntonLantukh: Can be resolved. Should be reactive

CarinaDraganJW: I used  `const i18n = getI18n();` and then `i18n.language`
Seems to still work 🙂

AntonLantukh: I think `useTranslation('menu')` is better here. The solution you had originally is fine.

AntonLantukh (approved):

CarinaDraganJW:
Hi @ChristiaanScheermeijer 👋🏻

The feature is ready, I fixed all the feedback Anton had and I added translations for all media on Blender  
Let me and Anton know if you have any more feedback on the code, otherwise I can proceed and merge the PR 🚀
Thank you for all the help ❤️

packages/hooks-react/src/series/useEpisodes.ts
```diff
@@ -28,22 +29,26 @@ export const useEpisodes = (
 } => {
   const apiService = getModule(ApiService);
 
+  // Determine currently selected language
+  const i18n = getI18n();
+  const language = i18n.language;
```
ChristiaanScheermeijer: This hook isn't reactive on itself when the `language` changes. It might work because the parent component re-renders, but this might be considered "unstable".

CarinaDraganJW: Good point, I can quickly change back to using `const { i18n } = useTranslation('menu');` instead 👍

ChristiaanScheermeijer: You can verify this by placing a `console.log` in the `Root` component:

```ts
const Root = () => {
  console.log('Current language', getI18n().language);
  // ...
}
```

This is only called once and doesn't get called when the language changes. This means that when using the `useSeries` or `useMedia` hook, it only updates because the component or parent is using the `useTranslation()` hook. But if you would use this hook in a different component without translations it will not update correctly.

CarinaDraganJW: I reverted to using `useTranslation` hook instead 👍 
Thank you for the quick check and in-depth explanation ❤️

PR #603, opened 2024-08-30
[IDM-172] - feat: generate passport

kmitrovv:
## Description

This PR introduces the use of the recently implemented [service](https://github.com/jwplayer/ott-web-app/pull/582) to generate a passport — a JWE token that encapsulates the list of plans the viewer is entitled to. The passport is now stored in local storage and will be utilized by the media metadata endpoint.

**Context**:
The passport serves as a critical trust token for seamless communication with various services, ensuring the viewer's entitlements are recognized. While it is currently used by the media metadata endpoint, future enhancements will expand its usage across additional services, enabling more secure and efficient data retrieval.

In subsequent PRs, we will implement mechanisms to refresh the passport and leverage it further for media playback and other related functionalities.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


platforms/web/.env
```diff
@@ -1,4 +1,4 @@
-APP_API_BASE_URL=https://cdn.jwplayer.com
+APP_API_BASE_URL=https://cdn-dev.jwplayer.com
```
budzworthy: Prob shouldn't commit this change to the main .env file. You can use .env.local or .env.dev.

packages/common/types/passport.ts
```diff
@@ -0,0 +1,4 @@
+export type Passport = {
```
budzworthy: Is this shared with the access bridge codebase?

kmitrovv: Yes, but made this branch from Develop so it's missing here, but later they will be merged. It's in the same place.

platforms/web/.env
kmitrovv: Agreed. This will be removed when ready. This PR should remain draft, we need this to look into dev, as the middleware at the moment communicates only with dev. We don't have all the services ready in other envs.

packages/common/src/controllers/AppController.ts
```diff
@@ -94,6 +96,11 @@ export default class AppController {
       await getModule(FavoritesController).initialize();
     }
 
+    // useAccessBridge is enabled when the APP_API_ACCESS_BRIDGE_URL environment variable is configured
+    if (useAccessBridge) {
```
kmitrovv: Was wondering if instead of doing this in multiple places, to conditionally bind two controllers (create one dummy with mocked methods that do nothing). Wdyt @dbudzins ?

packages/common/src/controllers/AccessController.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    // If the APP_API_ACCESS_BRIDGE_URL environment variable is defined, useAccessBridge will return true
+    // For the AVOD access model, signing and DRM are not supported, so passport generation is skipped
+    if (!this.apiAccessBridgeUrl || accessModel === 'AVOD') {
+      return;
+    }
+
+    // Not awaiting to avoid blocking the loading process,
+    // as the passport can be stored asynchronously without affecting the app's performance
+    this.generateOrRetrievePassport();
```
budzworthy: Adding void before the method call is the way you can make it explicit that you don't care about awaiting this async method: https://typescript-eslint.io/rules/no-floating-promises/

```suggestion
    void this.generateOrRetrievePassport();
```

packages/common/src/env.ts
```diff
@@ -1,6 +1,7 @@
 export type Env = {
   APP_VERSION: string;
   APP_API_BASE_URL: string;
+  APP_API_ACCESS_BRIDGE_URL: string;
```
budzworthy, line 4: As we discussed the other day, the middleware url will be different per customer, so should probably be an ini file param

packages/common/src/modules/register.ts
```diff
@@ -51,6 +53,7 @@ container.bind(FavoriteService).toSelf();
 container.bind(GenericEntitlementService).toSelf();
 container.bind(ApiService).toSelf();
 container.bind(SettingsService).toSelf();
+container.bind(AccessService).toSelf();
```
budzworthy, line 59: I think it might be more intuitive to only bind AccessService and Controller if the middleware url is present, instead of checking the middleware url in each method call.

packages/common/src/services/AccessService.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    if (!response.ok) {
+      return null;
+    }
+
+    return (await response.json()) as Passport;
+  };
+
+  setPassport = (passport: Passport) => {
```
budzworthy: Was there a reason you put the storage (set/get/remove) in the access service instead of in the access controller? It seems like it might be cleaner to have a controller that calls 2 services instead of a controller that calls a service that calls a service. Thoughts?

packages/common/types/passport.ts
```diff
@@ -0,0 +1,4 @@
...
+  passport: string;
```
budzworthy: It's a little confusing to have a type `Passport` with a property `passport`. Can we rename this something like `AccessTokens`?

platforms/web/ini/.webapp.dev.ini
```diff
@@ -2,3 +2,5 @@
 defaultConfigSource = gnnuzabk
 ; When developing, switching between configs is useful for test and debug
 UNSAFE_allowAnyConfigSource = true
+; Access Bridge service API url host
+apiAccessBridgeUrl=https://access-bridge-57322003213.europe-west1.run.app
```
budzworthy: I'm not sure we should hard-code a temporary POC deployment, even in the dev config. Anyone who forks the open source repo would start sending traffic there and we don't really want to encourage that.

packages/common/src/env.ts
budzworthy: nm. I see it's in both places.

packages/common/src/services/AccessService.ts
```diff
@@ -0,0 +1,51 @@
...
+    const expires = new Date(Date.now() + 3600 * 1000).getTime(); // Set expiration time to one hour from now
+    return this.storageService.setItem(PASSPORT_KEY, JSON.stringify({ ...passport, expires }), true);
```
budzworthy: Suggest using all async-await or none (`set` is currently different from `get` and `remove`)

packages/common/src/controllers/AccessController.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (existingPassport) {
+      return existingPassport;
+    }
+
+    const auth = await this.accountService.getAuthData();
+    const newPassport = await this.accessService.generatePassport(this.apiAccessBridgeUrl, this.siteId, auth?.jwt);
+    if (newPassport) {
+      await this.accessService.setPassport(newPassport);
+    }
+  };
```
budzworthy, line 57: Are you missing a `return newPassport` here?

packages/common/src/controllers/AccessController.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  generateOrRetrievePassport = async () => {
+    if (!this.apiAccessBridgeUrl) {
+      return;
+    }
+
+    const existingPassport = await this.accessService.getPassport();
+    if (existingPassport) {
+      return existingPassport;
```
budzworthy: Does this need to return the passport and refresh token or just the passport?

packages/common/src/services/AccessService.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@injectable()
+export default class AccessService {
+  private readonly storageService: StorageService;
+
+  constructor(@inject(StorageService) storageService: StorageService) {
+    this.storageService = storageService;
+  }
+
+  generatePassport = async (host: string, siteId: string, jwt?: string): Promise<Passport | null> => {
```
budzworthy: Can `host` be a private field in AccessService instead of being passed in? Seems like it can be encapsulated here and should be fairly much static.

packages/common/src/services/AccessService.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (!siteId) {
+      throw new Error('Site ID is required');
+    }
+
+    const pathname = `/v2/sites/${siteId}/access/generate`;
+    const url = `${host}${pathname}`;
```
budzworthy:
```suggestion
    const url = `${host}/v2/sites/${siteId}/access/generate`;
```

packages/common/src/services/AccessService.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const response = await fetch(url, {
+      method: 'PUT',
+      headers: {
+        Authorization: jwt ? `Bearer ${jwt}` : '',
+      },
+    });
+
+    if (!response.ok) {
+      return null;
```
budzworthy, line 30: Should we log something or even throw an exception in this case?

packages/common/src/controllers/AppController.ts
```diff
@@ -81,9 +82,10 @@ export default class AppController {
     // update settings in the config store
     useConfigStore.setState({ settings });
 
-    // when an integration is set, we initialize the AccountController
+    // when an integration is set, we initialize AccountController and AccessController
```
budzworthy:
```suggestion
    // when an integration is set, we initialize the AccountController and the AccessController
```

packages/common/src/services/integrations/jwp/JWPAccountService.ts
```diff
@@ -69,8 +69,9 @@ export default class JWPAccountService extends AccountService {
   assetId: number | null = null;
   svodOfferIds: string[] = [];
   sandbox = false;
+  siteId = '';
```
budzworthy: What is this used for?

packages/common/types/account.ts
```diff
@@ -333,3 +334,4 @@ export type GetWatchHistory = PromiseRequest<GetWatchHistoryArgs, SerializedWatc
 export type GetFavorites = PromiseRequest<GetFavoritesArgs, SerializedFavorite[]>;
 export type ExportAccountData = PromiseRequest<undefined, CommonAccountResponse>;
 export type DeleteAccount = PromiseRequest<DeleteAccountPayload, CommonAccountResponse>;
+export type GeneratePassport = () => Promise<Passport>;
```
budzworthy: I think this is unused now?

packages/ui-react/src/containers/AccountModal/forms/Login.tsx
```diff
@@ -34,8 +36,8 @@ const Login = () => {
     }),
     onSubmit: ({ email, password }) => accountController.login(email, password, window.location.href),
     onSubmitSuccess: () => {
+      accessController.generatePassport();
```
budzworthy: I think this would be better placed in the accountController login method. Logic in the components should be limited to UI flows.

packages/common/src/controllers/AppController.ts
```diff
@@ -81,9 +82,10 @@ export default class AppController {
...
     if (integrationType) {
       await getModule(AccountController).initialize(url, refreshEntitlements);
+      await getModule(AccessController).initialize();
```
budzworthy, line 92: Accounting for if access controller is not used:
```suggestion
      await getModule(AccessController)?.initialize();
```

packages/common/src/modules/register.ts
kmitrovv: That's something that has been bothering me as well, but since we moved the url in the .ini, we don't have that info on this point where we bind the services/controllers as that's a part of the Settings service.

packages/common/src/services/AccessService.ts
kmitrovv: Ah good point, will move these.

packages/common/types/passport.ts
kmitrovv: Yeah, probably should revert it as it originally was. AccessTokens.

packages/common/src/controllers/AccessController.ts
kmitrovv: no, if new passport is generated we just set it in the storage. This is later updated to generateOrRefresh passport. We don't actually need to return here anything.

packages/common/src/controllers/AccessController.ts
kmitrovv: this does not need to return anything. we just make sure the passport is set in local storage. however this is later updated in the next PR to generate or refresh.

packages/common/src/services/AccessService.ts
kmitrovv: That would mean to use the config store in the service, as the host is available in the settings of the config. Are we okay with that?

packages/common/src/services/integrations/jwp/JWPAccountService.ts
kmitrovv: this shouldnt be here. thanks.

packages/ui-react/src/containers/AccountModal/forms/Login.tsx
kmitrovv: But aren't we against using controllers inside other controllers? In that case I will need to use the access service here and make sure the needed params are provided.

PR #608, opened 2024-09-10
Feat / Maintenance 2024 q3

royschut:
## Maintenance 2024-Q3

At Videodock, we conduct quarterly maintenance to address any security risks and proactively upgrade the dependencies. This PR accomplishes that, with notable details outlined below.

### Prepare
- [x] Pull `main`, run `yarn`
- [x] Test dev server and release build

### Update dependencies
- [x] Run `yarn upgrade` and test dev server
- [x] Fix critical issues (must)
- [x] Fix high risk issues (should)
- [ ] Fix moderate risk issues
- [x] Run `$ npx syncyarnlock -skgl` to sync the `yarn.lock` file to the `package.json`
- [x] Run linting, unit tests, e2e tests
- [x] Test dev server and release build
- [x] Test dev server and release build

### Major updates
- `date-fns`: No refactor needed
- `dompurify`: No refactor needed
- `zustand`: Some import changes, and a fix for a 'set' error we made in `AppController`. Also: some deprecration warnings are introduced for the use of `shallow`. I suggest we fix this when Zustand 5 is here, which is already in 'rc'.

### Notes
- Only 2 moderate issues left, but both are sub-dependencies of the Inplayer sdk, which is currently being removed from the web app
- Removed obsolete resolutions
- Added 3 resolutions to fix high risk issues, one of which (`ws`) can be removed once the Inplayer sdk has been removed


langemike (approved):
Looks good! Thanks for the detailed PR description.

AntonLantukh (approved):

packages/common/src/stores/utils.ts
```diff
@@ -1,5 +1,5 @@
 import type { State, StateCreator } from 'zustand';
-import create from 'zustand';
+import { create } from 'zustand';
```
ChristiaanScheermeijer, line 2: The `State` type is deprecated, v5 is coming soon so let's remove it.

```diff
- export const createStore = <T extends State>
+ export const createStore = <T>

ChristiaanScheermeijer (commented):
Changes look good! I have 1 inline comment and one more here:

Running `npx syncpack lint` produces an error. Syncpack is not mandatory, but keeps the package.json files clean and sync between packages/platforms.

The failure is a false-negative because we don't define the local packages as peer dependency anymore. So we can remove this rule in the `.syncpackrc.json` file :-)

packages/common/src/stores/utils.ts
royschut: True, thanks

royschut:
> The failure is a false-negative because we don't define the local packages as peer dependency anymore. So we can remove this rule in the `.syncpackrc.json` file :-)

Thanks, updated!

ChristiaanScheermeijer (approved):

PR #609, opened 2024-09-11
Fix / demo config reset

royschut:
## Description

This PR fixes the "Unselect this config" button in demo mode, which was not responding. This happened because `useTrackConfigKeyChange` keeps undoing the change. Hard-setting the `window.location.href` fixes it, much like we do in all other config switch navigations. I have also removed some code that became obsolete.

Screenshot 2024-09-11 at 11 07 22

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/platforms/web/src/components/DemoConfigDialog/DemoConfigDialog.tsx
+++ b/platforms/web/src/components/DemoConfigDialog/DemoConfigDialog.tsx
@@ -1,6 +1,5 @@
 import React, { type ChangeEventHandler, type MouseEventHandler, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { type NavigateFunction, useNavigate } from 'react-router';
 import { Helmet } from 'react-helmet';
 import { createURL } from '@jwp/ott-common/src/utils/urlFormatting';
 import { CONFIG_QUERY_KEY } from '@jwp/ott-common/src/constants';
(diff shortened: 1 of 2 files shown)


platforms/web/src/components/DemoConfigDialog/DemoConfigDialog.tsx
```diff
@@ -170,7 +155,7 @@ const DemoConfigDialog = ({ query }: { query: BootstrapData }) => {
             helpLink={'https://docs.jwplayer.com/platform/docs/ott-create-an-app-config'}
             error={typeof state.error === 'string' ? undefined : state.error}
           >
-            <form method={'GET'} target={'/'}>
+            <form method={'GET'} target={PATH_HOME}>
```
langemike: the `target` attribute can be discarded. In our case, it does nothing.
When used, it should contain other values as described here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#target

langemike (approved):
Pragmatic approach! LGTM!

One minor comment about the target-attribute

AntonLantukh (approved):

platforms/web/src/components/DemoConfigDialog/DemoConfigDialog.tsx
royschut: You're right, it should be `_self`, which is the default.

royschut: That also required a snapshot update.

PR #610, opened 2024-09-12
feat(menu): support media type for menu

CarinaDraganJW:
## Description

- Support `media` type for menu items 

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code


packages/common/types/config.ts
```diff
@@ -44,13 +44,13 @@ export type Drm = {
 };
 
 export type PlaylistType = keyof typeof PLAYLIST_TYPE;
-
-export type PlaylistMenuType = Extract<PlaylistType, 'playlist' | 'content_list'>;
+export type PlaylistMenuType = Extract<PlaylistType, 'playlist' | 'content_list' | 'media'>;
```
AntonLantukh: Well, it looks like it is not only about playlists now. I would suggest the following:

```
PLAYLIST_TYPE -> APP_CONTENT_TYPE
PlaylistType -> AppContentType
PlaylistMenuType -> AppMenuType
PlaylistShelfType -> AppShelfType
```

packages/common/src/utils/urlFormatting.ts
```diff
@@ -117,6 +117,10 @@ export const mediaURL = ({
   );
 };
 
+export const singleMediaURL = (id: string, title?: string) => {
```
AntonLantukh, line 123: @ChristiaanScheermeijer I think it makes sense to rewrite `mediaURL` function to accept id and title instead of the whole media item. Though I also see that it is used in many places, so I would prefer to do it in a separate PR.

packages/ui-react/src/containers/Layout/Layout.tsx
```diff
@@ -24,6 +25,17 @@ import HeaderUserMenu from '../HeaderUserMenu/HeaderUserMenu';
 
 import styles from './Layout.module.scss';
 
+const determinePath = (type: PlaylistMenuType | undefined, contentId: string) => {
```
AntonLantukh: Can we accept arguments as an object here?

packages/common/src/utils/urlFormatting.ts
ChristiaanScheermeijer: Sounds good!

ChristiaanScheermeijer: Alternatively, for now you could do this as well to prevent refactoring:

```tsx
export const mediaURL = ({
  media,
  id,
  title,
  playlistId,
  play = false,
  episodeId,
}: ParamsWithMedia | ParamsWithId) => {
  const mediaId = media ? media.mediaid : id;
  const mediaTitle = media ? media.title : title;

  // ...
}
```

packages/common/src/utils/urlFormatting.ts
```diff
@@ -125,6 +132,17 @@ export const contentListURL = (id: string, title?: string) => {
   return createPath(PATH_CONTENT_LIST, { id, title: title ? slugify(title) : undefined });
 };
 
+export const determinePath = ({ type, contentId }: { type: AppMenuType | undefined; contentId: string }) => {
+  switch (type) {
+    case APP_CONTENT_TYPE.content_list:
+      return contentListURL(contentId);
+    case APP_CONTENT_TYPE.media:
+      return singleMediaURL(contentId);
+    default:
```
AntonLantukh: We should probably check playlist explicitly, as it is still possible we get something extra in the future.

packages/common/src/utils/urlFormatting.ts
```diff
@@ -125,6 +132,19 @@ export const contentListURL = (id: string, title?: string) => {
...
+export const determinePath = ({ type, contentId }: { type: AppMenuType | undefined; contentId: string }) => {
+  switch (type) {
+    case APP_CONTENT_TYPE.content_list:
+      return contentListURL(contentId);
+    case APP_CONTENT_TYPE.media:
+      return singleMediaURL(contentId);
+    case APP_CONTENT_TYPE.playlist:
+      return playlistURL(contentId);
+    default:
+      return playlistURL(contentId);
```
AntonLantukh:
```suggestion
      return ''
```

AntonLantukh (approved):

packages/common/src/utils/urlFormatting.ts
CarinaDraganJW: I created a separate ticket: https://jwplayer.atlassian.net/browse/OWA-89 🚀
I don't know if Christiaan can see it in Jira, but we'll tackle this in a separate PR 🙂

packages/common/src/constants.ts
```diff
@@ -86,9 +86,10 @@ export const EPG_TYPE = {
   viewNexa: 'viewnexa',
 } as const;
 
-export const PLAYLIST_TYPE = {
+export const APP_CONTENT_TYPE = {
```
ChristiaanScheermeijer: Should we use the term "content type" here? It might be confusing with the `contentTypes` feature 🤔

AntonLantukh: Well technically the property is `contentType`. We could use something like `APP_CONFIG_CONTENT_TYPE` or `APP_ITEM_TYPE`.

CarinaDraganJW: Maybe we could combine the ideas and use `APP_CONFIG_ITEM_TYPE`? 🙂

AntonLantukh: Works for me as well.

ChristiaanScheermeijer (approved):
Nice! 🎉

PR #611, opened 2024-09-12
Fix / first episode switching per season

ChristiaanScheermeijer:
## Description

We found a bug in the series screen when switching seasons. This bug causes the "Start watching" button to re-render because the first episode is based on the season episodes. 

We believe that the first episode should always be the first episode in the season and not the currently selected season. The episode in a different season can still be selected.

I also wanted to fix the "loading" state which empties the screen when loading a different episode. We can improve this by either optimistic load the next episode or keep the previous data by loading. This refactor became pretty big so I abandoned this for now. I could decrease the load time by loading the episode and seriesLookup in parallel. The media id was the same, although I'm not sure if this is always the case. @AntonLantukh do you know this?


royschut (approved):
Nice!
I'm thinking that maybe all series-related custom hooks and useEffects could move to `useSeries()`, to simplify the screen(s). But not urgent.

PR #613, opened 2024-09-18
Fix / card grid rendering previous items

ChristiaanScheermeijer:
## Description

When switching playlists pages or using the playlist filters in a wrapped shelf, the cards would render the previous item briefly. This creates some noise in the loading state as multiple page changes occur stacked.

This PR addresses that and also includes some performance improvements. The main cause of the old item being shown was the `key` being set to the column index in the grid. Because this doesn't change when a new playlist is set as prop, the card doesn't re-render immediately and keeps the previous poster image.

### DIFF
--- a/packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
+++ b/packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
@@ -9,7 +9,7 @@ type Props<Item> = {
   columnCount: number;
   data: Item[];
   renderCell: (item: Item, tabIndex: number) => JSX.Element;
-  getCellKey: (item: Item, rowIndex: number, columnIndex: number) => string;
+  getCellKey: (item: Item) => string;
 };
 
 const scrollIntoViewThrottled = throttle(function (focusedElement: HTMLElement) {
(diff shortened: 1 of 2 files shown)


langemike (approved):
Nice optimalisations

packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
```diff
@@ -9,14 +9,15 @@ type Props<Item> = {
   columnCount: number;
   data: Item[];
   renderCell: (item: Item, tabIndex: number) => JSX.Element;
+  getCellKey: (item: Item, rowIndex: number, columnIndex: number) => string;
```
royschut: I think you're not using the `rowIndex` and `columnIndex`.
(But perhaps you can just use `item.mediaid`, see next comment)

packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
```diff
@@ -118,10 +121,10 @@ const LayoutGrid = <Item extends object>({ className, columnCount, data, renderC
             <div
               role="gridcell"
               id={`layout_grid_${rowIndex}-${columnIndex}`}
-              key={columnIndex}
+              key={getCellKey(item, rowIndex, columnIndex)}
```
royschut: Why don't you use the required `item.mediaid` here straight?

royschut (commented):
Nice one! Although I think you've introduced unused code.

packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
ChristiaanScheermeijer: The `LayoutGrid` is made to be flexible (with any data). I initially added the `Item extends { mediaid: string }` but later added the `getCellKey` callback, but I forgot to remove the extends part.

I added `rowIndex` and `columnIndex` just in case, but it's not used in the current implementation.

packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
ChristiaanScheermeijer: See https://github.com/jwplayer/ott-web-app/pull/613#discussion_r1766640006

packages/ui-react/src/components/LayoutGrid/LayoutGrid.tsx
royschut: Got it. Then let's just remove the `extends`. :-)

royschut (approved):

PR #614, opened 2024-09-18
Feat / screen animations

ChristiaanScheermeijer:
## Description

This PR makes the transitions to screens (home, playlist and media) more smooth by adding a fade transition. 

To prevent layout jumps while navigating the app, I also set the body `overflow-y` to `scroll` to make the viewport width stable. 

The shelf list on the homepage uses a staggered animation (each shelf fades in with a small delay). 

**Edit** I noticed that the screen fade in animation didn't work every time. After some debugging, I found this is caused by a browser optimisation causing it to batch the paint. By manually triggering a reflow, the animation does work every time. There are alternatives like the [Web Animation API](https://caniuse.com/web-animation), but this isn't supported on older devices (TV platforms).

https://github.com/user-attachments/assets/8782548e-add9-41b4-ad84-255e4de5330e

Let me know what you think! 😄

### DIFF
--- a/packages/ui-react/src/components/Header/Header.module.scss
+++ b/packages/ui-react/src/components/Header/Header.module.scss
@@ -27,7 +27,8 @@
   // Make header static
   //
   &.static {
-    position: static;
+    position: relative;
+    z-index: 1;
     width: 100%;
   }
 }
--- a/packages/ui-react/src/components/Animation/Animation.tsx
+++ b/packages/ui-react/src/components/Animation/Animation.tsx
@@ -15,6 +15,8 @@ type Props = {
 
 export type Status = 'opening' | 'open' | 'closing' | 'closed';
 
+const triggerReflow = (element: HTMLElement | null) => element?.scrollTop;
+
 const Animation: React.FC<Props> = ({
   className,
   createStyle,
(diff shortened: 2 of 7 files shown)


langemike (approved):
Nice! Good choice to use `<Fade>` for this purpose!

Somewhere in the future it would be nice to use the  [View Transition API](https://developer.chrome.com/docs/web-platform/view-transitions) to make a cool transition from a shelf to a video detail page. Only [supported in modern browsers](https://caniuse.com/view-transitions), though ;) This idea was on my idea proposal list.

packages/ui-react/src/components/Header/Header.module.scss
```diff
@@ -27,7 +27,8 @@
   // Make header static
   //
   &.static {
-    position: static;
+    position: relative;
+    z-index: 1;
```
royschut, line 31: Are you sure this is needed? Does the usermenu popover still work?

royschut (approved):
Really great subtle improvement.
I have a small doubt about the `z-index` (+ the comments on #613)

packages/ui-react/src/components/Header/Header.module.scss
ChristiaanScheermeijer: Using CSS `will-change` causes layering issues. Now that the page is wrapped with the Fade component, the background image and fade cover the header. 

I tested most overlays and modals which all still look as expected.

ChristiaanScheermeijer:
I reverted the media screen fade when the mediaid changes because this causes the player to re-render when loading the next video. Although, this also happens when the media isn't cached, it's more noticeable now.

I believe we can solve this when we move the player to a parent component. This also solves some Airplay issues that we're having. This is a big change, so not for now...

AntonLantukh (approved):

PR #615, opened 2024-09-19
[IDM-174] - feat: use passport

kmitrovv:
## Description

This PR adds logic to handle the usage of the generated access tokens. There is a new endpoint for getting a media metadata, that requires passport to unlock the protected contents. 

Note: Will not affect the current flow, as will be enabled only if Access Bridge is used -> this is enabled only when a .ini value for the accessBridgeApi is present.

This PR also includes the following two:
- [[IDM-172] - feat: generate passport](https://github.com/jwplayer/ott-web-app/pull/603)
- [[IDM-173] - feat: refresh passport](https://github.com/jwplayer/ott-web-app/pull/604)

Design document: [AC - Sims Integration](https://www.notion.so/jwplayer/AC-Sims-Integration-1a61b486d8d74567971358901f41a34b)

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code

[IDM-172]: https://jwplayer.atlassian.net/browse/IDM-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDM-173]: https://jwplayer.atlassian.net/browse/IDM-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


darkoatanasovski (approved):

packages/hooks-react/src/useContentProtection.ts
```diff
@@ -20,20 +20,28 @@ const useContentProtection = <T>(
...
 
   const { data: token, isLoading } = useQuery(
     ['token', type, id, params],
     async () => {
+      if (isAcessBridgeEnabled) {
+        // if access control is set up we return nothing
+        // the useProtectedMedia hook will take care of it
+        return;
+      }
+
```
budzworthy: This is a bit confusing, and I think you can switch it around. What if instead, we don't enable this query when isAccessBridgeEnabled === true and DO enable the next one if isAccessBridgeEnabled === true

packages/hooks-react/src/useContentProtection.ts
```diff
@@ -20,20 +20,28 @@ const useContentProtection = <T>(
   const genericEntitlementService = getModule(GenericEntitlementService);
   const jwpEntitlementService = getModule(JWPEntitlementService);
 
-  const { configId, signingConfig, contentProtection, jwp, urlSigning } = useConfigStore(({ config }) => ({
+  const { configId, signingConfig, contentProtection, jwp, urlSigning, isAcessBridgeEnabled } = useConfigStore(({ config, settings }) => ({
```
budzworthy: Looks like the config store property has a misspelling that has propagated. Please replace all instances of `AcessBridge` with `AccessBridge`

budzworthy (approved):
2 small points, no blockers

PR #617, opened 2024-09-19
Fix / prevent recommendations as feed

ChristiaanScheermeijer:
## Description

This small fix prevents the recommendations playlist ending up in the feed param (`?r=:feedId`). This can cause problems when navigating to a different screen that does consume the feed for fetching the playlist.

**Reproduction:**

- Go to: https://app-preview.jwplayer.com/p/DWEqMBXB/?app-config=mcqrzgxv
- Open a video
- Click around until you find a live event in related videos grid
- Click on the live event
- Observe an API error and empty related videos grid

### DIFF
--- a/packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
+++ b/packages/ui-react/src/pages/ScreenRouting/mediaScreens/MediaEvent/MediaEvent.tsx
@@ -55,7 +55,7 @@ const MediaEvent: ScreenComponent<PlaylistItem> = ({ data: media, isLoading }) =
 
   // Media
   const { isLoading: isTrailerLoading, data: trailerItem } = useMedia(media?.trailerId || '');
-  const { isLoading: isPlaylistLoading, data: playlist } = usePlaylist(playlistId || '');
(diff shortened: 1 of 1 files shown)


langemike (approved):
Well spotted! Tested and works ✅

royschut (approved):

AntonLantukh:
@ChristiaanScheermeijer I think `r` param is also used for the client analytics purposes to pass the source playlist (where the click came from). In this case the source will be the previous playlist and not the one with recommendations.

ChristiaanScheermeijer:
@AntonLantukh do you suggest that we should use the recommendations feed by default in the MediaEvent page instead of the `r=` feed?

ChristiaanScheermeijer:
@AntonLantukh ?

AntonLantukh:
@ChristiaanScheermeijer sorry for the delay, lots of things happening. 
I think we can use recommendations list on the MediaEvent page the way we already do for MediaMovie, while using `const feedId = useQueryParam('r');` for previous / next item playback + to pass it to the player.

ChristiaanScheermeijer:
@AntonLantukh I rebased and updated the PR!

PR #627, opened 2024-10-03
fix: access bridge flaky tests

kmitrovv:
## Description

This PR fixes the unit tests on the access bridge side that were causing the workflow to fail from time to time.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code

### DIFF
--- a/platforms/access-bridge/src/pipeline/middleware.ts
+++ b/platforms/access-bridge/src/pipeline/middleware.ts
@@ -59,6 +59,7 @@ export class Middleware {
     if (err instanceof Stripe.errors.StripeError) {
       const accessBridgeError = handleStripeError(err);
       sendErrors(res, accessBridgeError);
+      return;
     }
 
     if (isJWError(err)) {
@@ -69,6 +70,7 @@ export class Middleware {
 
     logger.error('Unexpected error:', err);
     sendErrors(res, ErrorDefinitions.InternalError.create());
+    return;
   };
 
   /**
@@ -77,6 +79,7 @@ export class Middleware {
    */
   notFoundErrorHandler = (req: Request, res: Response, next: NextFunction) => {
     sendErrors(res, ErrorDefinitions.NotFoundError.create());
+    return;
   };
 
   /**
--- a/platforms/access-bridge/src/pipeline/middleware.ts
+++ b/platforms/access-bridge/src/pipeline/middleware.ts
@@ -70,7 +70,6 @@ export class Middleware {
 
     logger.error('Unexpected error:', err);
     sendErrors(res, ErrorDefinitions.InternalError.create());
-    return;
   };
 
   /**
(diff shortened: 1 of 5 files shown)


darkoatanasovski (approved):

platforms/access-bridge/src/pipeline/middleware.ts
```diff
@@ -59,6 +59,7 @@ export class Middleware {
     if (err instanceof Stripe.errors.StripeError) {
       const accessBridgeError = handleStripeError(err);
       sendErrors(res, accessBridgeError);
+      return;
```
AntonLantukh, line 62: Do wee need these empty returns?

kmitrovv: Yes, this is a general node.js practice. Without the return, the function might continue executing, potentially causing unintended behavior or attempting to send multiple responses, which would result in errors like Error: Can't set headers after they are sent.

kmitrovv: This is pretty much the case with javascript and most of the languages 🙂 
```
function test() {
  // do something
  return; // This exits the `test` function
}

function useTest() {
  if (something) {
    test(); // Call the `test` function
    // If you want to prevent further logic from running, you'd return here.
  }
  
  // This "other logic" will execute if the return isn't present after calling `test()`
  // other logic
}
```

darkoatanasovski: @AntonLantukh Do you have any concerns, or is this just an informative question? If it's the latter, we should discuss it in our shared Slack channel, not here.

AntonLantukh: @kiremitrov123 my comment was mostly about further usage of `return` for `notFoundErrorHandler` and `globalErrorHandler` where the execution of the function ends with `return` and no other conditions are expected. The example with `err instanceof Stripe.errors.StripeError` condition makes sense to me.

If you set `return` for `notFoundErrorHandler` will the execution go further to `globalErrorHandler`? Should we check the error type in  `notFoundErrorHandler` and call `next` if this is not related to api?

https://expressjs.com/en/guide/error-handling.html#error-handling

AntonLantukh: @darkoatanasovski I looking into the problem we tried to solve here. Fix for tests looks good to me.

kmitrovv: @AntonLantukh the question was addressed exactly on top of the Stripe one and I assumed you asked for every one of them. For the last return and one on the `notFoundErrorHandler`, they can be removed as it doesn't make any difference.

AntonLantukh (approved):
Thanks for looking into i!

PR #628, opened 2024-10-23
Feat / Featured shelf cosmetic update

royschut:
## Feat / Featured shelf cosmetic update

This PR improves the look and feel of the featured shelf. Instead of using the TileSlider with just a larger Card, this new version leverages the reserved space much more, with a smooth animation between slides. On mobile devices, you can drag the text off screen to bring in the next. 

See the screenshots/screen caps of desktop and mobile below.

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [x] Translations added
- [x] UX tested
- [x] Reviewed own code

Screenshot 2024-10-25 at 10 24 44
Screenshot 2024-10-25 at 10 24 21

https://github.com/user-attachments/assets/b8ac0c1c-0cd4-4c5c-99cc-79a8f66a7554

https://github.com/user-attachments/assets/ef917b90-3b55-4af5-b580-cb5ce3aeb1a2

### DIFF
--- a/packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
+++ b/packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
@@ -1,0 +1,62 @@
+import type { PlaylistItem } from '@jwp/ott-common/types/playlist';
+import classNames from 'classnames';
+import type { CSSProperties } from 'react';
+import ChevronRight from '@jwp/ott-theme/assets/icons/chevron_right.svg?react';
+import { mediaURL } from '@jwp/ott-common/src/utils/urlFormatting';
+import { useNavigate } from 'react-router';
+import { useTranslation } from 'react-i18next';
+import { isContentType, testId } from '@jwp/ott-common/src/utils/common';
+import { MEDIA_CONTENT_TYPE } from '@jwp/ott-common/src/constants';
+
+import TruncatedText from '../TruncatedText/TruncatedText';
+import StartWatchingButton from '../../containers/StartWatchingButton/StartWatchingButton';
+import Button from '../Button/Button';
+import Icon from '../Icon/Icon';
+
+import styles from './FeaturedShelf.module.scss';
+
+const FeaturedMetadata = ({
+  item,
+  loading,
+  playlistId,
+  style,
+  hidden,
+}: {
+  item: PlaylistItem | null;
+  loading: boolean;
+  playlistId: string | undefined;
+  style?: CSSProperties;
+  hidden?: boolean;
+}) => {
+  const navigate = useNavigate();
+  const { t } = useTranslation('common');
+
+  if (!item) return null;
+
+  const isWatchable =
+    !isContentType(item, MEDIA_CONTENT_TYPE.hub) && !isContentType(item, MEDIA_CONTENT_TYPE.liveChannel) && !isContentType(item, MEDIA_CONTENT_TYPE.page);
+  const hasVideo = item.sources.find((source) => source.file.indexOf('.m3u8') > -1 || source.file.indexOf('.mp4') > -1);
+
+  return (
+    <div
+      className={styles.metadata}
+      data-testid={testId(`featured-metadata--${hidden ? 'hidden' : 'visible'}`)}
+      style={{ ...style, visibility: hidden ? 'hidden' : undefined }}
+      aria-hidden={hidden ? 'true' : undefined}
+    >
+      <h2 className={classNames(loading ? styles.loadingTitle : styles.title)}>{!loading && item?.title}</h2>
+      <TruncatedText text={item?.description} maximumLines={3} />
+      <div>
+        {hasVideo && isWatchable && <StartWatchingButton item={item} playUrl={mediaURL({ id: item.mediaid, title: item.title, playlistId, play: true })} />}
+        <Button
+          label={t('common:more_info')}
+          onClick={() => navigate(mediaURL({ id: item.mediaid, title: item.title, playlistId }))}
+          startIcon={<Icon icon={ChevronRight} />}
+          size="large"
+        />
+      </div>
+    </div>
+  );
+};
+
+export default FeaturedMetadata;
--- a/packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
+++ b/packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
@@ -1,0 +1,372 @@
+@use '@jwp/ott-ui-react/src/styles/mixins/responsive';
+@use '@jwp/ott-ui-react/src/styles/variables';
+@use '@jwp/ott-ui-react/src/styles/theme';
+@use '@jwp/ott-ui-react/src/styles/mixins/typography';
+
+$desktop-height: 56.25vw;
+$desktop-max-height: 700px;
+$desktop-min-height: 275px;
+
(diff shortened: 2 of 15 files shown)


langemike:
Really nice improvement! I detected one visual problem; when you use the pagination dots, they move. So i need to reposition my mouse cursor anytime I want to click the dots.

Other points to consider:
1. Curious: why have you chosen to do inline JS/CSS animations as opposed to regular CSS? I think the later makes it easier to make styling changes. I understand it to calculate a dot size, but maybe the rest could be done in regular CSS.
2. I think we should use `aria-disabled` instead of `disabled` on the slider buttons/paginations. So the buttons can still keep their focus, while the screen-reader pronounces it's state. I know this is not the first time we have discussions about this ;-)
3. I think we should wrap the pagination within` <nav>` and `<ul>`. So the screen-reader pronounces the amount of items and makes it easier to navigate.

platforms/web/public/locales/en/common.json
```diff
@@ -27,5 +28,6 @@
   "slide_next": "Next slide",
   "slide_previous": "Previous slide",
   "slide_right": "Slide right",
+  "slide_to": "Slide to page {{page}} of {{pages}}",
```
langemike: Minor; I would go with "Slide to item x of y" or "Go to item x of y". When we wrap it in an `<ul>`, the "x of y" part is not needed in my eyes.

I think we should only use "pages" if there are more items per slide.

Using "pages" as a metaphor, could give the assumption a page can consist of multiple items. Such as

royschut: I agree!

langemike (approved):
Nice work! I'm satisfied with the applied height. I think the pagination alignment can be improved, but this is something for in the future to iterate on. Good enough for now!

packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
```diff
@@ -0,0 +1,59 @@
...
+  playlistId: string | undefined;
+  style?: CSSProperties;
+  hidden?: boolean;
+}) => {
+  const navigate = useNavigate();
+  const { t } = useTranslation('common');
+
+  if (!item) return null;
+
+  const hasVideo = item.sources.find((source) => source.file.indexOf('.m3u8') > -1 || source.file.indexOf('.mp4') > -1);
```
ChristiaanScheermeijer, line 38: We might need a util for this and also check for the content type. Some non-playable content types may contain dummy streams in the source list.

packages/ui-react/src/hooks/useScrolledDown.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useRef, useState } from 'react';
+import useEventCallback from '@jwp/ott-hooks-react/src/useEventCallback';
+
+export const useScrolledDown = (scrollUpHeight = 300, scrollDownHeight = 30) => {
```
ChristiaanScheermeijer, line 4: Let's refactor this (later) to a `useScrollProgress(start: number, end: number)` so we can fade the background smooth with the progress. IMHO, this mechanism works only for TV apps (spatial navigation).

packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
```diff
@@ -0,0 +1,308 @@
...
+.shelf {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  align-items: flex-start;
+  width: 100%;
+  height: 70vh;
+  min-height: 350px;
+  margin: 0 0 24px 0;
+  padding: 100px 56px;
```
ChristiaanScheermeijer:
```suggestion
  padding: 100px calc(variables.$base-spacing * 3);
```

packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
```diff
@@ -0,0 +1,308 @@
...
+  position: fixed;
+  top: 0;
+  right: 0;
+  left: 0;
+  z-index: 0;
+  display: flex;
+  justify-content: flex-end;
+  align-items: flex-end;
+  width: 100%;
+  height: calc(70vh + 56px);
```
ChristiaanScheermeijer: Is this for the header?

```suggestion
  height: calc(70vh + variables.$header-height);
```

packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
```diff
@@ -0,0 +1,308 @@
...
+  background-color: var(--featured-shelf-background-color);
+
+  @include responsive.mobile-and-small-tablet() {
+    height: calc(67vh + 56px);
```
ChristiaanScheermeijer:
```suggestion
    height: calc(67vh + variables.$header-height);
```

packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
```diff
@@ -0,0 +1,308 @@
...
+}
+
+.background {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  width: 100vw;
+  height: calc(70vh + 56px);
```
ChristiaanScheermeijer:
```suggestion
  height: calc(70vh + variables.$header-height);
```

packages/ui-react/src/hooks/useScrolledDown.ts
royschut: Actually, I was doing that, but that triggered lots of rerenders. So I left it.

packages/ui-react/src/components/FeaturedShelf/FeaturedShelf.module.scss
royschut: Yes, thanks!

packages/ui-react/src/hooks/useScrolledDown.ts
ChristiaanScheermeijer: I think we can do this outside React state:

**Hook interface**
```tsx
useScrollProgress(start: number, end: number, onChange: (progress: number) => void)
```

**Implementation**
```tsx
useScrollProgress(50, 600, progress => {
  posterRef.current.style.opacity = 1 - progress;
});
```

packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
royschut: I think we only want `movie` and `episode`?

packages/ui-react/src/hooks/useScrolledDown.ts
royschut: Yes, that should work. I will create a ticket for it.

packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
royschut: I see that `movie` is not a CONTENT_TYPE, so I just added an opt-out for `page`, `livechannel` and `hub` for now.

royschut:
Hi @AntonLantukh and @dbudzins, I hope you find a moment to have a look as well. :-)

ChristiaanScheermeijer (approved):

PR #629, opened 2024-10-28
Feat / UI updates and bring back support for light background colors

ChristiaanScheermeijer:
## Description

This PR contains many UI improvements and (re)adds support for using light background colors.

**Generic UI Improvements**

- Fix header and page alignment (spacing/gutters) for all breakpoints
- Increase the shelf spacing
- Increase the gap between the cards from 8 to 16 in the slider and card grid
- Increase the title spacing for all pages
- Cleanup unused SCSS variables

**Video details**

- Align the video details to the bottom of the Hero
- Refactor the Hero to make it re-usable for the Video details and Hub pages
- Update the background image mask to cover more of the page and hide the visual radial gradient

**Header**

- Increase the header height from 56 to 72 pixels
- Update the appearance of the search bar input

**Light background/header colors**

The OTT Web App originally supported using lighter background/header colors. This was not working properly anymore and contained many bugs. It's not pretty, but it does work and uses the correct contrast colors again. The following changes are related to this:

- Extract the text-shadow into a CSS variable and disable it when using a light background
- Update the `body-color` CSS variable based on the calculated contrast color
- Remove many `variables.$white` usages which overwrites the `--body-color` (and add some where it was needed)
- Use the modal color for the Sidebar (menu) and Panels on the User page
- Update the EPG layout to be full width and use the default JW blue background color

**Update 29/10**

- Use markdown for video descriptions
- Fix inline video layout alignment issues (title and description is aligned to the bottom of the page)
- Use dynamic background for the EPG, but not when using a light background color

**Screenshots**

localhost_8080__app-config= (1)

localhost_8080__app-config=

localhost_8080_u_my-account_app-config=2basfsic (1)

localhost_8080_u_my-account_app-config=2basfsic (2)

localhost_8080_u_my-account_app-config=2basfsic (3)

**Screenshots light background color**

localhost_8080_ (3)

localhost_8080_ (4)

localhost_8080__app-config=2basfsic


packages/hooks-react/src/usePlanByEpg.ts
```diff
@@ -72,19 +70,17 @@ const usePlanByEpg = ({
...
 };
 
-// Theme configuration for the Planby EPG with only the colors set that are used
-// Fixed values are used because the default highlightColor and backgroundColor are only available in SCSS
-export const makeTheme = (primaryColor?: string | null, backgroundColor?: string | null) => ({
+export const makeTheme = () => ({
   primary: {
-    600: backgroundColor || '#141523',
-    900: backgroundColor || '#141523',
+    600: '#141523',
```
royschut: Don't we want a neutral (grayscale) colour by default here?

royschut (approved):
Nice cosmetic improvement and great clean up as well. 👍

packages/hooks-react/src/usePlanByEpg.ts
ChristiaanScheermeijer: Yes, I still had doubts about that, but I think I found a way! I added a CSS variable for the EPG background color which is set when using a darker background color.

packages/ui-react/src/components/Epg/Epg.module.scss
```diff
@@ -11,7 +11,12 @@ $control-z-index: $base-z-index + 2;
   width: 100%;
   height: 100%;
   padding: calc(variables.$base-spacing * 3) 0 calc(variables.$base-spacing * 3) calc(variables.$base-spacing * 3);
-  background-color: var(--epg-background-color);
+  background-color: var(--epg-background-color, var(--body-background-color));
+
+  // Planby ScrollContainer
+  &.firstFrame > div:nth-child(2) > div > div:nth-child(2) {
+    scroll-behavior: auto !important;
+  }
```
royschut, line 19: Should we add test automation (later), to make sure it doesn't break when updating planBy?

langemike (approved):
Nice visual improvement! I tested it in different responsive sizes and in an old browser. Looks good!

ChristiaanScheermeijer:
I pushed two updates;

- Small style alignments with the features shelf PR #628 
- End-to-end fix for the collapsible description

packages/ui-react/src/components/Epg/Epg.module.scss
ChristiaanScheermeijer: I don't think that's needed. It prevents the "scroll to now" animation when opening the page, but it no crucial feature imho

langemike:
> Use markdown for video descriptions

How does it handle embedded images?

ChristiaanScheermeijer:
> > Use markdown for video descriptions
> 
> How does it handle embedded images?

Good question, should we only allow some markup syntax for the description?

langemike:
> Good question, should we only allow some markup syntax for the description?

I think we should only allow markup such as links and text markup (bold, italic, underline). It wouldn't make it really pretty, but let's leave that up to the end customer. 

So no: images, lists, code and headings.

Links should be presented differently on TV apps, because you don't want to leave the TV app, but this is out of scope for this PR.

ChristiaanScheermeijer:
Thanks @langemike! I have updated the markdown component to disable images and GitHub flavoured markdown. Most block elements are already disabled due to using `Marked#parseInline`.

PR #630, opened 2024-11-01
Feat / Registration recaptcha

royschut:
## Feat / Registration ReCaptcha

This PR adds optional Recaptcha validation, and adds the verified token to the registration api calls (currently for Cleeng).
Recaptcha [can be enabled](https://developers.cleeng.com/docs/fraud-prevention-solution-with-google-recaptcha) from the Cleeng dashboard. The site token should then be added as a custom parameter to the app-config: `captcha_site_key=<site_id>` to have the validator (by Google) added to the DOM. 

ReCaptcha can also be set as required for **payment** API calls. This is out-of-scope for this PR, but might be added later.

Screenshot 2024-10-31 at 16 28 18

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code


langemike (approved):
Simple and effective approach

ChristiaanScheermeijer (approved):
Nice addition @royschut!

My first thought, though, is if we should do the logic (specific to ReCaptcha) in the container or somewhere else. It feels a bit counterintuitive to use the ReCaptcha ref. But I also realise that error handling and form validation will suffer if we do it elsewhere... 🤔 Just putting this here for discussion!

royschut:
> My first thought, though, is if we should do the logic (specific to ReCaptcha) in the container or somewhere else. It feels a bit counterintuitive to use the ReCaptcha ref. But I also realise that error handling and form validation will suffer if we do it elsewhere... 🤔 Just putting this here for discussion!

Thanks! I've been considering alternatives, but the ReCaptcha validator should be added to the UI, and it should trigger validation onSubmit (so that form entering can be checked as a human action validator). So the alternative would be to send callbacks around, but I think that would only complicate things.

An alternative to the ref that I thought about, was to send form values to the Recaptcha component, so that it can validate by itself when certain values change. But I think onSubmit is the most reliable moment for success.

PR #631, opened 2024-11-01
Feat / Payment recaptcha

royschut:
## Feat / Payment ReCaptcha

In addition to #630, this PR optionally adds recaptcha validation to Payment endpoint calls. 

Recaptcha [can be enabled](https://developers.cleeng.com/docs/fraud-prevention-solution-with-google-recaptcha) from the Cleeng dashboard. The site token should then be added as a custom parameter to the app-config: `captcha_site_key=<site_id>` to have the validator (by Google) added to the DOM.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

Screenshot 2024-11-01 at 14 59 14


langemike (approved):
LGTM. Minor changes, compared with your previous PR

AntonLantukh:
@royschut is there any need to change the token dynamically (in custom params)? Did you consider making it part of the ini file?

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
```diff
@@ -30,6 +33,10 @@ const Checkout = () => {
   const welcomeUrl = modalURLFromLocation(location, 'welcome');
   const closeModalUrl = modalURLFromLocation(location, null);
 
+  const recaptchaRef = useRef<ReCAPTCHA>(null);
+  const captchaSiteKey = useConfigStore(({ config }) => (config.custom?.captchaSiteKey ? (config.custom?.captchaSiteKey as string) : undefined));
+  const getCaptchaValue = useEventCallback(async () => (captchaSiteKey ? (await recaptchaRef.current?.executeAsync()) || undefined : undefined));
+
```
AntonLantukh: Maybe we could have a custom hook here and also reuse it for registration?

royschut:
> @royschut is there any need to change the token dynamically (in custom params)? Did you consider making it part of the ini file?

The convenience of adding it to the app config is mainly for config switching, so for demonstration purposes. Otherwise config demonstration would not be possible for configs with captcha enabled.

packages/ui-react/src/containers/AccountModal/forms/Checkout.tsx
royschut: Not a bad idea indeed. Would you say we also return the ref from the custom hook? 
```javascript
const { recaptchaRef, captchaSiteKey } = useRecaptcha();
```

AntonLantukh: Yeah, something like this.

royschut: I've added `useRecaptcha()`

AntonLantukh (approved):

PR #634, opened 2024-11-06
Rework / UI improvements

royschut:
## Rework / UI improvements

This PR fixes some issues on the FeaturedShelf and general UI updates, found during our manual testing process:
- Fix: Too little vertical spacing on the playlist page
- Fix: Disabled featured pagination buttons have a pointer cursor
- Fix: Featured shelf not respecting safe area (mobile app)
- Fix: Searchbar not vertically centered (mobile)
- Fix: Broken featured shelf animation
- Fix: More info button not available (mobile). I've also made the More Info button into a CTA in every case that there is no Start Watching button
- Fix: Unwanted fixed hero background. We intend to make it fixed, with a fade out on scroll, like we do with the featured shelf, but this is scheduled for another moment.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/packages/ui-react/src/components/LayoutGrid/LayoutGrid.module.scss
+++ b/packages/ui-react/src/components/LayoutGrid/LayoutGrid.module.scss
@@ -4,5 +4,5 @@
 
 .cell {
   display: inline-block;
-  padding: 8px;
+  padding: 16px 8px;
 }
--- a/packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
+++ b/packages/ui-react/src/components/FeaturedShelf/FeaturedMetadata.tsx
@@ -21,12 +21,14 @@ const FeaturedMetadata = ({
   playlistId,
   style,
   hidden,
+  isMobile,
 }: {
   item: PlaylistItem | null;
   loading: boolean;
   playlistId: string | undefined;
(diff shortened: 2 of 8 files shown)


packages/ui-react/src/components/LayoutGrid/LayoutGrid.module.scss
```diff
@@ -4,5 +4,5 @@
 
 .cell {
   display: inline-block;
-  padding: 8px;
+  padding: 16px 8px;
```
langemike, line 7: Do we also want this for mobile? I personally think we want this for tablet and up. (Maybe we discussed differently yesterday, then i agree with that decision of course)

langemike (approved):
All changes look good to me! I also previewed them in my browser (desktop and mobile size).

ChristiaanScheermeijer (approved):
lgtm!

packages/ui-react/src/components/LayoutGrid/LayoutGrid.module.scss
royschut: This is not discussed, but I think I'd prefer to keep this the same for all breakpoints.

PR #638, opened 2024-11-07
Feat / Hero shelf

royschut:
## Feat / Hero shelf

This PR:
- Renames the new FeaturedShelf into `<HeroShelf />`, and makes sure it is only possible on the `0` index, as this is the only way it makes sense
- Re-introduces the additional `isFeatured` shelf state, the shelf with larger cards, on any position
- Leverages the newly introduced `layoutType` custom parameter on the shelves, which can be `hero`, `featured` or undefined
- Deprecates the `featured` boolean on the appconfig playlists

### Steps completed:

- [x] Acceptance criteria met
- [x] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/packages/common/types/config.ts
+++ b/packages/common/types/config.ts
@@ -51,6 +51,9 @@ export type Content = {
   contentId?: string;
   title?: string;
   type: AppShelfType;
+  /**
+   * @deprecated Use the custom shelf property `layoutType = 'hero' | 'featured' | undefined` instead
+   */
   featured?: boolean;
   backgroundColor?: string | null;
   custom?: Record<string, string>;
--- a/platforms/web/test-e2e/tests/home_test.ts
+++ b/platforms/web/test-e2e/tests/home_test.ts
@@ -39,9 +39,11 @@ Scenario('Header button navigates to playlist screen', async ({ I }) => {
   I.see('The Daily Dweebs');
 });
 
-Scenario('I can slide within the featured shelf', async ({ I }) => {
+// @todo: this test requires the testconfig to have `layoutType: 'hero'` on the first shelf
+// eslint-disable-next-line codeceptjs/no-skipped-tests
+Scenario.skip('I can slide within the hero shelf', async ({ I }) => {
   const isDesktop = await I.isDesktop();
-  const visibleTileLocator = locate({ css: `section[data-testid="shelf-${ShelfId.featured}"] div[data-testid="featured-metadata--visible"]` });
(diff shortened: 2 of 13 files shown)


ChristiaanScheermeijer (approved):
Nice @royschut! 🙌

packages/common/types/config.ts
```diff
@@ -51,6 +51,9 @@ export type Content = {
   contentId?: string;
   title?: string;
   type: AppShelfType;
+  /**
+   * @deprecated Use the custom shelf property `layoutType = 'hero' | 'featured' | undefined` instead
```
AntonLantukh, line 55:
```suggestion
   * @deprecated Use the custom shelf property `layoutType = 'featured' | 'featured' | undefined` instead
```

platforms/web/test-e2e/tests/home_test.ts
```diff
@@ -39,9 +39,11 @@ Scenario('Header button navigates to playlist screen', async ({ I }) => {
   I.see('The Daily Dweebs');
 });
 
-Scenario('I can slide within the featured shelf', async ({ I }) => {
+// @todo: this test requires the testconfig to have `layoutType: 'hero'` on the first shelf
+// eslint-disable-next-line codeceptjs/no-skipped-tests
+Scenario.skip('I can slide within the hero shelf', async ({ I }) => {
```
AntonLantukh, line 44: Do we expect it to start working when we add a custom param to the first shelf of `gnnuzabk` config?

royschut: Yes, I have tested that, but then the featured test will break. 
Should we change the tests and config to the following layoutTypes?
- Shelf 0: `hero`
- Shelf 1: undefined
- Shelf 2: `featured`
I suggest shelf 2 for featured, because shelf 1 is already an exception with the different aspect ratio's.

Alternatively we could change this later.

CarinaDraganJW (approved):
🚀

platforms/web/test-e2e/tests/home_test.ts
royschut: Considering the time, I suggest we postpone this test? @AntonLantukh

langemike (approved):
I think this is a really good decision

platforms/web/test-e2e/tests/home_test.ts
ChristiaanScheermeijer: Perhaps we can make a separate config for this in the next sprint?

AntonLantukh: It also works for me

PR #642, opened 2024-11-14
Feat / Hero and HeroShelf improvements

royschut:
## Description

This PR improves the Hero and HeroShelf and removes Card shadow:
- Fade out the background incrementally based on scroll position. This brings a much smoother scrolling experience.
- Add swiping gesture to HeroShelf for all breakpoints. Previously swiping was only possible for mobile devices, but on tablets or landscape mobile it is needed as well to navigate between slides. I have tried simplifying the code by merging the two implementations, leaving only style differences between mobile and the rest.
- Some minor style updates to HeroShelf: a bit more height on tablet, better support for mobile landscape and respecting the safe-area better (useful for non-web platforms).
- Make the Hero fixed as well, and add the same fade behaviour.
- Remove the Card box-shadow. Since we've re-introduced support for a light background, the card shadow seemed a bit overdone, or out-dated even in some opinions (have a look here: https://vd-ott.com/?app-config=kcm2oevm). We figured that the card shadow could be left out entirely, as it was once introduced mainly because of the blur background, which was removed quite some time ago. 

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [x] UX tested
- [x] Reviewed own code

### DIFF
--- a/packages/ui-react/src/components/HeroShelf/HeroShelf.module.scss
+++ b/packages/ui-react/src/components/HeroShelf/HeroShelf.module.scss
@@ -8,7 +8,7 @@ $desktop-max-height: 700px;
 $desktop-min-height: 275px;
 
 $tablet-height: 70vw;
-$tablet-min-height: 375px;
+$tablet-min-height: 550px;
 
 $mobile-height: 70vh;
 $mobile-min-height: 450px;
@@ -47,13 +47,13 @@ $mobile-landscape-height: 100vh;
   }
 
   @include responsive.mobile-only() {
-    height: calc($mobile-height - variables.$header-height - var(--safe-area-top, 0));
+    height: calc($mobile-height - variables.$header-height - var(--safe-area-top, 0px));
     min-height: calc($mobile-min-height - variables.$header-height);
-    margin-top: var(--safe-area-top, 0);
   }
 
   @include responsive.mobile-only-landscape() {
     height: calc($mobile-landscape-height - variables.$header-height);
+    min-height: initial;
     padding: 0;
   }
 }
@@ -113,6 +113,7 @@ $mobile-landscape-height: 100vh;
 
   @include responsive.mobile-only-landscape() {
     height: $mobile-landscape-height;
+    min-height: initial;
   }
 }
 
@@ -178,7 +179,14 @@ $mobile-landscape-height: 100vh;
   }
 }
 
-.metadataMobile {
+.swipeSlider {
+  display: flex;
+  align-items: flex-end;
+  width: 100%;
+  height: 100%;
+}
+
+.swipeSliderMobile {
   width: 100%;
   height: 100%;
 }
@@ -188,6 +196,7 @@ $mobile-landscape-height: 100vh;
   display: flex;
   flex-direction: column;
   gap: 24px;
+  width: 46%;
   max-width: 46%;
   padding-left: calc(variables.$base-spacing * 4);
 
@@ -225,6 +234,7 @@ $mobile-landscape-height: 100vh;
 
   @include responsive.mobile-only-landscape() {
     max-width: 70%;
+    min-height: initial;
     padding: 0 calc(variables.$base-spacing * 3) calc(variables.$base-spacing * 3) calc(variables.$base-spacing * 3);
   }
 
--- a/packages/ui-react/src/components/HeroShelf/HeroShelf.module.scss
+++ b/packages/ui-react/src/components/HeroShelf/HeroShelf.module.scss
@@ -49,6 +49,7 @@ $mobile-landscape-height: 100vh;
   @include responsive.mobile-only() {
     height: calc($mobile-height - variables.$header-height - var(--safe-area-top, 0));
     min-height: calc($mobile-min-height - variables.$header-height);
+    margin-top: var(--safe-area-top, 0);
   }
 
   @include responsive.mobile-only-landscape() {
@@ -97,6 +98,7 @@ $mobile-landscape-height: 100vh;
   height: 56.25vw;
   max-height: 700px;
   background-color: var(--hero-shelf-background-color);
+  transition: opacity ease-out 0.3s;
 
   @include responsive.tablet-only() {
     height: $tablet-height;
@@ -106,20 +108,12 @@ $mobile-landscape-height: 100vh;
   @include responsive.mobile-only() {
     height: $mobile-height;
     min-height: $mobile-min-height;
+    margin-top: var(--safe-area-top, 0);
   }
 
   @include responsive.mobile-only-landscape() {
     height: $mobile-landscape-height;
   }
-}
-
-.undimmed {
-  opacity: 1;
-  transition: opacity ease-out 0.3s;
-}
-
-.dimmed {
-  opacity: 0.01;
 }
 
 .background {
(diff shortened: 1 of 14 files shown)


langemike:
These changes add a lot of value! Swiping works better, indeed. I think we're on the right track, but I noticed some issues that are related to different (smartphone) resolutions.

There are 3 styling-related issues. These weren't there previously.
1. The Hero title can get cut-off on top. 
2. The buttons look crammed now (too narrow button container)

3. Also in some specific resolutions, the Start Watching button doesn't show, as can be seen here:
Screenshot 2024-11-14 at 13 18 07

And I also noticed 2 issues in the scrolling (fade to transparent) behaviour:
1. It should finish its transition when the hero is out of view. But it happens around the second regular shelf
2. It doesn't fade completely away

These happen on certain resolutions such as 2340x1080 (on my phone). On other resolution, I see it works as intended.

I tested this on the deploy preview.

Video recording showcasing styling issues 1 & 2 and scrolling issues 1 & 2.

https://github.com/user-attachments/assets/4f9bc07f-54d0-4a9b-a31f-3a3f9cc12c30

packages/ui-react/src/components/HeroShelf/HeroShelfMetadata.tsx
```diff
@@ -21,17 +22,18 @@ const HeroShelfMetadata = ({
   playlistId,
   style,
   hidden,
-  isMobile,
 }: {
   item: PlaylistItem | null;
   loading: boolean;
   playlistId: string | undefined;
-  style?: CSSProperties;
+  style: CSSProperties;
```
ChristiaanScheermeijer, line 29: Is this an unwanted change? I do believe `isMobile?: boolean;` can be removed

platforms/web/src/styles/main.scss
```diff
@@ -19,10 +19,13 @@
...
-  --safe-area-top: 0;
-  --safe-area-right: 0;
-  --safe-area-bottom: 0;
-  --safe-area-left: 0;
+  /* stylelint-disable length-zero-no-unit */
+  --safe-area-top: 0px; 
+  --safe-area-right: 0px;
+  --safe-area-bottom: 0px;
+  --safe-area-left: 0px;
+  /* stylelint-enable length-zero-no-unit */
```
ChristiaanScheermeijer, line 27: How come using `0` is a problem?

packages/ui-react/src/components/HeroShelf/HeroShelf.module.scss
```diff
@@ -97,6 +98,7 @@ $mobile-landscape-height: 100vh;
   height: 56.25vw;
   max-height: 700px;
   background-color: var(--hero-shelf-background-color);
+  transition: opacity ease-out 0.3s;
```
ChristiaanScheermeijer: IMO the scroll animation feels better without using an extra CSS transition (no delay). The scroll events already defines the animation timeline and also is more performance friendly, but that is probably not even noticeable.

packages/ui-react/src/components/HeroShelf/HeroShelf.tsx
```diff
@@ -27,20 +27,28 @@ const HeroShelf = ({ playlist, loading = false, error = null }: Props) => {
...
   const isMobile = breakpoint <= Breakpoint.sm;
-  const scrolledDown = useScrolledDown(500, 200);
+  const posterRef = useRef<HTMLDivElement>(null);
   const [direction, setDirection] = useState<'left' | 'right' | null>(null);
   const [animationPhase, setAnimationPhase] = useState<'init' | 'start' | 'end' | null>(null);
+  const [isSwipeAnimation, setIsSwipeAnimation] = useState(false);
 
-  const slideTo = (toIndex: number) => {
+  useScrolledDown(50, isMobile ? 200 : 700, (progress: number) => {
+    if (posterRef.current) posterRef.current.style.opacity = `${Math.max(1 - progress, isMobile ? 0 : 0.1)}`;
```
ChristiaanScheermeijer, line 36: Should we also fade away the pagination?

packages/ui-react/src/components/HeroShelf/HeroSwipeSlider.tsx
```diff
@@ -73,18 +88,17 @@ const HeroShelfMetadataMobile = ({ item, leftItem, rightItem, playlistId, loadin
 
   useLayoutEffect(() => {
     if (!containerRef.current) return;
-
-    if (direction === 'left') {
+    if (isSwipeAnimation && direction === 'left') {
       containerRef.current.style.transition = 'transform 0.2s ease-out';
       containerRef.current.style.transform = `translateX(100%)`;
-    } else if (direction === 'right') {
+    } else if (isSwipeAnimation && direction === 'right') {
```
ChristiaanScheermeijer, line 94: Would it look nicer for tablets and larger screen sizes to decrease the left/right offset and use opacity as well? It works great now, but the slide animation distance is pretty big (opinionated ofc)

royschut:
> There are 3 styling-related issues. These weren't there previously.
> 
> 1.

Thanks for the thorough testing @langemike! My last commit fixes the issues you mentioned.
About the scrolling fade: I have updated the end of the scroll/opacity window slightly, but if I change it too much, I find the transition a bit too sudden. You're now previewing on landscape, which is quite a specific viewport, I prefer to focus on more used viewports (without introducing too many variations). I have also removed the transition, as Christiaan suggested, which might help in your experience. How does the latest update feel?
Leaving 0.1 opacity is a choice by the way.

packages/ui-react/src/components/HeroShelf/HeroShelfMetadata.tsx
royschut: It's intentional, style should now always be set.
isMobile is removed in my last commit.

platforms/web/src/styles/main.scss
royschut: Chrome doesn't respect it within a `calc` function, and ignores setting the shelf height.

packages/ui-react/src/components/HeroShelf/HeroShelf.module.scss
royschut: Thanks, I agree. Changed!

packages/ui-react/src/components/HeroShelf/HeroShelf.tsx
royschut: I didn't find it necessary anymore. Before they were also fixed, but they scroll away now.

packages/ui-react/src/components/HeroShelf/HeroSwipeSlider.tsx
royschut: I think you mean the animation after a swipe gesture (because the normal slide hasn't changed). The thing is: when you drag the metadata over the screen, then let go, you want the gesture to be continued similarly. So I cannot really change the opacity, you expect it to be opaque. Same for the offset, it comes from outside of the screen. I could only update the transition perhaps?

langemike:
> How does the latest update feel?

Better but the video details section does a weird jump now during the slide animation.

> Leaving 0.1 opacity is a choice by the way.

Ok. Good to know!

I think the fade out to 0.1 still happens quite late. But I can live with that.

updated video recording:
https://github.com/user-attachments/assets/df89d9a8-8bd5-40a4-86ed-44551699325b

ps. This can also be reproduced in Chrome using "responsive mode" using 2340x1080 or compatible screen size.

royschut:
> I think the fade out to 0.1 still happens quite late. But I can live with that.

It's a compromise on mobile landscape, to bring the best experience on more common viewports (mobile portrait, tablet and desktop). The only alternative I see is introducing more variation for the scroll/fade position, but since it is done from javascript, I prefer to keep it simple. But open to suggestions.

langemike (approved):

ChristiaanScheermeijer (approved):

packages/ui-react/src/components/HeroShelf/HeroSwipeSlider.tsx
royschut: We can consider this later as well!

PR #643, opened 2024-11-14
Fix /  Translatable shelf titles for hub pages

langemike:
This fixes an issue where translated playlist titles didn't show up as translated shelf titles on the hub page.

### DIFF
--- a/packages/ui-react/src/containers/ShelfList/ShelfList.tsx
+++ b/packages/ui-react/src/containers/ShelfList/ShelfList.tsx
@@ -75,7 +75,7 @@ const ShelfList = ({ rows }: Props) => {
           const posterAspect = parseAspectRatio(playlist.cardImageAspectRatio || playlist.shelfImageAspectRatio);
           const visibleTilesDelta = parseTilesDelta(posterAspect);
 
-          const translatedKey = custom?.[`title-${language}`];
+          const translatedKey = custom?.[`title-${language}`] || (playlist?.[`title-${language}`] as string);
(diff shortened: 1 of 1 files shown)


AntonLantukh (approved):

ChristiaanScheermeijer (approved):

packages/ui-react/src/containers/ShelfList/ShelfList.tsx
```diff
@@ -75,7 +75,7 @@ const ShelfList = ({ rows }: Props) => {
           const posterAspect = parseAspectRatio(playlist.cardImageAspectRatio || playlist.shelfImageAspectRatio);
           const visibleTilesDelta = parseTilesDelta(posterAspect);
 
-          const translatedKey = custom?.[`title-${language}`];
+          const translatedKey = custom?.[`title-${language}`] || (playlist?.[`title-${language}`] as string);
```
royschut, line 78:
```suggestion
          const translatedKey = custom?.[`title-${language}`] || (playlist?.[`title-${language}`] as string | undefined);
```

royschut (approved):

packages/ui-react/src/containers/ShelfList/ShelfList.tsx
langemike: TypeScript already infers that the value could be undefined due to optional chaining (?.).

royschut: You're bypassing that because of the typecast. @langemike

PR #644, opened 2024-11-14
Feat / Deterministic public URLs

langemike:
When you build an OTT app for another platform than web, we cannot rely on the value of `window.location.origin`.

In a lot of hybrid web-apps, this variable will contain a URL to a local webserver. Eg. `capacitor://localhost` for Capacitor based platforms.

This caused two bugs:
1. It caused a "not valid URL" error message in the "Reset password" form - [OTT-2707](https://videodock.atlassian.net/browse/OTT-2707)
2. It caused an infinite spinner when using "Send password reset mail" from your account settings. This happens because of the same (but uncaught) error. - [OTT-2804](https://videodock.atlassian.net/browse/OTT-2804)

While we fixed this issue, we took it a step further to completely remove `window.location.origin` references throughout the project.

When `APP_PUBLIC_URL` is not defined within your (web) `.env`-file, we still fallback on the the same value, making this a non-breaking change.

### DIFF
--- a/packages/common/src/env.ts
+++ b/packages/common/src/env.ts
@@ -5,6 +5,7 @@ export type Env = {
   APP_PLAYER_ID: string;
   APP_FOOTER_TEXT: string;
   APP_DEFAULT_LANGUAGE: string;
+  APP_PUBLIC_URL: string;
 
   APP_DEFAULT_CONFIG_SOURCE?: string;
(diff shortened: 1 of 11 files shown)


ChristiaanScheermeijer (approved):
Nice! 🙌

royschut (approved):

PR #648, opened 2024-11-20
OWA-99: fix: do not perform logout if password reset with token

mirovladimitrovski:
## Description

This PR solves [OWA-99](https://jwplayer.atlassian.net/browse/OWA-99).

Whenever a password change or reset is requested and finished successfully, we're performing a logout request afterwards. This is fine is the password change is requested by a logged in user. But the logout request should NOT be performed when a logged out user has forgotten their password and starts a process for resetting their password. With changes in this PR the logout request will only be made if there is a user logged in.

### Steps completed:

- [x] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [x] Reviewed own code

[OWA-99]: https://jwplayer.atlassian.net/browse/OWA-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/common/src/services/integrations/jwp/JWPAPIService.ts
+++ b/packages/common/src/services/integrations/jwp/JWPAPIService.ts
@@ -51,7 +51,7 @@ export default class JWPAPIService {
   };
 
   removeToken = async () => {
-    await Promise.all([this.storageService.removeItem(INPLAYER_TOKEN_KEY), this.storageService.removeItem(INPLAYER_IOT_KEY)]);
+    await Promise.all([this.storageService.removeItem(INPLAYER_TOKEN_KEY, false), this.storageService.removeItem(INPLAYER_IOT_KEY, false)]);
   };
 
   isAuthenticated = async () => {
--- a/packages/ui-react/src/containers/AccountModal/forms/EditPassword.tsx
+++ b/packages/ui-react/src/containers/AccountModal/forms/EditPassword.tsx
@@ -36,6 +36,8 @@ const ResetPassword = ({ type }: { type?: 'add' }) => {
     const resetToken = resetPasswordToken || resetPasswordTokenParam;
 
     try {
+      let pathname = location.pathname;
+
       if (resetToken) {
         await accountController.changePasswordWithToken(emailParam || '', password, resetToken, passwordConfirmation);
       } else {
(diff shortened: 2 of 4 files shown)


packages/common/src/services/integrations/jwp/JWPAPIService.ts
```diff
@@ -51,7 +51,7 @@ export default class JWPAPIService {
   };
 
   removeToken = async () => {
-    await Promise.all([this.storageService.removeItem(INPLAYER_TOKEN_KEY), this.storageService.removeItem(INPLAYER_IOT_KEY)]);
+    await Promise.all([this.storageService.removeItem(INPLAYER_TOKEN_KEY, false), this.storageService.removeItem(INPLAYER_IOT_KEY, false)]);
```
AntonLantukh, line 54: Can't we also use a prefix for these items?

mirovladimitrovski: The auth token has been stored under the key `inplayer_token` for some time now, it's the way the Inplayer SDK stored it while we used. We don't have to stick to that anymore, but if we change the key (add a prefix) we'd make existing tokens unreachable, unless we implement a small backward-compatibility change for "converting" them (in reality, it's just removing the item with the old key and store it with the new prefixed key).

mirovladimitrovski: Btw, for notifications we're still using the Inplayer SDK, and it uses a separate token in local storage under the name `inplayer_iot`. No prefix either.

packages/ui-react/src/containers/AccountModal/forms/EditPassword.tsx
```diff
@@ -32,24 +32,23 @@ const ResetPassword = ({ type }: { type?: 'add' }) => {
...
       } else {
-        if (!resetToken) {
+        if (!user) {
           setErrors({ form: t('reset.invalid_link') });
-
           return setSubmitting(false);
         }
-        await accountController.changePasswordWithToken(emailParam || '', password, resetToken, passwordConfirmation);
+
+        await accountController.changePasswordWithOldPassword(oldPassword || '', password, passwordConfirmation);
```
AntonLantukh, line 51: Could you also try to change the password for JWP within your account? I see infinite screen loader now.

AntonLantukh (approved):

PR #650, opened 2024-11-21
Feat / Replace hard-coded English string with translatable strings

langemike:
This change adds missing translations for the OTT platform. We believe that there are no hard-coded English labels present anymore in the code. This means we 100% of the text occurrences can be translated.

Adds:
- Translations for credit card (number + expiry date)
- Abbreviations for hours and minutes
- Translated open/close aria label for search

Let me know if there is still a hard-coded English string anywhere.

Ticket: [OTT-2762](https://videodock.atlassian.net/browse/OTT-2762)

### DIFF
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -1,9 +1,11 @@
-export const formatDurationTag = (seconds: number) => {
+import type { DurationAbbreviation } from '../../types/i18n';
+
+export const formatDurationTag = (seconds: number, { minutesAbbreviation }: Pick<DurationAbbreviation, 'minutesAbbreviation'>) => {
   if (!seconds) return '';
 
   const minutes = Math.ceil(seconds / 60);
 
-  return `${minutes} min`;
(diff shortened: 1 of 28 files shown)


royschut (approved):
Nice!

About the abbreviation: I think in some languages the number can come after the abbr, and in other languages they have a space between them (`1hr`, `1 h` French, and I think something like `h 1` in Arabic?). That could be fixed if we add the number to the translation `{{count}}hr`. But perhaps we can implement that later, this PR makes it much better.

langemike:
> That could be fixed if we add the number to the translation {{count}}hr. But perhaps we can implement that later, this PR makes it much better.

Good idea. I will create a ticket for this in our backlog. Since I already introduced the new (de/fr/nl) languages based on the current approach, it is quite a hassle to synchronize this across branches. Do-able of course, but my current priority is to get these languages to new platforms first.

CarinaDraganJW (approved):
🚀

AntonLantukh (approved):

PR #652, opened 2024-11-22
Fix / Missing playlist grid title

langemike:
We found another missing translation from the back-end for the Playlist Grid.

### DIFF
--- a/packages/ui-react/src/pages/ScreenRouting/playlistScreens/PlaylistGrid/PlaylistGrid.tsx
+++ b/packages/ui-react/src/pages/ScreenRouting/playlistScreens/PlaylistGrid/PlaylistGrid.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import { Helmet } from 'react-helmet';
+import { useTranslation } from 'react-i18next';
 import { shallow } from '@jwp/ott-common/src/utils/compare';
 import type { Playlist, PlaylistItem } from '@jwp/ott-common/types/playlist';
 import { useAccountStore } from '@jwp/ott-common/src/stores/AccountStore';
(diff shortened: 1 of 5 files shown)


royschut (approved):
👍

packages/ui-react/src/pages/ScreenRouting/playlistScreens/PlaylistGrid/PlaylistGrid.tsx
```diff
@@ -30,7 +32,8 @@ const PlaylistGrid: ScreenComponent<Playlist> = ({ data, isLoading }) => {
     setFilter('');
   }, [data.feedid]);
 
-  const pageTitle = `${data.title} - ${config.siteName}`;
+  const title = (data?.[`title-${i18n.language}`] as string) || data.title;
```
AntonLantukh: @langemike Do you think we could create a small function to create a translation key? Something like this:

```
export const useTranslationKey = (key: string) => {
   const { i18n } = useTranslation();
   
   return `${key}-${i18n.language}`
}
```
We could also reuse it for ShelfList component.

langemike: Using a hook can be quite limiting, because it can only be called from the body of the functional component (e.g. it cannot be used within a loop). An alternative approach could be to create an i18n utility function. The only benefit will be, that we will have an increased overview where translation keys are being used in our code. But since it does so little, I'm a bit hesitant to add it. Curious about what you or others think.

```
export const translatedKey = (obj: Record<string, any>, key: string, language: string) => {
   return obj?.[`${keyPrefix}-${language}`];
}
```

AntonLantukh: It is not a blocker here, but could still be useful to have easier control over the translation key pattern in case additional change / logic is needed.

AntonLantukh (approved):

packages/ui-react/src/pages/ScreenRouting/playlistScreens/PlaylistGrid/PlaylistGrid.tsx
langemike: I discussed it with @royschut and of course, because the hook only gives back a string it will only be called once. A hook is sufficient. I will add it!

AntonLantukh (approved):

PR #656, opened 2024-11-29
Chore / maintenance 2024 q4

ChristiaanScheermeijer:
## Maintenance 2024 Q4

As the title already suggests, this PR is a maintenance update which means that we fix security vulnerabilities by upgrading dependencies. In some cases there is no update available and we need to use resolutions to upgrade nested dependencies.

We also look into packages that can be replaced or migrated to more modern packages, fix deprecation warnings and do some housekeeping.

I've split each change per commit with self explaining changes, but here is a list of the changes including the motivation where necessary:

[chore: add access bridge test env](https://github.com/jwplayer/ott-web-app/pull/656/commits/ccf21b0262b3e398c697f91e7e6fbf9c6fad26f1)

When running `$ yarn test` locally, I ran into an error. By adding a `env.test` file in the `access-bridge` platform this error is cleared. I could then also remove the redundant env overrides in the unit test workflow file.

[chore: upgrade yup to 1.4.0](https://github.com/jwplayer/ott-web-app/pull/656/commits/11814e1851b4db7aa30bd850b1d86a4bbf3ae5a5)

Yup had a major upgrade (0.32.11 -> 1.4.0) which came with a few breaking changes. The biggest in this project was the usage of `SchemaOf` which is renamed to `ObjectSchema`. But after the upgrade, the typings (correctly) spotted many mistakes between the yup schema and TS types. So, I corrected them in the yup schema. This could mean that the config validation becomes more strict. Especially in cases where we don't expect `null` values. @AntonLantukh do you have the exact typings used in the app config API so we can ensure that we have the correct validation schema now?

PS I also considered to migrate to zod...

[chore: upgrade date-fns to 4.1.0](https://github.com/jwplayer/ott-web-app/pull/656/commits/897b0e8866efc8d69ff204193dbbc57d50e67409)

date-fns also had a major upgrade (3.6.0 -> 4.1.0) which adds timezone support. It didn't had breaking changes for us.

[chore: upgrade marked to 15.0.3](https://github.com/jwplayer/ott-web-app/pull/656/commits/1c0117cc203655944afc0a5c6db23c3755f45aa7)

Marked had many major releases (4.3.0 -> 15.0.3). We don't need the separate `@types/marked` package anymore because the codebase has been rewritten to TypeScript. There are some changes in the render pipeline which meant I had to revert some recent changes to disable images when rendering inline markdown (video details, footer, ...).

[chore: upgrade i18next to 24.0.2 and react-i18next to 15.1.3](https://github.com/jwplayer/ott-web-app/pull/656/commits/461fc788c39b8f7755dab539d89c4a994ed66ec4)

No changes were needed for the i18next (22 -> 24) upgrade.

[chore: upgrade vite to 5.4.11](https://github.com/jwplayer/ott-web-app/pull/656/commits/bfceaad52a638675c7faa10a29b9a60b8cf86690)
[chore: upgrade vitest to 2.1.6](https://github.com/jwplayer/ott-web-app/pull/656/commits/6cfb016f7f4a2a744e04ad03a380f97601126d8a)

This was no major upgrade but I did it separate anyway to ensure this was tested isolated. [Vite v6](https://vite.dev/blog/announcing-vite6) has been released 7 days ago, but I didn't want to migrate this yet and give this a little more time to mature. 

... some minor fixes in between ...

[chore: update workflows to node 22](https://github.com/jwplayer/ott-web-app/pull/656/commits/73c25dcf5ecd1b3f51676eb1cd490d12284150ec)

[Node 22](https://nodejs.org/en/about/previous-releases) is the active release now, so I upgraded the Node workflows to use version 22.

[chore: upgrade to sass-embedded](https://github.com/jwplayer/ott-web-app/pull/656/commits/59a6ae1a6f1738cf1c93cae19abc15fc8dd5f57e)

It is suggested by [Vite](https://vite.dev/config/shared-options#css-preprocessoroptions) to use the `sass-embedded` package over the `sass` package and use the `modern-compiler` API. There was also a warning for this in the Vite output when running `yarn test` or `yarn start`. 

[chore: fix sass deprecation warnings](https://github.com/jwplayer/ott-web-app/pull/656/commits/c55a24287641fa43b8fa069fd5f2d96f13835d4c)

After upgrading to `sass` (or `sass-embedded` 1.81.1) a few warnings showed in the console. Most of them are related to this: https://sass-lang.com/documentation/breaking-changes/mixed-decls/. The problem is that when using mixings like so;

```scss
.selector {
  @include mixin;
  font-size: 16px;
}
```

The sass deprecation detection notified that we should use the `& {}` workaround. I was able to fix this by changing the declaration order in stylelint. This results in:

```scss
.selector {
  font-size: 16px;
  @include mixin;
}
```

The remaining commits should be clear by itself.


langemike (approved):
Nice maintenance PR. Thanks for the elaborate motivation per commit/change. The end result looks clean.

PR #658, opened 2024-12-04
Chore / payments e2e tests stability improvements

ChristiaanScheermeijer:
## Description

I found that the payments e2e tests were failing many times and while looking into this I found some problems in the tests and a bug in the payments form. These fixes should be rebased to #656 and #657.

- The loading times can vary a lot and the biggest culprit was the timeout of `I.waitForLoaderDone()`, so I increased the timeout for this test specifically
- The SubscriptionService couldn't be injected due to an error in the test build caused by a missing `@inject()` annotation
- The Checkout form was showing the Adyen card payment form while no payment was needed. This caused the coupon test to fail


langemike (approved):
Nice appropriate fixes. Glad you were able to fix this e2e issue

PR #659, opened 2024-12-05
Fix / crash when personal shelf data is not valid

ChristiaanScheermeijer:
## Description

This PR fixes two problems with the personal shelves (favorites & watch history) when we try to load an account with invalid data.

The first problem is that the web app crashes while validating the data. This is a bit to aggressive and can be prevented.

When a validation error occurs, it basically means we can't guarantee that the integrity or the origin of the data. So instead of overwriting the data, we now prevent saving the data so we don't accidentally reset the favorites or watch history list.

When this happens, the user isn't notified, but we log an error to the console. We can improve this in the future to show a message to the user in this scenario.


langemike (approved):
I haven't encountered this issue myself, yet. But your approach is solid and easy to understand. Nice improvement!

PR #664, opened 2025-01-07
Refactor / sync ott changes

ChristiaanScheermeijer:
## Description

We have made changes in the fork that allows us using the `useWatchHistory` hook for capacitor and react native platforms as well. For this, the JW Player instance parameter is removed in the hooks and more player events are exposed as callback prop.

This PR also contains a small change that decreases the shelf title font size for mobile devices.

Both changes have been in the fork for a while and are well tested.

### DIFF
--- a/packages/common/types/playlist.ts
+++ b/packages/common/types/playlist.ts
@@ -17,6 +17,7 @@ export type DRM = {
   };
   fairplay?: {
     processSpcUrl: string;
+    certificateUrl: string;
   };
 };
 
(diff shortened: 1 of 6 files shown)


langemike (approved):

royschut (approved):
Let's sync!

e2e issues however. Let's first rerun it..

ChristiaanScheermeijer:
@langemike @royschut I added the missing DRM property as well

PR #666, opened 2025-01-29
Refactor/move user routes to app routes

ChristiaanScheermeijer:
## Description

This refactor moves the route definitions for the user screen to the AppRoutes file. The biggest advantage with this change is that the User screen doesn't need to be hoisted (copied for other platforms) when only wanting to change a single screen.


langemike (approved):
This has been internally proposed already.

royschut (approved):

PR #667, opened 2025-02-17
Feat / sync generic changes from our private repository

ChristiaanScheermeijer:
## Description

We applied some fixes and changes in our private repository which still needed to be synced. I'm currently collecting these changes and group them by size/feature. This PR contains multiple small fixes and changes that don't need separate PRs (let me know if you disagree).

**chore: use explicit common or module js extensions**
In preparation for ModuleJS, which isn't possible quite yet, we have renamed the files that are compatible with ModuleJS.

**feat(entitlement): add access type to generic entitlement payload**
To optimize the entitlement service, we added an accessType parameter so we can optimistically check for entitlement at the correct provider. E.g. when an item is free, we only need to fetch the item using the JWP APIs and don't need to validate elsewhere.

fix(entitlement): refresh token before calling entitlement
The Cleeng token is only valid for 30m and wasn't getting refreshed when staying on the media page.

**fix: catch error when retrieving the offers fail**
This prevents an infinite spinner when the offers couldn't be fetched

**feat: add external payment methods and mapping**
There are more payment methods added, although I think we can move this part to the integration instead. The integration should tell the UI that an external subscription provider is used with an optional URL to manage the subscription.

**fix: prevent google fonts preload when using system font**
A small fix to prevent a preload tag when using font a non-Google font (e.g. `system:Verdana`)


langemike (approved):
Nice improvements

royschut (approved):

PR #669, opened 2025-02-17
Feat: cookie banner and defer GTM load

ChristiaanScheermeijer:
## Description

This PR introduces an optional Cookie banner when using GTM. It is possible to only load the GTM script when the user accepts the cookie/privacy policy to be (GDPR) compliant.

The same applies to the Google ReCaptcha, which uses a cookie to identity the user.

This can be configured using build environment variables:

```dotenv
APP_CONSENT_COOKIE_NAME=iaccept
APP_CONSENT_COOKIE_POLICY_URL=https://policy.yoursite.com
APP_GTM_TAG_ID=GTM-123456-4
APP_GTM_TAG_SERVER=https://gtm.yoursite.com
APP_GTM_LOAD_ON_ACCEPT=true
```

Screenshot 2025-02-17 at 17 04 44


langemike (approved):
Already internally reviewed. Looks the same.
I think the "Code scanning / CodeQL" alert can be dismissed, as this isn't user generated content.

royschut (approved):

PR #670, opened 2025-02-18
Fix / infinite scrolling not working

ChristiaanScheermeijer:
## Description

The infinite scrolling wasn't working anymore. For apps the scroll view may not be the `window`, so I kept the `useWindow={false}` and added a `getScrollParent` utility that uses the `document.scrollingElement` variable.

@royschut @langemike did we disable this on purpose by any chance? 🤔


langemike:
This `useWindow={false}` was added originally to capacitor to fix infinite scrolling. This was before we override the `scrollingElement`. It is related to this commit: [5d2a44fee3ae62e20074d6c17a695dc845e7b989](https://github.com/Videodock/vd-ott-apps/pull/53/commits/5d2a44fee3ae62e20074d6c17a695dc845e7b989)

But since then, the same prop and value also have been applied across different platforms and in `ui-react`. I don't know what its intention is there.

It could be that it is irrelevant. But I have to dive deep before I can come to such a conclusion.

royschut (approved):

PR #651, opened 2024-11-22
Feat / Add new languages (nl/fr/de)

langemike:
This change adds and activates 3 new languages:
- Dutch
- French
- German

I kept `us_state.json` in English.

These are all translated using ChatGTP.

Tickets: [OTT-2756](https://videodock.atlassian.net/browse/OTT-2756) & [OTT-2763](https://videodock.atlassian.net/browse/OTT-2763)

PR #675, opened 2025-02-28
Fix / sync translations

ChristiaanScheermeijer:
## Description

After merging all pull-requests I found that some changes were made after the #651 PR which adds more translations. I ran `$ yarn i18next` and filled the missing translations.

I also made one change to the "no_subscription" translation which included "Movies and series". We can't assume that, because a subscription could also give selectively access or the there could be no movies and series at all (like for Symphony).


platforms/web/public/locales/fr/user.json
```diff
@@ -97,12 +97,13 @@
     "expiry_date": "Date d'expiration",
     "granted_subscription": "Abonnement accordé",
     "hidden_transactions_one": "Une autre transaction",
+    "hidden_transactions_many": "",
```
royschut: You have some empty translations

ChristiaanScheermeijer: That's correct! I was aware of that and assumed that i18next would fallback to the same language `_other` translation. But this isn't the case and it falls back to the English plural translation instead.

I have modified the i18next-workspaces script to re-use the `_other` translation when the `_many` is empty.

royschut (approved):

PR #677, opened 2025-03-12
OWA-107: Fix broken SCA flow payment finalization

mirovladimitrovski:
## Description

This PR solves [OWA-107](https://jwplayer.atlassian.net/browse/OWA-107)

### Steps completed:

- [ ] Acceptance criteria met
- [ ] Unit tests added
- [ ] Translations added
- [ ] UX tested
- [ ] Reviewed own code

[OWA-107]: https://jwplayer.atlassian.net/browse/OWA-107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ChristiaanScheermeijer:
Hi @mirovladimitrovski, I believe this fix is for JWP integrations only, should we handle this from the JWP integration instead of making it a separate component/modal? 

I think if we do something like this in the `JWPAccountService.ts` together with the reload subscription it should work basically the same, right?

```ts
  private async handlePPVPaymentIntent(url: string) {
    const searchParams = new URLSearchParams(url.split('?')[1]);
    const paymentIntent = searchParams.get('payment_intent');

    if (paymentIntent) {
      await this.apiService.post<CommonResponse>('/payments', { pi_id: paymentIntent }, { withAuthentication: true });
    }
  }

  initialize = async (config: Config, url: string, _logoutFn: () => Promise<void>) => {
    const jwpConfig = config.integrations?.jwp;

    await this.handlePPVPaymentIntent(url);
```

mirovladimitrovski:
> Hi @mirovladimitrovski, I believe this fix is for JWP integrations only, should we handle this from the JWP integration instead of making it a separate component/modal?
> 
> I think if we do something like this in the `JWPAccountService.ts` together with the reload subscription it should work basically the same, right?
> 
> ```ts
>   private async handlePPVPaymentIntent(url: string) {
>     const searchParams = new URLSearchParams(url.split('?')[1]);
>     const paymentIntent = searchParams.get('payment_intent');
> 
>     if (paymentIntent) {
>       await this.apiService.post<CommonResponse>('/payments', { pi_id: paymentIntent }, { withAuthentication: true });
>     }
>   }
> 
>   initialize = async (config: Config, url: string, _logoutFn: () => Promise<void>) => {
>     const jwpConfig = config.integrations?.jwp;
> 
>     await this.handlePPVPaymentIntent(url);
> ```

This is a good idea actually, much lighter in code. I'll have it done shortly.

mirovladimitrovski:
@ChristiaanScheermeijer 
In practice, your suggestion introduces a different problem. The problem is that the confirmation request is done very early during the app's boot-up, before we're subscribed to the notification websocket that is supposed to notify us that access has been granted. I still think the idea is good though, so I'm working on solving the problem.

packages/ui-react/src/components/FinalizePayment/FinalizeStripePpvPayment.tsx
```diff
@@ -0,0 +1,37 @@
...
+import CheckoutController from '@jwp/ott-common/src/controllers/CheckoutController';
+
+const FinalizeStripePpvPayment = () => {
+  const checkoutController = getModule(CheckoutController);
+
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const finalize = useCallback(async (paymentIntent: string) => {
+    try {
+      await checkoutController.finalizeStripePpvPayment(paymentIntent);
```
AntonLantukh, line 13: Did you try to reuse everything in FanalizeAdyenPayment except using a different method instead of a `checkoutController.finalizeAdyenPayment` one?

mirovladimitrovski: I did think it through and couldn't think of anything that would improve the reusability further than what it is with this PR. Do you have anything specific on your mind?

AntonLantukh (approved):

PR #682, opened 2025-04-07
[OWA-104] Reference implementation of the labels filter in the web app

CarinaDraganJW:
## Description

- Added reference example implementation for labels filtering in the web app

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -74,9 +74,6 @@ the given screen.
 **menu[].filterTags** (optional)
 
 You can optionally define a list of comma separated tags which are used in the "filters" section on the screen.
-The OTT Web App offers a demonstration of its filtering capabilities. This example can be enabled by configuring the `enableLabelsFiltering` custom parameter.
-
-Filtering by device type, country, and day of week is supported. See the Home.tsx file for a code example.
 
 ---
 
@@ -145,6 +142,16 @@ Controls if the playlist should be used as a large "Featured" shelf on your JW O
 **content[].backgroundColor** (optional)
 
 You can change the background color of the shelf with the help of this property (e.g., #ff0000).
+
+---
+
+**content[].filterTags** (optional)
+
+You can optionally define a list of comma separated labels which are used to filter shelves.
+
+The OTT Web App offers a demonstration of its filtering capabilities. This example can be enabled by configuring the `enableLabelsFiltering` custom parameter.
+
+Filtering by device type, country, and day of week is supported. See the `Home.tsx` file for a code example.
 
 ---
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -74,6 +74,9 @@ the given screen.
 **menu[].filterTags** (optional)
 
 You can optionally define a list of comma separated tags which are used in the "filters" section on the screen.
+The OTT Web App offers a demonstration of its filtering capabilities. This example can be enabled by configuring the `enableLabelsFiltering` custom parameter.
+
+Filtering by device type, country, and day of week is supported. See the Home.tsx file for a code example.
 
 ---
 
@@ -104,8 +107,8 @@ each shelf separately.
 
 **content[].contentId**
 
-The eight-character Playlists IDs from the JW Player dashboard. These IDs populate the video "shelves" on your site. *
-*contentId** is not required if you use `continue_watching` or `favorites` **type**.
+The eight-character Playlists IDs from the JW Player dashboard. These IDs populate the video "shelves" on your site. \*
+\*contentId** is not required if you use `continue_watching` or `favorites` **type\*\*.
 
 ---
 
@@ -305,11 +308,11 @@ Note, this setting is ignored if Cleeng is not enabled (i.e. there is not Cleeng
 **integrations.cleeng.monthlyOffer** (optional)
 
 If Cleeng is enabled, and you want to show the Payments and Subscription functionality, you need to include at least one
-offer ID (either this or the yearly offer property). 
-The application uses this ID to map to an offer that you've configured in your Cleeng environment under Offers to 
-represent your monthly subscription. 
+offer ID (either this or the yearly offer property).
+The application uses this ID to map to an offer that you've configured in your Cleeng environment under Offers to
+represent your monthly subscription.
 Note that only the data used from the Cleeng offer is the price, the free days, and the free period,
-and the app does not verify if the offer length is actually monthly. 
+and the app does not verify if the offer length is actually monthly.
 If no monthly or yearly offer is configured, the Payments section will not be shown.
 
 ---
--- a/packages/ui-react/src/hooks/useFilteredContent.ts
+++ b/packages/ui-react/src/hooks/useFilteredContent.ts
@@ -37,8 +37,7 @@ const getCountryByTimezone = () => {
     return undefined;
   }
 
-  const countryCode = TIMEZONES[timezone];
-  return COUNTRIES[countryCode];
+  return COUNTRIES[TIMEZONES[timezone]];
 };
 
 const getCurrentDay = () => new Date().toLocaleString('en-US', { weekday: 'long' });
@@ -47,6 +46,7 @@ const filterDefaultContent = (item: Content) => !item?.filterTags?.length;
 
 const filterContentByDevice = (item: Content) => {
   const { isMobile, isTablet, isDesktop } = getDeviceType();
+
   const filterTags = item?.filterTags?.split(',') || [];
 
   if (filterTags.includes(DEVICE_FILTER_LABELS.mobile) && isMobile) return true;
--- a/packages/ui-react/src/hooks/useFilteredContent.ts
+++ b/packages/ui-react/src/hooks/useFilteredContent.ts
@@ -1,11 +1,64 @@
 import type { Content } from '@jwp/ott-common/types/config';
 import useBreakpoint, { Breakpoint } from '@jwp/ott-ui-react/src/hooks/useBreakpoint';
 
-// TODO: Add more filtering logic here
-export const useFilteredContent = (content: Content[]) => {
+const DEVICE_FILTER_LABELS = {
+  mobile: 'mobile',
+  tablet: 'tablet',
+  desktop: 'desktop',
+};
+
+const COUNTRY_FILTER_LABELS = {
+  nl: 'NL',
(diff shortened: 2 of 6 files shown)


packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
+import type { Content } from '@jwp/ott-common/types/config';
+import useBreakpoint, { Breakpoint } from '@jwp/ott-ui-react/src/hooks/useBreakpoint';
+
+const DEVICE_FILTER_LABELS = {
+  mobile: 'mobile-device',
+  tablet: 'tablet-device',
+  desktop: 'desktop-device',
```
AntonLantukh: I think we can switch it back to `mobile`, `tablet`, `desktop`.

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (item?.filterTags?.includes(DEVICE_FILTER_LABELS.mobile) && isMobile) return true;
+  if (item?.filterTags?.includes(DEVICE_FILTER_LABELS.tablet) && isTablet) return true;
+  if (item?.filterTags?.includes(DEVICE_FILTER_LABELS.desktop) && isDesktop) return true;
+};
+
+const filterContentByWeekDay = (item: Content, currentDay: string) => item.filterTags?.includes(currentDay);
+
+const filterContentByCountry = (item: Content, country: string | undefined) => country && item.filterTags?.includes(country);
+
+export const useFilterContent = (content: Content[]) => {
```
AntonLantukh: I was thinking about this reference implementation and it looks better to me to enable this functionality explicitly on the customer's side using a custom param. That way we can avoid potential conflicts with current tags / labels use cases. It can be a `enableLabelsFiltering` custom param with `true` value.

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
...
+    c: ['FR'],
+  },
+  'Europe/Amsterdam': {
+    u: 60,
+    d: 120,
+    c: ['NL'],
+  },
+  'Europe/Bucharest': {
+    u: 120,
+    d: 180,
```
AntonLantukh: Do we need `u` and `d` props here?

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
...
+    c: ['RO'],
+  },
+};
+
+const useDeviceType = () => {
+  const breakpoint = useBreakpoint();
```
AntonLantukh: I think there is no need for use to use this hook here. Take a look at how it is implemented, it uses `getScreenSize` function under the hood. I suggest we only take this function as we don't really need event handlers.

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
...
+  const { isMobile, isTablet, isDesktop } = useDeviceType();
+  const currentDay = getCurrentDay();
+  const country = getCountryByTimezone();
+
+  return content?.filter((item) =>
+    [
+      filterDefaultContent(item),
+      filterContentByDevice(item, isMobile, isTablet, isDesktop),
+      filterContentByWeekDay(item, currentDay),
+      filterContentByCountry(item, country),
```
AntonLantukh: I suggest we do `device`, `country` and `weekDay` calculation inside every function. So in the end we would have something like:

```
return content?.filter((item) =>
    [
        filterDefaultContent, 
        filterContentByDevice,
        filterContentByWeekDay, 
        filterContentByCountry
      ].some(fn => fn(item)),
```

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
+import type { Content } from '@jwp/ott-common/types/config';
+import useBreakpoint, { Breakpoint } from '@jwp/ott-ui-react/src/hooks/useBreakpoint';
+
+const DEVICE_FILTER_LABELS = {
```
AntonLantukh: Could you also add some basic description of how it works in `docs` folder?

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,92 @@
...
+export const useFilterContent = (content: Content[]) => {
+  const { isMobile, isTablet, isDesktop } = useDeviceType();
+  const currentDay = getCurrentDay();
+  const country = getCountryByTimezone();
+
+  return content?.filter((item) =>
+    [
+      filterDefaultContent(item),
+      filterContentByDevice(item, isMobile, isTablet, isDesktop),
+      filterContentByWeekDay(item, currentDay),
```
AntonLantukh: What if tag with a week day is not set, will this item be shown?

AntonLantukh: I think in case a tag is not set, an item should still be displayed (for all filters). What do you think?

packages/ui-react/src/hooks/useFilteredContent.ts
CarinaDraganJW: No, I removed them, nice catch 🚀

packages/ui-react/src/hooks/useFilteredContent.ts
CarinaDraganJW: Nice catch, I updated the code 🚀

packages/ui-react/src/hooks/useFilteredContent.ts
CarinaDraganJW: Yes, I agree, I will check if it gets included, if I remember correctly, during the tests I did, this should be the case 👀
But I will double check 👀

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+  'Europe/Paris': {
+    c: ['FR'],
+  },
+  'Europe/Amsterdam': {
+    c: ['NL'],
+  },
+  'Europe/Bucharest': {
+    c: ['RO'],
+  },
```
AntonLantukh:
```suggestion
const TIMEZONES = {
  'Europe/Berlin': 'DE',
  'Europe/Paris': 'FR',
  'Europe/Amsterdam': 'NL',
  'Europe/Bucharest': 'RO',
```

docs/configuration.md
```diff
@@ -74,6 +74,9 @@ the given screen.
 **menu[].filterTags** (optional)
 
 You can optionally define a list of comma separated tags which are used in the "filters" section on the screen.
+The OTT Web App offers a demonstration of its filtering capabilities. This example can be enabled by configuring the `enableLabelsFiltering` custom parameter.
```
AntonLantukh: In only works for `**content[].filterTags** (optional)`

AntonLantukh: Looks like we also need to add a new property here.

packages/ui-react/src/pages/Home/Home.tsx
```diff
@@ -3,18 +3,21 @@ import { shallow } from '@jwp/ott-common/src/utils/compare';
...
 import { useTranslation } from 'react-i18next';
+import { useFilterContent } from '@jwp/ott-ui-react/src/hooks/useFilteredContent';
 
 import ShelfList from '../../containers/ShelfList/ShelfList';
 
 const Home = () => {
   const { config } = useConfigStore(({ config, accessModel }) => ({ config, accessModel }), shallow);
   const content: Content[] = config?.content;
   const { t } = useTranslation('common');
+  const labelsFilteringEnabled = Boolean(config?.custom?.enableLabelsFiltering);
```
AntonLantukh: This one is always a string. Feel free to use a `isTruthyCustomParamValue` function.

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+const filterContentByWeekDay = (item: Content) => {
+  const currentDay = getCurrentDay();
+  return item.filterTags?.includes(currentDay);
+};
+
+const filterContentByCountry = (item: Content) => {
+  const country = getCountryByTimezone();
+  return country && item.filterTags?.includes(country);
```
AntonLantukh: We receive `filterTags` as a string, so I think we should first split it by comma.

docs/configuration.md
CarinaDraganJW: Oh, I added the description in the wrong section, I will move it to `content` 🙂

CarinaDraganJW: Nice catch! 🚀

packages/ui-react/src/hooks/useFilteredContent.ts
CarinaDraganJW: `includes` also works for strings, but I updated to use `split` 👍🏻

packages/ui-react/src/hooks/useFilteredContent.ts
```diff
@@ -0,0 +1,78 @@
...
+  NL: 'Netherlands',
+  RO: 'Romania',
+} as const;
+
+const TIMEZONES: Record<string, string> = {
+  'Europe/Berlin': 'DE',
+  'Europe/Paris': 'FR',
+  'Europe/Amsterdam': 'NL',
+  'Europe/Bucharest': 'RO',
+} as const;
```
AntonLantukh, line 22: As an alternative we can also consider using `new Intl.DateTimeFormat().resolvedOptions().locale`

AntonLantukh (approved):

packages/ui-react/src/hooks/useFilteredContent.ts
CarinaDraganJW: That could be an alternative, but it might be a bit more tricky to implement. For instance, I get locale `en-GB` in my browser but timezone `Europe/Amsterdam` 🤔

I think the risk here is that while certain locales might be associated with specific timezones (en-US and America/New_York), there isn't a one-to-one mapping. A single country can have multiple timezones, and multiple countries might share a timezone 🤔


Screenshot 2025-04-30 at 11 02 52

PR #689, opened 2025-06-19
[OWA-87] feat: add e2e tests for content lists

CarinaDraganJW:
- Fix content lists filters not showing up 
- Add e2e tests for content lists


platforms/web/test-e2e/tests/content_list_test.ts
```diff
@@ -0,0 +1,62 @@
+import { testConfigs } from '@jwp/ott-testing/constants';
+
+import constants from '#utils/constants';
+import { checkSelectedFilterButton, selectFilterAndCheck } from '#utils/filters';
+
+const allFilters = ['Fantasy', 'Drama', 'All'];
+const dramaFilms = ['Spring', 'Tears of Steel Trailer'];
```
AntonLantukh: If it is a trending list, then there is no guarantee we will always see this content. It may change depending on recent views. So it could make sense just to check the presence of the content, but not actual items.

platforms/web/test-e2e/tests/content_list_test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+Scenario('Header button navigates to content list screen', async ({ I }) => {
+  I.see('All Popular');
+  I.seeInCurrentUrl(`${constants.baseUrl}n/`);
+});
+
+Scenario('Content list screen loads', async ({ I }) => {
+  I.seeElement('div[role="grid"]');
+  I.seeElement('div[role="gridcell"]');
+  I.seeElement('div[id="layout_grid_0-0"]');
```
AntonLantukh: Maybe we could add a `data-testid` to div[role="grid"] and then click on a first child of it? Just thinking how we could avoid using "role" in the selector.

AntonLantukh (approved):

PR #691, opened 2025-07-01
Chore / maintenance 2025 q2

ChristiaanScheermeijer:
## Description

I've performed quarterly maintenance and fixed all `high` and `critical` vulnerabilities. There are two `moderate` vulnerabilities remaining which can be solved if we (major) upgrade the `@adyen/adyen-web` dependency.

Noteworthy changes/upgrades:

- babel dependencies are moved to the root package.json to force hoisting
- vitest `2.1.8` -> `3.0.8`
  - workspace file is deprecated so I refactored this to vitest.config.ts
- vite `5.4.11` -> `7.0.0` 
- Due to the vite/vitest upgrade all snapshots have been updated with a new className hash

There are some major upgrades coming up which will be submitted separately. At least `inversify@7` and `zustand@5` are on the way. 

@AntonLantukh who should I add as reviewers?


package.json
```diff
@@ -32,37 +32,33 @@
     "pre-commit": "yarn depcheck && lint-staged",
     "prepare": "husky install",
     "test": "TZ=UTC LC_ALL=en_US.UTF-8 vitest run",
+    "test-update": "TZ=UTC LC_ALL=en_US.UTF-8 vitest run -u",
```
AntonLantukh, line 35: Did we have to update snapshots because of `vitest` version change?

package.json
```diff
@@ -32,37 +32,33 @@
...
     "test-watch": "TZ=UTC LC_ALL=en_US.UTF-8 vitest",
     "web": "yarn --cwd platforms/web",
     "access-bridge": "yarn --cwd platforms/access-bridge"
   },
   "devDependencies": {
-    "@commitlint/cli": "^19.6.0",
-    "@commitlint/config-conventional": "^19.6.0",
-    "@types/node": "^22.10.1",
+    "@babel/core": "^7.26.10",
+    "@babel/plugin-proposal-decorators": "^7.25.9",
```
AntonLantukh, line 42: Should we make them part of "dependencies"?

AntonLantukh: Also, do we want to keep babel here to support it also for other "platforms"?

package.json
ChristiaanScheermeijer: Yes, that's correct

package.json
ChristiaanScheermeijer: > Should we make them part of "dependencies"?

Hmm, the docs state to use `--save-dev`, but I agree that this can be considered a dependency. We use the following (simplified) rules: 

- If the dependency is imported from a src file it is a dependency
- If the dependency is not imported or imported/used by build scripts, it is a dev dependency

The grey area here is when the build tool adds code to the bundled code. IMO many dependencies affect the bundled code (vite, babel, ...), but we've agreed to declare these as dev dependencies anyways.

> Also, do we want to keep babel here to support it also for other "platforms"?

This was actually needed to prevent duplicate installations of babel. We have 5 platforms which all use vite or expo, but after this update, the babel version was not sync anymore causing issues. Adding the dependency here fixed that, but I hope that this is temporary

AntonLantukh (approved):

PR #692, opened 2025-07-01
Chore / upgrade Inversify

ChristiaanScheermeijer:
## Description

This PR upgrades the Inversify library to v7. This upgrade has some breaking changes in the types export and removed support for the synchronous `rebind` method. 

For this reason it is now needed to needed to do:

```ts
await container.unbind(SettingsService);
container.bind(SettingsService).to(CustomSettingsService);
```

This means we need to wait for the registration before bootstrapping the app (see `platforms/web/modules/register.ts`).

When extending an existing class (the SettingsService for example) you'll need to add a second decorator to ensure that the parent class receives the injected services in the constructor method. This is not needed when the concrete class has a constructor and calls `super`.

In our React Native app we have a custom implementation for the SettingsService to prevent the ini load mechanism. The `@injectFromBase` decorator is needed here so the SettingsService constructor still works.

```ts
@injectable()
@injectFromBase({ extendConstructorArguments: true })
export default class ReactNativeSettingsService extends SettingsService {
```

### DIFF
--- a/platforms/web/src/modules/register.ts
+++ b/platforms/web/src/modules/register.ts
@@ -29,44 +29,45 @@ import { LocalStorageService } from '#src/services/LocalStorageService';
  * })
  * ```
  */
+export const register = async () => {
+  container.bind(StorageService).to(LocalStorageService);
 
-container.bind(StorageService).to(LocalStorageService);
+  container.bind(BROADCAST_CHANNEL).toConstantValue(new BroadcastChannel('jwp-refresh-token-channel'));
 
-container.bind(BROADCAST_CHANNEL).toConstantValue(new BroadcastChannel('jwp-refresh-token-channel'));
+  // Currently, this is only used for e2e testing to override the customer ip from a browser cookie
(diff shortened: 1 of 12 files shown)


platforms/web/src/modules/register.ts
```diff
@@ -29,44 +29,45 @@ import { LocalStorageService } from '#src/services/LocalStorageService';
  * })
  * ```
  */
+export const register = async () => {
```
AntonLantukh, line 32: Hm, do we need to wrap it in async in case we don't use `rebind`. Or do you have it within your local implementation? I remember we wanted to make it sort of react agnostic and do this binding not in the component.

ChristiaanScheermeijer: Yes, that is correct. It could be non-async here, but then it needs to be changed when rebinding a service. In our fork, we've swapped some services, so it needs to be async. Do you prefer it to be sync be default?

AntonLantukh: I don't mind having it as async.

AntonLantukh (approved):

AntonLantukh:
@ChristiaanScheermeijer could you resolve conflicts here?

